PR #22, opened 2019-12-27
fix: make the getStorageClass() method public

### LINKED ISSUE #21: Cloud Storage: getStorageClass in SetStorageClassLifecycleAction is not visible
getStorageClass() method in SetStorageClassLifecycleAction class is only visible inside the package. Its not accessible outside the package.

I am trying to read storage class set by the lifecycle rule through SDK in the following way. looks like its not possible. Can you suggest any other alternative to access if this is not possible.

```
f (lifecycleAction.getActionType().equals(SetStorageClassLifecycleAction.TYPE)) {
     return ((SetStorageClassLifecycleAction)lifeCycleRule.getAction()).getStorageClass(); 
}
```

**// Compiler error : The method getStorageClass() from the type BucketInfo.LifecycleRule.SetStorageClassLifecycleAction is  not visible**

#### Environment details

OS type and version: Windows 10
Java version: JDK 11
sdk version: google-cloud-storage-1.7.8.0.jar

dmitry-fa:
Fixes #21

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -663,7 +663,7 @@ public class BucketInfo implements Serializable {
             .toString();
       }
 
-      StorageClass getStorageClass() {
+      public StorageClass getStorageClass() {
         return storageClass;
       }
     }


chingor13:
This one is not up to me. Adding Frank

dmitry-fa:
@frankyn gentle ping

PR #28, opened 2020-01-03
feat: integrate disable Lifecycle rule api for to remove lifecycle rule of bucket

### LINKED ISSUE #20: Cloud Storage: Removing lifecyclerules from bucket is not working
I was successfully able to add the lifecycle rules to a bucket. 
```
Storage storage=/* connection*/;
Builder builder = BucketInfo.newBuilder(name);
builder.setStorageClass(COLDLINE);
builder.setLifecycleRules( lifecycleRules);
storage.create(builder.build());
```

But seems there is no way to remove the lifecycle rules from bucket

```
Storage storage=/* connection*/;
Builder builder = BucketInfo.newBuilder(name);
builder.setLifecycleRules(null);
storage.update(builder.build());
```

Looks like there is no support in SDK for PATCH update where as Rest API for lifecycle configuration is available .

https://cloud.google.com/storage/docs/managing-lifecycles

```
curl -X PATCH --data-binary @[LIFECYCLE_CONFIG_FILE].json \
  -H "Authorization: Bearer [OAUTH2_TOKEN]" \
  -H "Content-Type: application/json" \
  "https://storage.googleapis.com/storage/v1/b/[BUCKET_NAME]?fields=lifecycle"

```

Environment details
OS type and version: Windows 10
Java version: JDK 11
sdk version: google-cloud-storage-1.7.8.0.jar

athakor:
Fixes #20

### DIFF
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -92,6 +92,14 @@
       <groupId>org.threeten</groupId>
       <artifactId>threetenbp</artifactId>
     </dependency>
+    <dependency>
+      <groupId> com.google.api.grpc</groupId>
+      <artifactId>proto-google-cloud-iamcredentials-v1</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+    </dependency>
 
     <!-- Test dependencies -->
     <dependency>
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -190,10 +190,6 @@
       <artifactId>httpcore</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.google.protobuf</groupId>
-      <artifactId>protobuf-java</artifactId>
-    </dependency>
   </dependencies>
 
   <build>
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -96,10 +96,6 @@
       <groupId> com.google.api.grpc</groupId>
       <artifactId>proto-google-cloud-iamcredentials-v1</artifactId>
     </dependency>
-    <dependency>
-      <groupId>com.google.code.gson</groupId>
-      <artifactId>gson</artifactId>
-    </dependency>
 
     <!-- Test dependencies -->
     <dependency>
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -73,10 +73,6 @@
       <artifactId>opencensus-api</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.google.cloud</groupId>
-      <artifactId>google-cloud-iamcredentials</artifactId>
-    </dependency>
-    <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>proto-google-iam-v1</artifactId>
     </dependency>
@@ -91,10 +87,6 @@
     <dependency>
       <groupId>org.threeten</groupId>
       <artifactId>threetenbp</artifactId>
-    </dependency>
-    <dependency>
-      <groupId> com.google.api.grpc</groupId>
-      <artifactId>proto-google-cloud-iamcredentials-v1</artifactId>
     </dependency>
 
     <!-- Test dependencies -->
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -73,6 +73,10 @@
       <artifactId>opencensus-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-iamcredentials</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.api.grpc</groupId>
       <artifactId>proto-google-iam-v1</artifactId>
     </dependency>
@@ -189,6 +193,10 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpcore</artifactId>
       <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
     </dependency>
   </dependencies>
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
@@ -1140,8 +1140,9 @@ public class Bucket extends BucketInfo {
    *
    * @return {@code true} if the bucket lifecycle rules were deleted.
    * @throws StorageException upon failure
+   * @throws IOException
    */
-  public boolean deleteLifecycleRules() {
+  public boolean deleteLifecycleRules() throws IOException {
     return storage.deleteLifecycleRules(getName());
   }
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
@@ -1115,24 +1115,27 @@ public class Bucket extends BucketInfo {
   }
 
   /**
-   * Disable or delete lifecycle rules of the requested bucket.
+   * Delete lifecycle rules of the requested bucket.
    *
-   * <p>Example of disable or delete lifecycle rules of the bucket.
+   * <p>Example of delete lifecycle rules of the bucket.
    *
    * <pre>{@code
    * String bucketName = "my-unique-bucket";
    * String serviceAccount = "client-email";
-   * Bucket bucket = storage.create(BucketInfo.newBuilder(bucketName)
-   *     .setStorageClass(StorageClass.COLDLINE)
-   *     .setLocation("us-central1")
-   *     .setLifecycleRules(ImmutableList.of(
-   *             new LifecycleRule(
-   *                 LifecycleAction.newDeleteAction(),
-   *                 LifecycleCondition.newBuilder().setAge(2).build())))
-   *     .build());
-   * boolean isDisable  = bucket.disableLifeCycleRule(bucketName,serviceAccount);
-   * if (isDisable) {
-   *   // the lifecycle rules was disabled or deleted
+   * Bucket bucket =
+   *     storage.create(
+   *         BucketInfo.newBuilder(bucketName)
+   *             .setStorageClass(StorageClass.COLDLINE)
+   *             .setLocation("us-central1")
+   *             .setLifecycleRules(
+   *                 ImmutableList.of(
+   *                     new LifecycleRule(
+   *                         LifecycleAction.newDeleteAction(),
+   *                         LifecycleCondition.newBuilder().setAge(2).build())))
+   *             .build());
+   * boolean isDeleted = bucket.deleteLifecycleRules(bucketName, serviceAccount);
+   * if (isDeleted) {
+   *   // the lifecycle rules was deleted
    * }
    * }</pre>
    *
@@ -1141,8 +1144,8 @@ public class Bucket extends BucketInfo {
    * @return {@code true} if the bucket lifecycle rules was deleted.
    * @throws StorageException upon failure
    */
-  public boolean disableLifeCycleRules(String bucket, String serviceAccount) {
-    return storage.disableLifeCycleRules(bucket, serviceAccount);
+  public boolean deleteLifecycleRules(String bucket, String serviceAccount) {
+    return storage.deleteLifecycleRules(bucket, serviceAccount);
   }
 
   /**
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
@@ -1115,13 +1115,12 @@ public class Bucket extends BucketInfo {
   }
 
   /**
-   * Delete lifecycle rules of the requested bucket.
+   * Delete the lifecycle rules of this bucket.
    *
-   * <p>Example of delete lifecycle rules of the bucket.
+   * <p>Example of delete the lifecycle rules of this bucket.
    *
    * <pre>{@code
    * String bucketName = "my-unique-bucket";
-   * String serviceAccount = "client-email";
    * Bucket bucket =
    *     storage.create(
    *         BucketInfo.newBuilder(bucketName)
@@ -1133,19 +1132,17 @@ public class Bucket extends BucketInfo {
    *                         LifecycleAction.newDeleteAction(),
    *                         LifecycleCondition.newBuilder().setAge(2).build())))
    *             .build());
-   * boolean isDeleted = bucket.deleteLifecycleRules(bucketName, serviceAccount);
-   * if (isDeleted) {
-   *   // the lifecycle rules was deleted
+   * boolean rulesDeleted = bucket.deleteLifecycleRules();
+   * if (rulesDeleted) {
+   *   // the lifecycle rules were deleted
    * }
    * }</pre>
    *
-   * @param bucket name of the bucket.
-   * @param serviceAccount client_email which is present in credential.json file.
-   * @return {@code true} if the bucket lifecycle rules was deleted.
+   * @return {@code true} if the bucket lifecycle rules were deleted.
    * @throws StorageException upon failure
    */
-  public boolean deleteLifecycleRules(String bucket, String serviceAccount) {
-    return storage.deleteLifecycleRules(bucket, serviceAccount);
+  public boolean deleteLifecycleRules() {
+    return storage.deleteLifecycleRules(getName());
   }
 
   /**
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
@@ -1141,8 +1141,8 @@ public class Bucket extends BucketInfo {
    * @return {@code true} if the bucket lifecycle rules was deleted.
    * @throws StorageException upon failure
    */
-  public boolean disableLifeCycleRule(String bucket, String serviceAccount) {
-    return storage.disableLifeCycleRule(bucket, serviceAccount);
+  public boolean disableLifeCycleRules(String bucket, String serviceAccount) {
+    return storage.disableLifeCycleRules(bucket, serviceAccount);
   }
 
   /**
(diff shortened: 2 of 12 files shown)


JesseLovelace:
Does this API remove all lifecycle rules? if so, it should be named disableLifecycleRules rather than rule, and if not there should be documentation on which rule specifically would be removed and how to control it

athakor:
> Does this API remove all lifecycle rules? if so, it should be named disableLifecycleRules rather than rule, and if not there should be documentation on which rule specifically would be removed and how to control it

@JesseLovelace yes,this API remove all lifecycle rules. so you are right we should named it as `disableLifecycleRules`.

athakor:
@JesseLovelace PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2779,6 +2779,38 @@ Blob create(
...
+   *         BucketInfo.newBuilder(bucketName)
+   *             .setStorageClass(StorageClass.COLDLINE)
+   *             .setLocation("us-central1")
+   *             .setLifecycleRules(
+   *                 ImmutableList.of(
+   *                     new LifecycleRule(
+   *                         LifecycleAction.newDeleteAction(),
+   *                         LifecycleCondition.newBuilder().setAge(2).build())))
+   *             .build());
+   * boolean isDeleted = storage.deleteLifecycleRules(bucketName, serviceAccount);
```
JesseLovelace: isDeleted should be rulesDeleted

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2779,6 +2779,38 @@ Blob create(
...
+   * if (isDeleted) {
+   *   // the lifecycle rules was deleted
```
JesseLovelace: "was deleted" should be "were deleted"

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
```diff
@@ -1114,6 +1114,40 @@ public boolean deleteDefaultAcl(Entity entity) {
...
+   *   // the lifecycle rules was deleted
+   * }
+   * }</pre>
+   *
+   * @param bucket name of the bucket.
+   * @param serviceAccount client_email which is present in credential.json file.
+   * @return {@code true} if the bucket lifecycle rules was deleted.
+   * @throws StorageException upon failure
+   */
+  public boolean deleteLifecycleRules(String bucket, String serviceAccount) {
```
JesseLovelace: Since this method within a bucket, bucket shouldn't be an argument here. I don't believe lifecycle rules are tied to serviceaccounts either, so you should be able to remove that argument and just use the default service account. So the method signature should just be deleteLifecycleRules(), which makes a lot more sense for a customer to call from a bucket object

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
```diff
@@ -1114,6 +1114,40 @@ public boolean deleteDefaultAcl(Entity entity) {
     return storage.deleteDefaultAcl(getName(), entity);
   }
 
+  /**
+   * Delete lifecycle rules of the requested bucket.
```
JesseLovelace: Let's change this to delete the lifecycle rules of *this* bucket

JesseLovelace:
Also, your IT is failing because you aren't using a real service account email

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
athakor: done

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java

athakor:
> Also, your IT is failing because you aren't using a real service account email

@chingor13 @JesseLovelace system test failed due to Permission denied for the` IAM Service Account Cred.` Could you add the permission for this project?

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
athakor: correct bucket shouldn't be an argument here and also method signature should be `deleteLifecycleRules().`But we need service account email to generate the accesstoken.you can see [here](https://github.com/googleapis/java-storage/pull/28/files#diff-ea0375140a7d72eccf35d58e1544f63bR1031)

JesseLovelace: Use the default service account, i.e. storage.getServiceAccount(projectId)

JesseLovelace:
Try using storage.getServiceAccount(projectId) instead of extracting it from the json like you're doing now

athakor:
> Try using storage.getServiceAccount(projectId) instead of extracting it from the json like you're doing now

done @JesseLovelace PTAL

athakor:
@JesseLovelace PTAL

athakor:
@frankyn @JesseLovelace  gentle ping.

frankyn:
@JesseLovelace is on point for this one. I'll bring it up in our meeting early next week (Tuesday).

Thank you for your patience @athakor

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
```diff
@@ -1114,6 +1114,37 @@ public boolean deleteDefaultAcl(Entity entity) {
     return storage.deleteDefaultAcl(getName(), entity);
   }
 
+  /**
+   * Delete the lifecycle rules of this bucket.
+   *
+   * <p>Example of delete the lifecycle rules of this bucket.
```
JesseLovelace: delete->deleting

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2833,6 +2833,36 @@ Blob create(
    */
   boolean deleteDefaultAcl(String bucket, Entity entity);
 
+  /**
+   * Delete the lifecycle rules of the requested bucket.
+   *
+   * <p>Example of delete the lifecycle rules of the requested bucket.
```
JesseLovelace: delete->deleting

JesseLovelace:
LGTM overall just fix the typos above

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
athakor: done

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java

JesseLovelace:
Will need to look into the linkage monitor failure.

As a follow-up, do you think it'd be possible to make so that setting the lifecycle rules to null and updating also works?

athakor:
> Will need to look into the linkage monitor failure.
> 
> As a follow-up, do you think it'd be possible to make so that setting the lifecycle rules to null and updating also works?

@JesseLovelace here we cant do anything to fix this  linkage issue,we can fix it in [java-storage-nio-FakeStorageRpc](https://github.com/googleapis/java-storage-nio/blob/df030b86e43dd8ea627b25e2ea7700510abf2ba3/google-cloud-nio/src/main/java/com/google/cloud/storage/contrib/nio/testing/FakeStorageRpc.java#L78). I dont think so,it'would be possible to set the lifecycle rules to null and updating works without api call.

athakor:
@frankyn @JesseLovelace  here, how we can move ahead with above situation?

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
```diff
@@ -1114,6 +1114,37 @@ public boolean deleteDefaultAcl(Entity entity) {
...
+   *         BucketInfo.newBuilder(bucketName)
+   *             .setStorageClass(StorageClass.COLDLINE)
+   *             .setLocation("us-central1")
+   *             .setLifecycleRules(
+   *                 ImmutableList.of(
+   *                     new LifecycleRule(
+   *                         LifecycleAction.newDeleteAction(),
+   *                         LifecycleCondition.newBuilder().setAge(2).build())))
+   *             .build());
+   * boolean rulesDeleted = bucket.deleteLifecycleRules();
```
frankyn, line 1135: How do I delete 1 lifecycle rule?

athakor: @frankyn I think there is no possible way to delete or disable any 1 lifecycle rule please check https://cloud.google.com/storage/docs/managing-lifecycles#disable

google-cloud-storage/pom.xml
```diff
@@ -172,6 +172,10 @@
       <version>4.5.12</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+    </dependency>
```
frankyn, line 178: this doesn't look necessary. Could you please remove?

pom.xml
```diff
@@ -65,6 +65,8 @@
     <site.installationModule>google-cloud-storage-parent</site.installationModule>
     <google.core.version>1.93.4</google.core.version>
     <google.api-common.version>1.9.0</google.api-common.version>
+    <google.iamcredentials.version>0.43.0-alpha</google.iamcredentials.version>
+    <protobuf.version>3.11.1</protobuf.version>
```
frankyn, line 69: version isn't tokenized below.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3208,4 +3209,36 @@ public void testBucketLogging() throws ExecutionException, InterruptedException
       RemoteStorageHelper.forceDelete(storage, loggingBucket, 5, TimeUnit.SECONDS);
     }
   }
+
+  @Test
+  public void testDeleteLifecycleRules() throws ExecutionException, InterruptedException {
```
frankyn, line 3214: Following up, did you try deleting a single lifecycle rule?

athakor: @frankyn 

I have tried to find the feasible way to delete a single lifecycle rule but its look like there is no possible way to do that and also found that across the all languages have same behavior which we have currently implemented.

**Suggestion** 
-  I think we should have to update the method name like `clearLifecycleRules `or `disableLifecycleRules `instead of `deleteLifecycleRules `to be more clear, `deleteLifecycleRules `might confused to user.

- we can also place a  `Note` there like delete individual rules through the console currently  this library have limited support  to delete single rule.

pom.xml
athakor: done

google-cloud-storage/pom.xml
athakor: done

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
frankyn: Hi @athakor, apologies for the delay.
```java

    String lifecycleTestBucketName = RemoteStorageHelper.generateBucketName();
    storage.create(
            BucketInfo.newBuilder(lifecycleTestBucketName)
                    .setLocation("us")
                    .setLifecycleRules(
                            ImmutableList.of(
                                    new LifecycleRule(
                                            LifecycleAction.newSetStorageClassAction(StorageClass.COLDLINE),
                                            LifecycleCondition.newBuilder()
                                                    .setAge(1)
                                                    .setNumberOfNewerVersions(3)
                                                    .setIsLive(false)
                                                    .setCreatedBefore(new DateTime(System.currentTimeMillis()))
                                                    .setMatchesStorageClass(ImmutableList.of(StorageClass.COLDLINE))
                                                    .build()),
                                    new LifecycleRule(
                                            LifecycleAction.newDeleteAction(),
                                            LifecycleCondition.newBuilder()
                                                    .setAge(1)
                                                    .build()
                                    )))
                    .build());
    // Delete OLM rule.
    Bucket remoteBucket =
            storage.get(lifecycleTestBucketName, Storage.BucketGetOption.fields(BucketField.LIFECYCLE));
    int priorSize = remoteBucket.getLifecycleRules().size();
    ArrayList<LifecycleRule> lifecycleRules = new ArrayList(remoteBucket.getLifecycleRules());
    Iterator<LifecycleRule> iterator = lifecycleRules.iterator();
    while(iterator.hasNext()) {
      LifecycleRule rule = iterator.next();
      if (rule.getAction().getActionType().equals(LifecycleRule.DeleteLifecycleAction.TYPE)) {
        iterator.remove();
      }
    }
    remoteBucket.toBuilder().setLifecycleRules(lifecycleRules).build().update();
```

frankyn: Did this help? I think we can merge this change once you add a helper to Storage interface but no change is required for StorageRpc.java.

athakor: @frankyn thanks for info, I tried your sample code but its not working please check the below response.

```
[
  LifecycleRule{
    lifecycleAction=DeleteLifecycleAction{
      actionType=Delete
    },
    lifecycleCondition=LifecycleCondition{
      age=1,
      createBefore=null,
      numberofNewerVersions=null,
      isLive=null,
      matchesStorageClass=null
    }
  },
  LifecycleRule{
    lifecycleAction=SetStorageClassLifecycleAction{
      actionType=SetStorageClass,
      storageClass=COLDLINE
    },
    lifecycleCondition=LifecycleCondition{
      age=1,
      createBefore=2020-05-15,
      numberofNewerVersions=3,
      isLive=false,
      matchesStorageClass=[
        COLDLINE
      ]
    }
  }
]
```
does it works on your end?

frankyn: Good catch, looks like there's an issue with updating LifecycleRules when the metadata information is set.

Here's a workaround for the bug:

```java
package com.example.storage;

// [START storage_quickstart]
// Imports the Google Cloud client library
import com.google.api.client.util.DateTime;
import com.google.cloud.storage.Bucket;
import com.google.cloud.storage.BucketInfo;
import com.google.cloud.storage.Storage;
import com.google.cloud.storage.StorageClass;
import com.google.common.collect.ImmutableList;

import java.util.ArrayList;
import java.util.Iterator;

public class QuickstartSample {
  public static void main(String... args) throws Exception {
    // Instantiates a client

    // The name for the new bucket
    String bucketName = "anima-frank-lifecycletestsdf";  // "my-new-bucket";

    // Creates the new bucket
    Bucket bucket = storage.create(BucketInfo.newBuilder(bucketName)
            .setLocation("us")
            .setLifecycleRules(
                    ImmutableList.of(
                            new BucketInfo.LifecycleRule(
                                    BucketInfo.LifecycleRule.LifecycleAction.newSetStorageClassAction(StorageClass.COLDLINE),
                                    BucketInfo.LifecycleRule.LifecycleCondition.newBuilder()
                                            .setAge(1)
                                            .setNumberOfNewerVersions(3)
                                            .setIsLive(false)
                                            .setCreatedBefore(new DateTime(System.currentTimeMillis()))
                                            .setMatchesStorageClass(ImmutableList.of(StorageClass.COLDLINE))
                                            .build()),
                            new BucketInfo.LifecycleRule(
                                    BucketInfo.LifecycleRule.LifecycleAction.newDeleteAction(),
                                    BucketInfo.LifecycleRule.LifecycleCondition.newBuilder()
                                            .setAge(1)
                                            .build()
                            )))
            .build());
    // Delete OLM rule.
    Bucket remoteBucket =
            storage.get(bucketName, Storage.BucketGetOption.fields(Storage.BucketField.LIFECYCLE));
    int priorSize = remoteBucket.getLifecycleRules().size();
    ArrayList<BucketInfo.LifecycleRule> lifecycleRules = new ArrayList(remoteBucket.getLifecycleRules());
    Iterator<BucketInfo.LifecycleRule> iterator = lifecycleRules.iterator();
    while(iterator.hasNext()) {
      BucketInfo.LifecycleRule rule = iterator.next();
      if (rule.getAction().getActionType().equals(BucketInfo.LifecycleRule.DeleteLifecycleAction.TYPE)) {
        iterator.remove();
        System.out.println("???");
      }
    }
    System.out.printf("Updated size: %d %n", lifecycleRules.size());
    System.out.println(lifecycleRules);
    remoteBucket =
            storage.get(bucketName, Storage.BucketGetOption.fields(Storage.BucketField.STORAGE_CLASS));

    remoteBucket = remoteBucket.toBuilder().setLifecycleRules(lifecycleRules).build().update();
    System.out.println(remoteBucket.getLifecycleRules());
    storage.delete(bucketName);
    System.out.printf("Bucket %s (updated: %d, prior: %d) created.%n", bucket.getName(), remoteBucket.getLifecycleRules().size(), priorSize);
  }
}
```

athakor: @frankyn sorry for the delay and thanks for the workaround. Partially this will works in case of `empty bucket` but what will happen if `bucket `contain list of `blobs`? in that case it will gets failed with following error :
```javascript
{
  "code" : 409,
  "errors" : [ {
    "domain" : "global",
    "message" : "The bucket you tried to delete was not empty.",
    "reason" : "conflict"
  } ],
  "message" : "The bucket you tried to delete was not empty."
}
```
The main problem with this solution is that, it will remove whole `bucket `and that never get back. so i think we need to go with current implementation or we need to find something new for this.

frankyn: `storage.delete(bucketName)` is clean up for the example and not required by remove lifecycle rule. 

Please let me know if there is still confusion with the workaround. In short, the library should not require the workaround.

athakor: @frankyn thanks for the clarification. It's works i will raise separate PR by adding these helper to Storage interface.  

Thank you for your help.

frankyn: Thank you! I appreciate your patience.

frankyn: @athakor I may have may confused, you're going to close this PR and make another right?

PR #66, opened 2020-01-15
fix deprecations and other warnings

elharo:
@BenWhitehead


BenWhitehead:
Contents of change seem fine to me, but it looks like you'll need to run the code formatter to get the lints to pass.

elharo:
ran the formatter

PR #95, opened 2020-01-28
Update HTTP client

elharo:
@suztomo

[WARNING] Rule 2: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for org.apache.httpcomponents:httpclient:4.5.10 paths to dependency are:
+-com.google.cloud:google-cloud-storage:1.103.1
  +-com.google.http-client:google-http-client:1.34.1
    +-org.apache.httpcomponents:httpclient:4.5.10 (managed) <-- org.apache.httpcomponents:httpclient:4.5.11
]

### DIFF
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -24,8 +24,17 @@
       <artifactId>guava</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.http-client</groupId>
+      <artifactId>google-http-client-jackson2</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.api-client</groupId>
       <artifactId>google-api-client</artifactId>
+      <version>1.30.7</version>
     </dependency>
     <dependency>
       <groupId>com.google.apis</groupId>
@@ -42,14 +51,6 @@
     <dependency>
       <groupId>com.google.api</groupId>
       <artifactId>gax</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.google.http-client</groupId>
-      <artifactId>google-http-client</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.google.http-client</groupId>
-      <artifactId>google-http-client-jackson2</artifactId>
     </dependency>
     <dependency>
       <groupId>com.google.auth</groupId>
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -20,6 +20,10 @@
   </properties>
   <dependencies>
     <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.apis</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -133,11 +133,6 @@
         <artifactId>google-api-services-storage</artifactId>
         <version>v1-rev20191011-1.30.3</version>
       </dependency>
-      <dependency>
-        <groupId>com.google.guava</groupId>
-        <artifactId>guava</artifactId>
```
suztomo: Nice. This is duplicate. Guava is the first entry in this dependencyManagement section.

google-cloud-storage/pom.xml
```diff
@@ -165,11 +165,6 @@
       <artifactId>google-cloud-conformance-tests</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
```
suztomo: Isn’t this dependency used in the file below?
https://github.com/googleapis/java-storage/blob/4ab4af704ead2331d4514a20c980cf878fcee858/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java#L120

suztomo:
I've never seen RequireUpperBoundDeps complaining on only one dependency path. Interesting.

Memo:
Link to the error: https://source.cloud.google.com/results/invocations/5dbef53d-62c1-42f9-89ba-11dd9613f28c/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fjava-storage%2Fpresubmit%2Fdependencies/log

Checking further.

suztomo:
httpclient 4.5.10 comes from google-api-client-parent:

```
~/java-storage $ mvn help:effective-pom -Dverbose=true

...
        <dependency>
          <groupId>org.apache.httpcomponents</groupId>  <!-- com.google.api-client:google-api-client-parent:1.30.7, line 125 -->
          <artifactId>httpclient</artifactId>  <!-- com.google.api-client:google-api-client-parent:1.30.7, line 126 -->
          <version>4.5.10</version>  <!-- com.google.api-client:google-api-client-parent:1.30.7, line 127 -->
        </dependency>
```

@elharo Is this what you're looking for?

suztomo:
Next question: where is 4.5.**11** coming from?

elharo:
Possibly. 4.5.11 is coming from google-http-client. That's not a mystery.

suztomo:
RequireUpperBoundDeps complains when dependencyManagement overrides a version to older one.

Screen Shot 2020-01-28 at 9 09 15

[Its documentation](https://maven.apache.org/enforcer/enforcer-rules/requireUpperBoundDeps.html) says:

> This rule requires that the version for each dependency resolved during a build, is equal to or higher than all transitive dependency declarations. 

This dependencyManagement problem matches the description.

elharo:
The problem is google-api-client doesn't publish a BOM and we're pulling in its artifact instead.

google-cloud-storage/pom.xml
suztomo: I assume now this dependency declaration should come back?

elharo: No, it shouldn't. It caused problems here so it needs to move to the only pom.xml where it's used.

PR #102, opened 2020-01-30
feat: add delimiter BlobListOption

### LINKED ISSUE #99: Add delimiter BlobListOption
Currently, it's possible to use '/' as a delimiter by using BlobListOption.currentDirectory()

That's a confusing name, let's add the delimiter option and direct users to use delimiter('/') instead.

athakor:
Fixes #99

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
@@ -988,9 +988,14 @@ public interface Storage extends Service<StorageOptions> {
       return new BlobListOption(StorageRpc.Option.DELIMITER, true);
     }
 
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -988,6 +988,11 @@ public static BlobListOption currentDirectory() {
       return new BlobListOption(StorageRpc.Option.DELIMITER, true);
     }
 
+    /** Returns an option to set a delimiter '/'. */
+    public static BlobListOption delimiter() {
```
frankyn: Delimiter can be user defined, generally '/' is the one used most often, but I'd like to keep this open to other delimiters.

athakor:
@frankyn PTAL

PR #114, opened 2020-02-04
feat: update V4 signature to pass conformance tests

JesseLovelace:
This updates the conformance tests and adds the consistency changes necessary to make them pass, as per https://github.com/googleapis/java-storage/issues/84 and https://github.com/googleapis/java-storage/issues/85


google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -1038,6 +1053,7 @@ public static BlobListOption fields(BlobField... fields) {
       HOST_NAME,
       PATH_STYLE,
       VIRTUAL_HOSTED_STYLE,
+      BUCKET_BOUND_HOST_NAME,
```
frankyn: Jesse, I want to confirm that HOST_NAME can't be in the same way as BUCKET_BOUND_HOST_NAME. Is this accurate?

PR #116, opened 2020-02-04
feat: add blob.uploadfrom(inputstream)

### LINKED ISSUE #62: Add Blob.uploadFrom(InputStream)
We have [Blob.downloadTo(OutputStream)](https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/storage/Blob.html#downloadTo-java.io.OutputStream-com.google.cloud.storage.Blob.BlobSourceOption...-) (and other downloadFrom variants)

Should probably have uploadFrom with a stream. People currently have to buffer/chunk their writes [manually](https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/storage/Blob.html#writer-com.google.cloud.storage.Storage.BlobWriteOption...-).

/cc @crwilcox

dmitry-fa:
Fixes #62

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -38,9 +38,11 @@ import com.google.common.base.Function;
 import com.google.common.io.BaseEncoding;
 import com.google.common.io.CountingOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.OutputStream;
 import java.net.URL;
+import java.nio.ByteBuffer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.Key;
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +262,80 @@ public void downloadTo(Path path) {
...
   }
 
+  /**
+   * Uploads the given file path to this blob using specified blob write options.
+   *
+   * @param path file to be uploaded
+   * @param options blob write options
+   * @throws StorageException upon failure
+   */
+  public void uploadFrom(Path path, BlobWriteOption... options) {
```
frankyn: IIUC, Blob() is a private constructor so these helpers should be moved under Storage.java instead. 

That's my only issue with implementation right now. It also follows the pattern we have for other RPC's. DownloadTo was placed under Blob by accident and should have been placed under Storage.java.

dmitry-fa: Hi @frankyn,

Thanks for your feedback. Sorry, but I don't completely agree with your points.

> DownloadTo was placed under Blob by accident and should have been placed under Storage.java

uploadFrom() methods do not introduce anything fundamentally new, they are just helpers upon
`storage.writer(Blob, Options...)`.
Their code could be moved under Storage, but I personally see no benefits from doing that.
If this is implemented as part of the Storage interface like:
```
interface Storage {
   void downloadTo(Blob, Path, Options...);
   void uploadFrom(Blob, Path, Options...);
}
```
to use this API one should have two references: storage and blob.
If it's implemented under Blob a single reference blob will be enough (each Blob instance keeps the reference to its storage)

Moreover, there are many other methods under the Blob class similar to download/upload:
`exist()`, `getContent()`, `reload()`,` update()`, `delete()`,` copyTo()` ...
I believe they are on the right place.


> Blob constructor is private and would require at least a GET request to get the Blob which will fail because the object would not exist yet.

I assume the following usage:

```
        BlobInfo blobInfo = BlobInfo.newBuilder(bucketName, blobName).build();
        Blob blob = storage.create(blobInfo);
        blob.uploadFrom(somefile);
```
There should be no GET requests to non existing objects.

frankyn: Thanks @dmitry-fa,

This example has the issue that `storage.create(blobInfo)` will create an empty object, meaning it will send a request. Given the existing performance issues with the Java library, it would be better to move these helpers under Storage instead of Blob to reduce the necessary requests.

```java
        BlobInfo blobInfo = BlobInfo.newBuilder(bucketName, blobName).build();
        // First upload request is made here.
        Blob blob = storage.create(blobInfo);
        // Second upload request is made here.
        blob.uploadFrom(somefile);
```

dmitry-fa: Thanks for explanation. I see your point now. You suggest using the `Storage.writer(BlobInfo, BlobWriteOption...)` method for the sake of performance. Agree, it makes sense. On the other hand, having uploadFrom/downloadTo methods in the Blob class will allow to reuse the same blob instance multiple times without recreating.
Perhaps, the right way would be to implement upload/download under storage, but keep existing methods in Blob.
I will think how to implement this.

frankyn: I'm not necessarily saying to use Storage.writer() in this case and instead referring to uploadFrom() being under Storage specifically. 

Additionally, as you recommended, I agree the existing helpers should stay where they are located in Blob. 

Thank you for your understanding @dmitry-fa

dmitry-fa: Thanks for your comments. I'll come up with an updated version.

dmitry-fa: Hi @frankyn,

You mentioned a very important aspect: performance. What I noticed, uploading stuff with WriteChannel is ~20% slower than direct upload. Storage provides only two methods for direct uploads:  `create(BlobInfo, byte[])` and `create(BlobInfo, InputStream)`.
The second one is declared as deprecated, which means not recommended to use.
I'm thinking about introducing `create(BolbInfo, Path)`.
I don't know what is the best way to implement
- use slower but reliable WriterChannel
- use faster but deprecated approach
- introduce two more methods `createLarge(BlobInfo, InputStream)/createLarge(BlobInfo, Path)` implemented over writer channels, and remove 'deprecated' label from `create(BlobInfo, InputStream)` with adding notes about risks.

Your advice would be much appreciated.

Thanks

dmitry-fa: @frankyn 

What I noticed, uploads based on `storage.writer(BlobInfo)` without creating a Blob instance work even faster than uploads which use deprecated `storage.create(InputStream)` .

Would it be okay, if I add a static method the Blob class to upload file without creating an instance and keep storage Storage interface as is? In this case we don't need to decide how to implement storage.create(File)

PR #127, opened 2020-02-12
feat: add new gapic generated client and protos for grpc storage client

BenWhitehead:
* New synth.py config necessary to generate gapic client and protos
* New maven config to add new modules:
  * grpc-google-cloud-storage-v1
  * proto-google-cloud-storage-v1
  * google-cloud-storage-bom

_NOTE_ This PR is targeting an integration branch we can work on independent from master

/cc @kolea2 @mfschwartz


google-cloud-storage-bom/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+  <artifactId>google-cloud-storage-bom</artifactId>
+  <version>1.103.2-SNAPSHOT</version><!-- {x-version-update:google-cloud-storage:current} -->
+  <packaging>pom</packaging>
+  <parent>
+    <groupId>com.google.cloud</groupId>
+    <artifactId>google-cloud-shared-config</artifactId>
+    <version>0.4.0</version>
+  </parent>
+
+  <name>Google Cloud Secret Manager BOM</name>
```
kolea2: rename

BenWhitehead: fixed

PR #139, opened 2020-02-20
Clarify documentation on date formats.

johnalowry:
I added the following code:

System.out.printf("Blob updatetime: %s", blob.getUpdateTime());

The output was

Blob updatetime: 1582157090238

I verified that this was ms after epoch:

$ date -d @1582157090
Wed 19 Feb 2020 04:04:50 PM PST

I also traced the code to this function:

https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/util/DateTime.java#L53

I assumed that create and delete times used the same format.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
@@ -795,17 +795,20 @@ public class BlobInfo implements Serializable {
     return metageneration;
   }
 
-  /** Returns the deletion time of the blob. */
+  /** Returns the deletion time of the blob expressed as the number of milliseconds since the Unix
+   * epoch. */
   public Long getDeleteTime() {
     return deleteTime;
   }
 
-  /** Returns the last modification time of the blob's metadata. */
+  /** Returns the last modification time of the blob's metadata expressed as the number of
+   * milliseconds since the Unix epoch. */
   public Long getUpdateTime() {
     return updateTime;
   }
 
-  /** Returns the creation time of the blob. */
+  /** Returns the creation time of the blob expressed as the number of milliseconds since the Unix
+   * epoch. */
   public Long getCreateTime() {
     return createTime;
   }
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -795,17 +795,18 @@ public Long getMetageneration() {
     return metageneration;
   }
 
-  /** Returns the deletion time of the blob. */
+  /** Returns the deletion time of the blob expressed as the number of ms since the Unix epoch. */
```
elharo: ms --> milliseconds
"Returns the deletion time of the blob" --> "Returns the time the blob was deleted"

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -795,17 +795,18 @@ public Long getMetageneration() {
...
   public Long getDeleteTime() {
     return deleteTime;
   }
 
-  /** Returns the last modification time of the blob's metadata. */
+  /** Returns the last modification time of the blob's metadata expressed as the number of ms since
```
elharo: milliseconds

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -795,17 +795,18 @@ public Long getMetageneration() {
...
+   * the Unix epoch. */
   public Long getUpdateTime() {
     return updateTime;
   }
 
-  /** Returns the creation time of the blob. */
+  /** Returns the creation time of the blob expressed as the number of ms since the Unix epoch. */
```
elharo: milliseconds

johnalowry:
I can't figure out how to rewrite the commit messages to conform to the expected format. I'll close out this PR and start again.

elharo:
And if you're like me, you never will. I haven't. Just ignore that check.

frankyn:
Sorry about that @johnalowry, thanks for sending a new one.

PR #140, opened 2020-02-20
feat: examples of creating a signed url for a blob with generation

### LINKED ISSUE #37: Cloud Storage: Signed URL determine generation from BlobInfo
For V2: URL returned by signURL() method doesn't include parameter part: `&generation=<number>`, one should add it manually.
For V4: there is no way to obtain URL pointing to not the latest version of an object, adding `&generation=<number>` doesn't help.

As part of cloud java-storage project only V2 problem could be fixed.

V4 problem also exists in nodejs-storage:  [Issue googleapis/google-cloud-java#953](https://github.com/googleapis/nodejs-storage/issues/953)
And very similar issue is [Issue googleapis/google-cloud-java#7044](https://github.com/googleapis/google-cloud-java/issues/7044)
It should be fixed on the server side.

Steps to reproduce:

Create a blob with two generation:

```
#> gsutil versioning set on gs://bucket_generation_signed
Enabling versioning for gs://bucket_generation_signed/...
#> echo The very first version > some.txt
#> gsutil cp some.txt gs://bucket_generation_signed
Copying file://some.txt [Content-Type=text/plain]...
/ [1 files][   23.0 B/   23.0 B]                                                
Operation completed over 1 objects/23.0 B.                                       
#> echo Updated version of the file > some.txt
#> gsutil cp some.txt gs://bucket_generation_signed
Copying file://some.txt [Content-Type=text/plain]...
/ [1 files][   28.0 B/   28.0 B]                                                
Operation completed over 1 objects/28.0 B.                                       
#> gsutil ls -a gs://bucket_generation_signed/some.txt
gs://bucket_generation_signed/some.txt#1576656755290328
gs://bucket_generation_signed/some.txt#1576656816500788
```
Run the following code:
```
    public static void main(String... args) throws Exception {
        List<String> al = Arrays.asList("https://www.googleapis.com/auth/cloud-platform");
        String googleCredentialsJson = new String(Files.readAllBytes(Paths.get("bucket_generation_signed.json")));
        GoogleCredentials credentials = GoogleCredentials
                .fromStream(new ByteArrayInputStream(googleCredentialsJson.getBytes())).createScoped(al);
        Storage storage = StorageOptions.newBuilder().setCredentials(credentials).build().getService();

        String bucketName = "bucket_generation_signed";
        String blobName = "some.txt";
        long generation = 1576656755290328L;

        BlobId blobId = BlobId.of(bucketName, blobName, generation);
        Blob blob = storage.get(blobId, Storage.BlobGetOption.generationMatch());
        System.out.println("blob: " + blob);
        System.out.println("v2 " + blob.signUrl(20, TimeUnit.MINUTES, Storage.SignUrlOption.withV2Signature()));
        System.out.println("v4 " + blob.signUrl(20, TimeUnit.MINUTES, Storage.SignUrlOption.withV4Signature()));
    }
```
The output will look like:
```
blob: Blob{bucket=bucket_generation_signed, name=some.txt, generation=1576656755290328, size=23, content-type=text/plain, metadata=null}

v2 https://storage.googleapis.com/bucket_generation_signed/some.txt?GoogleAccessId=my-service-account...uJWWFsyd9cfa4FrOSmw%3D%3D

v4 https://storage.googleapis.com/bucket_generation_signed/some.txt?X-Goog-Algorithm=GOOG4-RSA-SHA256&...076f845e952a7
```
Both returned URLs will point to the latest version:

Updated version of the file

 Despite the blob generation is explicitly specified: `generation=1576656755290328`

dmitry-fa:
Fixes #37

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -779,6 +779,13 @@ public class Blob extends BlobInfo {
    *     ServiceAccountCredentials.fromStream(new FileInputStream(keyPath))));
    * }</pre>
    *
+   * <p>Example of creating a signed URL for a blob generation.
+   *
+   * <pre>{@code
+   * URL signedUrl = blob.signUrl(1, TimeUnit.HOURS,
+   *     SignUrlOption.withQueryParams(ImmutableMap.of("generation", "1576656755290328")));
+   * }</pre>
+   *
    * @param duration time until the signed URL expires, expressed in {@code unit}. The finer
    *     granularity supported is 1 second, finer granularities will be truncated
    * @param unit time unit of the {@code duration} parameter
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -764,14 +764,14 @@ public class Blob extends BlobInfo {
    * </ol>
    *
    * <p>Example of creating a signed URL for the blob that is valid for 2 weeks, using the default
-   * credentials for signing the URL.
+   * credentials for signing the URL:
    *
    * <pre>{@code
    * URL signedUrl = blob.signUrl(14, TimeUnit.DAYS);
    * }</pre>
    *
    * <p>Example of creating a signed URL for the blob passing the {@link
-   * SignUrlOption#signWith(ServiceAccountSigner)} option, that will be used to sign the URL.
+   * SignUrlOption#signWith(ServiceAccountSigner)} option, that will be used to sign the URL:
    *
    * <pre>{@code
    * String keyPath = "/path/to/key.json";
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -779,6 +779,13 @@ public WriteChannel writer(BlobWriteOption... options) {
    *     ServiceAccountCredentials.fromStream(new FileInputStream(keyPath))));
    * }</pre>
    *
+   * <p>Example of creating a signed URL for a blob generation.
```
elharo: "generation." --> "generation:"

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2498,6 +2498,19 @@ Blob create(
    * <p>Note that the {@link ServiceAccountSigner} may require additional configuration to enable
    * URL signing. See the documentation for the implementation for more details.
    *
+   * <p>Example of creating a signed URL for a blob with generation.
```
elharo: colon

dmitry-fa: done, thanks for catching this.

PR #145, opened 2020-02-25
revert opencensus 0.25.0 to 0.24.0

frankyn:


frankyn:
Don't merge until we get consensus on version for opencensus.

elharo:
Decision was to stick with the version gRPC and http-java-client use, which, for now, is 0.24.0. If they update first, we'll update.

Issue #149, opened 2020-02-26
Blob.reload() does not work as intuitively expected


dmitry-fa:
Documentation for Blob.writer(BlobWriterOption...) recommends the following way to upload information:

```
 byte[] content = "Hello, World!".getBytes(UTF_8);
 try (WriteChannel writer = blob.writer()) {
   try {
     writer.write(ByteBuffer.wrap(content, 0, content.length));
   } catch (Exception ex) {
     // handle exception
   }
 }
```

But doc is silent on how to renew the blob object, which is a bit tricky because neither `blob.reload()` nor `storage.getBlob(blob.getBlobId())` does not help. Behavior varies depending on versioning setting for the bucket.

If versioning is enabled:

Attempt to reload as recommended:
`blob.reload(Blob.BlobSourceOption.generationNotMatch())` causes `StorageException '304 Not Modified'`

`blob.reload()`  returns the same blob.
`storage.get(blob.getBlobId())` returns the same blob (that seems very confusing).
And only `storage.get(blobId)` returns the updated blob.

If versioning is Suspended:
before upload: `blob.getContent()` returns an empty array
After upload: `blob.getContent()` throws `StorageException: 404 Not Found , No such object`
`storage.get(blob.getBlobId())`, `blob.reload()` and `blob.reload(Blob.BlobSourceOption.generationNotMatch())` return null.

And again only storage.get(blobId) returns the updated blob.

dmitry-fa:
For those who want to play around with it:
Setting versioning on a bucket with gsutil:
`#> gsutil versioning set on gs://bucket_versioning_on`
`#> gsutil versioning set off gs://bucket_versioning_off`

The code:
```
package storage;

import com.google.cloud.WriteChannel;
import com.google.cloud.storage.*;
import java.nio.ByteBuffer;

import static java.nio.charset.StandardCharsets.UTF_8;

public class Writer {

    public static void main(String... args) throws Exception {

        Storage storage = StorageOptions.getDefaultInstance().getService();
        String blobName = "my-blob-name";
        String bucketName = "bucket_versioning_off";

        BlobId blobId = BlobId.of(bucketName, blobName);
        byte[] content = "Hello, World!".getBytes(UTF_8);
        BlobInfo blobInfo = BlobInfo.newBuilder(blobId).setContentType("text/plain").build();
        Blob blob = storage.create(blobInfo);
        System.out.println("-- initial --");
        System.out.println(blob);
[...]

Issue #100, opened 2020-01-29
Blob.downloadTo() methods do not wrap RetryHelper$RetryHelperException


dmitry-fa:
Quoting source code for one of downloadTo methods:
```
  /**
   * Downloads this blob to the given file path using specified blob read options.
   *
   * @param path destination
   * @param options blob read options
   * @throws StorageException upon failure
   */
  public void downloadTo(Path path, BlobSourceOption... options) {
```
However `StorageException` is never throwing, but confusing `RetryHelper$RetryHelperException` is throwing instead:
```
        String bucketName = "my-bucket";
        String blobName = "my.txt";
        BlobInfo blobInfo = BlobInfo.newBuilder(bucketName, blobName).build();
        String key = "JVzfVl8NLD9FjedFuStegjRfES5ll5zc59CIXw572OA=";

        Blob blob1 = storage.create(blobInfo, Storage.BlobTargetOption.encryptionKey(key));
        blob1.writer().write(ByteBuffer.wrap("Hello".getBytes("UTF8")));

        Blob blob2 = storage.get(blobInfo.getBlobId());
        Path tempFileTo = Files.createTempFile("my_", ".tmp");
        try {
            blob2.downloadTo(tempFileTo);
        } catch (StorageException e) {
            // nice handling
        }
```
Output:
```
Exception in thread "main" com.google.cloud.RetryHelper$RetryHelperException: com.google.cloud.storage.StorageException: 400 Bad Request
The target object is encrypted by a customer-supplied encryption key.
    at com.google.cloud.RetryHelper.runWithRetries(RetryHelper.java:54)
    at com.google.cloud.storage.Blob.downloadTo(Blob.java:236)
    at com.google.cloud.storage.Blob.downloadTo(Blob.java:220)
    at com.google.cloud.storage.Blob.downloadTo(Blob.java:263)
   ...
```

dmitry-fa:
The root  cause:

Implementation relies on RetryHelper.runWithRetries()
```
  public void downloadTo(OutputStream outputStream, BlobSourceOption... options) {
    final CountingOutputStream countingOutputStream = new CountingOutputStream(outputStream);
    final StorageRpc storageRpc = this.options.getStorageRpcV1();
    final Map<StorageRpc.Option, ?> requestOptions = StorageImpl.optionMap(getBlobId(), options);
    runWithRetries(...);
  }
```

could be fixed by wrapping RetryHelper$RetryHelperException into StorageException:

```
try {
[...]

PR #162, opened 2020-03-04
feat: add blob.uploadfrom(inputstream)

### LINKED ISSUE #62: Add Blob.uploadFrom(InputStream)

dmitry-fa:
Fixes #62

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -269,19 +269,15 @@ public class Blob extends BlobInfo {
    * @param path file to be uploaded
    * @param options blob write options
    * @return updated blob
-   * @throws StorageException upon failure
+   * @throws IOException on I/O error
+   * @throws StorageException on failure
    */
-  public Blob uploadFrom(Path path, BlobWriteOption... options) {
-    if (!Files.exists(path)) {
-      throw new StorageException(0, path + ": No such file");
-    }
+  public Blob uploadFrom(Path path, BlobWriteOption... options) throws IOException {
     if (Files.isDirectory(path)) {
       throw new StorageException(0, path + ": Is a directory");
     }
     try (InputStream input = Files.newInputStream(path)) {
       return uploadFrom(input, options);
-    } catch (IOException e) {
-      throw new StorageException(e);
     }
   }
 
@@ -291,19 +287,18 @@ public class Blob extends BlobInfo {
    * @param input content to be uploaded
    * @param options blob write options
    * @return updated blob
-   * @throws StorageException upon failure
+   * @throws IOException on I/O error
+   * @throws StorageException on failure
    */
-  public Blob uploadFrom(InputStream input, BlobWriteOption... options) {
+  public Blob uploadFrom(InputStream input, BlobWriteOption... options) throws IOException {
     try (WriteChannel writer = storage.writer(this, options)) {
       upload(input, writer);
-    } catch (IOException e) {
-      throw new StorageException(e);
     }
     BlobId blobId = getBlobId();
     return storage.get(BlobId.of(blobId.getBucket(), blobId.getName()));
   }
 
-  static void upload(InputStream input, WriteChannel writer) {
+  static void upload(InputStream input, WriteChannel writer) throws IOException {
     upload(input, writer, DEFAULT_CHUNK_SIZE);
   }
 
@@ -321,23 +316,23 @@ public class Blob extends BlobInfo {
    * <pre>{@code
    * Path file = Paths.get("humongous.file");
    * try (InputStream input = Files.newInputStream(file); WriteChannel writer = storage.writer(blobInfo)) {
-   *     Blob.upload(input, writer, 150 * 1024 * 1024);
+   *   Blob.upload(input, writer, 150 * 1024 * 1024);
+   * } catch (IOException e) {
+   *   // your handler
    * }
    *
    * @param input content to be uploaded
    * @param writer channel
    * @param bufferSize size of the buffer to read from input and send over writer
+   * @throws IOException on I/O error
    */
-  public static void upload(InputStream input, WriteChannel writer, int bufferSize) {
+  public static void upload(InputStream input, WriteChannel writer, int bufferSize)
+      throws IOException {
     bufferSize = Math.max(bufferSize, MIN_BUFFER_SIZE);
-    try {
-      byte[] buffer = new byte[bufferSize];
-      int length;
-      while ((length = input.read(buffer)) >= 0) {
-        writer.write(ByteBuffer.wrap(buffer, 0, length));
-      }
-    } catch (IOException e) {
-      throw new StorageException(e);
+    byte[] buffer = new byte[bufferSize];
+    int length;
+    while ((length = input.read(buffer)) >= 0) {
+      writer.write(ByteBuffer.wrap(buffer, 0, length));
     }
   }
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -266,7 +266,7 @@ public class Blob extends BlobInfo {
   /**
    * Uploads the given file path to this blob using specified blob write options.
    *
-   * @param path file to be uploaded
+   * @param path file to upload
    * @param options blob write options
    * @return updated blob
    * @throws IOException on I/O error
@@ -274,7 +274,7 @@ public class Blob extends BlobInfo {
    */
   public Blob uploadFrom(Path path, BlobWriteOption... options) throws IOException {
     if (Files.isDirectory(path)) {
-      throw new StorageException(0, path + ": Is a directory");
+      throw new StorageException(0, path + " is a directory");
     }
     try (InputStream input = Files.newInputStream(path)) {
       return uploadFrom(input, options);
@@ -284,7 +284,7 @@ public class Blob extends BlobInfo {
   /**
    * Uploads the given content to this blob using specified blob write options.
    *
-   * @param input content to be uploaded
+   * @param input content to upload
    * @param options blob write options
    * @return updated blob
    * @throws IOException on I/O error
@@ -309,14 +309,13 @@ public class Blob extends BlobInfo {
 
   /**
    * Uploads the given content to the storage using specified write channel and the given buffer
-   * size. Other uploadFrom() methods invokes this one with the buffer size of 15 MiB, users may
-   * pass alternative values. Larger buffer sizes may improve the upload performance but require
-   * more memory. It could cause OutOfMemoryError or add significant garbage collection overhead.
-   * Smaller buffer sizes will reduce memory consumption, that will be noticeably in case of
-   * parallel uploads of many objects. Buffer sizes which are less than 256 KiB are not allowed,
-   * they will be treated as 256 KiB.
+   * size. Other uploadFrom() methods invoke this one with a buffer size of 15 MiB, users can pass
+   * alternative values. Larger buffer sizes might improve the upload performance but require more
+   * memory. This can cause an OutOfMemoryError or add significant garbage collection overhead.
+   * Smaller buffer sizes reduce memory consumption, that is noticeable when uploading many objects
+   * in parallel. Buffer sizes less than 256 KiB are treated as 256 KiB.
    *
-   * <p>This method does not close either InputStream or WriterChannel
+   * <p>This method does not close either the InputStream or the WriterChannel
    *
    * <p>Example of uploading:
    *
@@ -325,13 +324,13 @@ public class Blob extends BlobInfo {
    * BlobInfo blobInfo = BlobInfo.newBuilder(blobId).setContentType("video/webm").build();
    * Path file = Paths.get("humongous.file");
    * try (InputStream input = Files.newInputStream(file); WriteChannel writer = storage.writer(blobInfo)) {
-   *   Blob.upload(input, writer, 150 * 1024 * 1024);
+   *   Blob.uploadFrom(input, writer, 150 * 1024 * 1024);
    * } catch (IOException e) {
    *   // your handler
    * }
    * }</pre>
    *
-   * @param input content to be uploaded
+   * @param input content to upload
    * @param writer channel
    * @param bufferSize size of the buffer to read from input and send over writer
    * @throws IOException on I/O error
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -38,9 +38,11 @@ import com.google.common.base.Function;
 import com.google.common.io.BaseEncoding;
 import com.google.common.io.CountingOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.OutputStream;
 import java.net.URL;
+import java.nio.ByteBuffer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.Key;
@@ -73,7 +75,7 @@ public class Blob extends BlobInfo {
         }
       };
 
-  private static final int DEFAULT_CHUNK_SIZE = 2 * 1024 * 1024;
+  private static final int DEFAULT_CHUNK_SIZE = 15 * 1024 * 1024;
 
   /** Class for specifying blob source options when {@code Blob} methods are used. */
   public static class BlobSourceOption extends Option {
@@ -258,6 +260,84 @@ public class Blob extends BlobInfo {
    */
   public void downloadTo(Path path) {
     downloadTo(path, new BlobSourceOption[0]);
+  }
+
+  /**
+   * Uploads the given file path to this blob using specified blob write options.
+   *
+   * @param path file to be uploaded
+   * @param options blob write options
+   * @return updated blob
+   * @throws StorageException upon failure
+   */
+  public Blob uploadFrom(Path path, BlobWriteOption... options) {
+    if (!Files.exists(path)) {
+      throw new StorageException(0, path + ": No such file");
+    }
+    if (Files.isDirectory(path)) {
+      throw new StorageException(0, path + ": Is a directory");
+    }
+    try (InputStream input = Files.newInputStream(path)) {
+      return uploadFrom(input, options);
+    } catch (IOException e) {
+      throw new StorageException(e);
+    }
+  }
+
+  /**
+   * Uploads the given content to this blob using specified blob write options.
+   *
+   * @param input content to be uploaded
+   * @param options blob write options
+   * @return updated blob
+   * @throws StorageException upon failure
+   */
+  public Blob uploadFrom(InputStream input, BlobWriteOption... options) {
+    try (WriteChannel writer = storage.writer(this, options)) {
+      upload(input, writer);
+    } catch (IOException e) {
+      throw new StorageException(e);
+    }
+    BlobId blobId = getBlobId();
+    return storage.get(BlobId.of(blobId.getBucket(), blobId.getName()));
+  }
+
+  static void upload(InputStream input, WriteChannel writer) {
+    upload(input, writer, DEFAULT_CHUNK_SIZE);
+  }
+
+  /**
+   * Uploads the given content to the storage using specified write channel and the given
+   * buffer size. The default buffer size is 15 MB. Larger buffer sizes may improve the upload
+   * performance but require more memory. It could cause OutOfMemoryError or add significant garbage
+   * collection overhead. Buffer sizes which are less than 256 KB are not allowed, they will be
+   * treated as 256 KB.
+   *
+   * <p>Note that this method does not close neither InputStream nor WriterChannel</p>
+   *
+   * <p>Example of uploading:
+   *
+   * <pre>{@code
+   * Path file = Paths.get("humongous.file");
+   * try (InputStream input = Files.newInputStream(file); WriteChannel writer = storage.writer(blobInfo)) {
+   *     Blob.upload(input, writer, 150 * 1024 * 1024);
+   * }
+   *
+   * @param input content to be uploaded
+   * @param writer channel
+   * @param bufferSize size of the buffer to read from input and send over writer
+   */
+  public static void upload(InputStream input, WriteChannel writer, int bufferSize) {
+    bufferSize = Math.max(bufferSize, 262144);
+    try {
+      byte[] buffer = new byte[bufferSize];
+      int length;
+      while ((length = input.read(buffer)) >= 0) {
+        writer.write(ByteBuffer.wrap(buffer, 0, length));
+      }
+    } catch (IOException e) {
+      throw new StorageException(e);
+    }
   }
 
   /** Builder for {@code Blob}. */
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -76,6 +76,7 @@ public class Blob extends BlobInfo {
       };
 
   private static final int DEFAULT_CHUNK_SIZE = 15 * 1024 * 1024;
+  private static final int MIN_BUFFER_SIZE = 256 * 1024;
 
   /** Class for specifying blob source options when {@code Blob} methods are used. */
   public static class BlobSourceOption extends Option {
@@ -308,12 +309,12 @@ public class Blob extends BlobInfo {
 
   /**
    * Uploads the given content to the storage using specified write channel and the given
-   * buffer size. The default buffer size is 15 MB. Larger buffer sizes may improve the upload
+   * buffer size. The default buffer size is 15 MiB. Larger buffer sizes may improve the upload
    * performance but require more memory. It could cause OutOfMemoryError or add significant garbage
-   * collection overhead. Buffer sizes which are less than 256 KB are not allowed, they will be
-   * treated as 256 KB.
+   * collection overhead. Buffer sizes which are less than 256 KiB are not allowed, they will be
+   * treated as 256 KiB.
    *
-   * <p>Note that this method does not close neither InputStream nor WriterChannel</p>
+   * <p>This method does not close either InputStream or WriterChannel</p>
    *
    * <p>Example of uploading:
    *
@@ -328,7 +329,7 @@ public class Blob extends BlobInfo {
    * @param bufferSize size of the buffer to read from input and send over writer
    */
   public static void upload(InputStream input, WriteChannel writer, int bufferSize) {
-    bufferSize = Math.max(bufferSize, 262144);
+    bufferSize = Math.max(bufferSize, MIN_BUFFER_SIZE);
     try {
       byte[] buffer = new byte[bufferSize];
       int length;
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -309,13 +309,13 @@ public class Blob extends BlobInfo {
 
   /**
    * Uploads the given content to the storage using specified write channel and the given buffer
-   * size. Other uploadFrom() methods invoke this one with a buffer size of 15 MiB, users can pass
+   * size. Other uploadFrom() methods invoke this one with a buffer size of 15 MiB. Users can pass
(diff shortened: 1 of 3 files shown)


dmitry-fa:
This is the first part of the feature.

I'm going to add similar functionality to the Storage interface as a separate pull request:

```
public Blob create(BlobInfo blobInfo, Path path, BlobWriteOption... options);
```

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +262,84 @@ public void downloadTo(Path path) {
...
+    if (!Files.exists(path)) {
+      throw new StorageException(0, path + ": No such file");
+    }
+    if (Files.isDirectory(path)) {
+      throw new StorageException(0, path + ": Is a directory");
+    }
+    try (InputStream input = Files.newInputStream(path)) {
+      return uploadFrom(input, options);
+    } catch (IOException e) {
+      throw new StorageException(e);
```
elharo: IOException should be thrown directly here. It should not be converted to a runtime exception (and StorageException should probably not be a runtime exception)

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +262,84 @@ public void downloadTo(Path path) {
...
+   * @param input content to be uploaded
+   * @param options blob write options
+   * @return updated blob
+   * @throws StorageException upon failure
+   */
+  public Blob uploadFrom(InputStream input, BlobWriteOption... options) {
+    try (WriteChannel writer = storage.writer(this, options)) {
+      upload(input, writer);
+    } catch (IOException e) {
+      throw new StorageException(e);
```
elharo: runtime exception is not appropriate

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +262,84 @@ public void downloadTo(Path path) {
...
+  }
+
+  /**
+   * Uploads the given content to the storage using specified write channel and the given
+   * buffer size. The default buffer size is 15 MB. Larger buffer sizes may improve the upload
+   * performance but require more memory. It could cause OutOfMemoryError or add significant garbage
+   * collection overhead. Buffer sizes which are less than 256 KB are not allowed, they will be
+   * treated as 256 KB.
+   *
+   * <p>Note that this method does not close neither InputStream nor WriterChannel</p>
```
elharo: delete "Note that"

double negative; neither nor --> either or

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +262,84 @@ public void downloadTo(Path path) {
...
+   * try (InputStream input = Files.newInputStream(file); WriteChannel writer = storage.writer(blobInfo)) {
+   *     Blob.upload(input, writer, 150 * 1024 * 1024);
+   * }
+   *
+   * @param input content to be uploaded
+   * @param writer channel
+   * @param bufferSize size of the buffer to read from input and send over writer
+   */
+  public static void upload(InputStream input, WriteChannel writer, int bufferSize) {
+    bufferSize = Math.max(bufferSize, 262144);
```
elharo: why 262144? This magic number could use a named constant for clarity.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +262,84 @@ public void downloadTo(Path path) {
...
+    try {
+      byte[] buffer = new byte[bufferSize];
+      int length;
+      while ((length = input.read(buffer)) >= 0) {
+        writer.write(ByteBuffer.wrap(buffer, 0, length));
+      }
+    } catch (IOException e) {
+      throw new StorageException(e);
```
elharo: bubble the IOException

google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
```diff
@@ -662,4 +668,111 @@ public Long answer() throws Throwable {
...
+  public void testUploadFromStream() throws Exception {
+    final byte[] dataToSend = {1, 2, 3, 4, 5};
+    WriteChannel channel = createWriteChannelMock(dataToSend);
+    initializeExpectedBlob(1);
+    BlobId blobId = BlobId.of(BLOB_INFO.getBucket(), BLOB_INFO.getName());
+    expect(storage.getOptions()).andReturn(mockOptions);
+    expect(storage.writer(eq(expectedBlob))).andReturn(channel);
+    expect(storage.get(blobId)).andReturn(expectedBlob);
+    replay(storage);
+    blob = new Blob(storage, new BlobInfo.BuilderImpl(BLOB_INFO));
```
elharo: This has shown up three times; just initialize the field

google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
```diff
@@ -662,4 +668,111 @@ public Long answer() throws Throwable {
...
+    InputStream input = new ByteArrayInputStream(dataToSend);
+    blob = blob.uploadFrom(input);
+    assertSame(expectedBlob, blob);
+  }
+
+  @Test
+  public void testUpload() throws Exception {
+    replay(storage);
+    final byte[] dataToSend = {1, 2, 3, 4, 5};
```
elharo: final not needed (and thus not allowed, per Google style)

google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
```diff
@@ -662,4 +668,111 @@ public Long answer() throws Throwable {
...
+    final byte[] dataToSend = new byte[100_000];
+    WriteChannel channel = createWriteChannelMock(dataToSend);
+    InputStream input = new ByteArrayInputStream(dataToSend);
+    Blob.upload(input, channel, 100);
+  }
+
+  @Test
+  public void testUploadMultiplePortions() throws Exception {
+    replay(storage);
+    final int TOTAL_SIZE = 400_000;
```
elharo: these are local variables, not constants

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -2939,4 +2941,61 @@ public void testBucketLogging() throws ExecutionException, InterruptedException
       RemoteStorageHelper.forceDelete(storage, loggingBucket, 5, TimeUnit.SECONDS);
     }
   }
+
+  @Test
+  public void testUpload() throws Exception {
+    String blobName = "test-upload-static";
+    BlobId blobId = BlobId.of(BUCKET, blobName);
+    byte stringBytes[] = BLOB_STRING_CONTENT.getBytes(UTF_8);
```
elharo: pull repeated code like this up to a field

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -2939,4 +2941,61 @@ public void testBucketLogging() throws ExecutionException, InterruptedException
...
+    Path tempFileFrom = Files.createTempFile("ITStorageTest_", ".tmp");
+    Files.write(tempFileFrom, stringBytes);
+    Blob blob = storage.create(blobInfo);
+    blob = blob.uploadFrom(tempFileFrom, Storage.BlobWriteOption.encryptionKey(KEY));
+
+    Path tempFileTo = Files.createTempFile("ITStorageTest_", ".tmp");
+    try {
+      blob.downloadTo(tempFileTo);
+    } catch (Exception e) {
```
elharo: never catch Exception, only catch what you expect; especially in tests

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: fixed

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: MIN_BUFFER_SIZE constant added

google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
dmitry-fa: fixed

google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
dmitry-fa: fixed

google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
dmitry-fa: fixed

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
dmitry-fa: Exception has been replaced with RetryHelper.RetryHelperException. As part of fix of #100 it should turn into StorageException.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
dmitry-fa: Fixed with using BLOB_BYTE_CONTENT constant.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: Thanks for looking at this PR.
I agree with all your comments, except ones related to IOException handling.
There are many places in the storage code where it is wraping to StorageException, that allows user to have a single handler for all kind of exceptions.  I'm not sure that IOException deserves a dedicated handler.

@frankyn  recommended extending Storage interface with upload functionality. I'm going to implement this in a separate PR.
Storage already has `create(BlobInfo, InputStream)` method which doesn't throw IOException, so adding `create(BlobInfo, Path)` which throws IOException in addition to StorageException will not look logically. I'm afraid people will write their own utility methods to wrap unnecessary exceptions.

elharo: StorageException is a huge mistake. It should never have been a runtime exception. Consistency is not a sufficient reason to propagate this broken design to new code.

dmitry-fa: Throwing IOException for a few methods will not fix the problem with existing design but make it only worse.
I agree,  BaseServiceException should be checked for better design. Does it mean that we cannot extend functionality unless https://github.com/googleapis/java-core/issues/57 is fixed?

elharo: It won't fix the existing design, but it won't make the problem worse. Adding new methods that also make this mistake will make the problem worse.

dmitry-fa: fixed

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: fixed

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,79 @@ public void downloadTo(Path path) {
...
+  }
+
+  static void upload(InputStream input, WriteChannel writer) throws IOException {
+    upload(input, writer, DEFAULT_CHUNK_SIZE);
+  }
+
+  /**
+   * Uploads the given content to the storage using specified write channel and the given
+   * buffer size. The default buffer size is 15 MiB. Larger buffer sizes may improve the upload
+   * performance but require more memory. It could cause OutOfMemoryError or add significant garbage
```
elharo: Do we really need to give the user the option to set the buffer size? Do we provide sufficient tools and documentation to enable someone to do this?

It would be better to simply pick the best buffer size directly. If we're uncertain, lets leave this method out until we are, since it's easier to add it in later than take it out.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,79 @@ public void downloadTo(Path path) {
...
+   * collection overhead. Buffer sizes which are less than 256 KiB are not allowed, they will be
+   * treated as 256 KiB.
+   *
+   * <p>This method does not close either InputStream or WriterChannel</p>
```
elharo: These details can be moved to the previous method.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: I think we really need to give the user the option to play with the buffer size. There is no optimal value for the buffer size. By default, the buffer size is equal to the default Write Channel chunk size (15 MiB), but for uploading large files increasing the buffer will give significant performance benefit. Some statistics of uploading 1.5GB file with various buffer sizes:
```
15MiB:  162 seconds
30MiB:  149 seconds
75MiB:  141 seconds
100MiB: 140 seconds
150MiB: 139 seconds
300MiB: 138 seconds
```
On the other hand, one may need to upload hundreds file in parallel, this may require reducing the buffer size to avoid running out of memory. So, giving such flexibility will be certainly a plus.

> Do we provide sufficient tools and documentation to enable someone to do this?

I'm afraid no. There was a request to provide a sort of such documentation: https://github.com/googleapis/java-storage/issues/40 which is closed as duplicate of the current one (add Blob.uploadFrom())

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: The previous one is intentionally made package private, it will be invoked from the StorageImpl class.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
elharo: In that case this PR needs to include such documentation. I've seen from experience that tunable I/O buffer size parameters are very effective foot-seeking missiles. I'd estimate that somewhere above 80% of invocations of this method will decrease performance.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
elharo: I'm confused about the difference between upload and uploadFrom, so probably users will be too. Stocking with the public methods, I don't see any particular difference between the names `upload` and `uploadFrom`. I suggest making both `upload`.

Also, why does one return a blob and one doesn't. Both should or neither. 

Since I am confused, you might have good reasons for this distinction, but in that case let's make sure the docs and the method names make these reasons obvious.

dmitry-fa: Returning a Blob instance after upload has completed requires an extra GET request which is not always needed. So, `upload` works faster than `uploadFrom`, especially on small files.
I'm going to document this difference in the new method: `Storage.upload(BlobInfo, Path)`

`uploadFrom` is named after `downloadTo` (as requested in the issue).
`upload` is static and its name is different. Would it be more clear to use `writeContnet` instead?

```
/**
  * Writes the given content to the storage using specified write channel and the given
  * buffer size.
  */
public static void Blob.writeContent(InputStream content, WriterChannel writer, int bufferSize)
```

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: OK, I'll add notice, that this method is for those who want to experiment with the upload time and the amount of the consumed memory.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
elharo: I'm not sure.The distinction between the names uploadFrom and upload seems very hazy. Honestly, if I had to guess I'd expect uploadFrom was static rather than the other way around.

Do we actually need both of these? Can we delete the one that makes an extra GET request?

Consider this case:

1. First request to upload a blob succeeds.
2. GET request fails.
3. Exception thrown

It won't be obvious to the user that the content was successfully upload to the server.

dmitry-fa: Yes, this is a good example, quite rare but still valid.
To inform user of successful upload I re-throw an exception with the attendant message.

> Do we actually need both of these? Can we delete the one that makes an extra GET request?

Yes, I believe we need both. If  `blob.uploadFrom()` returns `void`, then `blob` instance will become useless right after upload.  As  #149 demonstrates, it's not obvious how to get the updated blob object, so returning the updated updated would be helpful.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
     downloadTo(path, new BlobSourceOption[0]);
   }
 
+  /**
+   * Uploads the given file path to this blob using specified blob write options.
+   *
+   * @param path file to be uploaded
```
elharo: to be uploaded --> to upload

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+   * @param options blob write options
+   * @return updated blob
+   * @throws IOException on I/O error
+   * @throws StorageException on failure
+   */
+  public Blob uploadFrom(Path path, BlobWriteOption... options) throws IOException {
+    if (Files.isDirectory(path)) {
+      throw new StorageException(0, path + ": Is a directory");
```
elharo: nit: path +" is a directory"

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
     downloadTo(path, new BlobSourceOption[0]);
   }
 
+  /**
+   * Uploads the given file path to this blob using specified blob write options.
```
elharo: I probably just don't understand how this API works, but this sounds off to me. I'd expect a static method that uploads the path to a new Blob and returns it. What does it mean to "upload to this blob"? I'm missing something here.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+    BlobId blobId = getBlobId();
+    try {
+      return storage.get(BlobId.of(blobId.getBucket(), blobId.getName()));
+    } catch (StorageException e) {
+      throw new StorageException(
+          e.getCode(), "Content has been uploaded successfully. Failed to retrieve blob.", e);
+    }
+  }
+
+  static void uploadFrom(InputStream input, WriteChannel writer) throws IOException {
```
elharo: cant his be private?

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+   * <p>This method does not close either InputStream or WriterChannel
+   *
+   * <p>Example of uploading:
+   *
+   * <pre>{@code
+   * BlobId blobId = BlobId.of(bucketName, blobName);
+   * BlobInfo blobInfo = BlobInfo.newBuilder(blobId).setContentType("video/webm").build();
+   * Path file = Paths.get("humongous.file");
+   * try (InputStream input = Files.newInputStream(file); WriteChannel writer = storage.writer(blobInfo)) {
+   *   Blob.upload(input, writer, 150 * 1024 * 1024);
```
elharo: `upload` here but this method is `uploadFrom`

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+    }
+  }
+
+  static void uploadFrom(InputStream input, WriteChannel writer) throws IOException {
+    uploadFrom(input, writer, DEFAULT_CHUNK_SIZE);
+  }
+
+  /**
+   * Uploads the given content to the storage using specified write channel and the given buffer
+   * size. Other uploadFrom() methods invokes this one with the buffer size of 15 MiB, users may
```
elharo: a buffer size. Users can

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+   * pass alternative values. Larger buffer sizes may improve the upload performance but require
```
elharo: may --> might or can

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+   * more memory. It could cause OutOfMemoryError or add significant garbage collection overhead.
```
elharo: It could cause --> This can cause an

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+   * Smaller buffer sizes will reduce memory consumption, that will be noticeably in case of
```
elharo: will reduce --> reduce (tech writing lives in the eternal present)
noticeably in case of parallel... --> is noticeable when uploading many objects in parallel.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+   * parallel uploads of many objects. Buffer sizes which are less than 256 KiB are not allowed,
```
elharo: which are less than --> less than

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+   * they will be treated as 256 KiB.
```
elharo: run-on sentence

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,89 @@ public void downloadTo(Path path) {
...
+   *
+   * <p>This method does not close either InputStream or WriterChannel
```
elharo: either the

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: The storage hierarchy is the following:
Class **BlobId**: A BlobId object includes the name of the containing bucket, the blob's name and possibly the blob's generation. If getGeneration() is null the identifier refers to the latest blob's generation.

Class **BlobInfo**: encapsulate **BlobId** and set of properties.

Class **Blob** extends **BlobInfo** and adds a number of methods to modify the blob like update, copyTo, downloadTo, etc. Objects of this class are immutable. Operations that modify the blob return a new object.

New `Blob.uploadFrom()` methods will allow those who already have a blob instance to upload some content to it as simply as `blob = blob.uploadFrom(myFile);`

> I'd expect a static method that uploads the path to a new Blob and returns it.

A static method would require an extra parameter `storage`, which is not very convenient.
As I said previously I'm going to extend the Storage interface with new methods:
```
public Blob create(BlobInfo blobInfo, Path path, BlobWriteOption... options);
```

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: this method will be invoked from the StorageImpl class

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: @elharo Thanks a lot for your comments! I very much appreciate it.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
elharo: This is what's throwing me: the object is immutable but has methods to modify it? Why?

This seems to be the existing API, but it's still extremely wonky. After you call uploadFrom (or downloadTo) how does the new Blob bject returned differ from the original Blob object?

dmitry-fa: You can consider a Blob instance like a pointer to {bucket, name, generation} and methods to operate with that pointer. Each modification of the blob causes the generation change, so existing pointer becomes outdated and the Blob instance useless(Immutability does not allow to modify the generation in place). A new object returned by uploadFrom has the new generation value pointing to the latest blob version.

elharo: It would be nice if the docs said that, or the classes were named accurately. but c'est la vie.

https://github.com/googleapis/java-storage/issues/176

My immediate question is what exactly has changed in the object? What field has been modified and how?

dmitry-fa: new Blob objects are constructed from data received from GCS.
In case of uploadFrom only 'generation' and 'size' will be changed. (These fields are inherited from  the BlobInfo class.)
uploadFrom  uses WriteChannel, so it has to perform a separate GET request to obtain the fresh data.
upload method performs the RPC request and obtain the data to construct new object from the response.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,88 @@ public void downloadTo(Path path) {
...
+    }
+  }
+
+  static void uploadFrom(InputStream input, WriteChannel writer) throws IOException {
+    uploadFrom(input, writer, DEFAULT_CHUNK_SIZE);
+  }
+
+  /**
+   * Uploads the given content to the storage using specified write channel and the given buffer
+   * size. Other uploadFrom() methods invoke this one with a buffer size of 15 MiB, users can pass
```
elharo: 15MiB. Users

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,88 @@ public void downloadTo(Path path) {
...
+   * alternative values. Larger buffer sizes might improve the upload performance but require more
+   * memory. This can cause an OutOfMemoryError or add significant garbage collection overhead.
+   * Smaller buffer sizes reduce memory consumption, that is noticeable when uploading many objects
+   * in parallel. Buffer sizes less than 256 KiB are treated as 256 KiB.
+   *
+   * <p>This method does not close either the InputStream or the WriterChannel
```
elharo: period at end

dmitry-fa: fixed

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: fixed

PR #163, opened 2020-03-04
fix: use %s instead of %d format specifier in checkArgument

### LINKED ISSUE #159: Using %d instead of %s for formating in checkArgument
There are currently being used %d instead of %s in template string to checkArguments in
https://github.com/googleapis/java-storage/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L692

This makes the output from precondition failure come out awkwardly.

I guess someone with more routine in getting changes into this codebase can make this fix faster than me.

athakor:
Fixes #159

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -689,7 +689,7 @@ public class HttpStorageRpc implements StorageRpc {
     Span span = startSpan(HttpStorageRpcSpans.SPAN_NAME_READ);
     Scope scope = tracer.withSpan(span);
     try {
-      checkArgument(position >= 0, "Position should be non-negative, is %d", position);
+      checkArgument(position >= 0, "Position should be non-negative, is %s", position);
       Get req = createReadRequest(from, options);
       StringBuilder range = new StringBuilder();
       range.append("bytes=").append(position).append("-").append(position + bytes - 1);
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -689,7 +689,7 @@ public long read(
     Span span = startSpan(HttpStorageRpcSpans.SPAN_NAME_READ);
     Scope scope = tracer.withSpan(span);
     try {
-      checkArgument(position >= 0, "Position should be non-negative, is %d", position);
+      checkArgument(position >= 0, "Position should be non-negative, is %s", position);
```
elharo: I can't tell the difference between these two. %d looks correct to me. How about we use string concatenation with a plus sign here instead to avoid the issue.

I do notice this should probably not be inside the try block.

runevalle-cognite: %d had been correct with standard formater, but is incorrect when using guava's Precondition library formatting. (You can check documentation and code linked in the associated issue)

elharo: The mere fact that we have to check documentation and that this is not obvious strongly suggests that we shouldn't use a format string at all. Just use string concatenation and call it a day.

athakor: Done. I have used string concatenation `+` instead of `%s`. @elharo PTAL

Issue #179, opened 2020-03-13
add Storage.upload(Path)


dmitry-fa:
In addition to Blob.uploadFrom() methods which require a Blob instance to be created first there should be a way to upload a resource to the storage without creating an instance of Blob:
 
```
interface Storage {
    void upload(BlobInfo blobInfo, Path path, BlobWriteOption... options);
    void upload(BlobInfo blobInfo, InputStream content, BlobWriteOption... options);
}
```

Implementing this feature means extending of public interface.

dmitry-fa:
Extension the Storage interface is incompatible change. Such a minor feature as adding helper upload methods should not break compatibility. Instead of extending the main interface it's better to introduce a new class like StorageUtilities and put various helpers there.

frankyn:
How is Storage interface change incompatible change in this case? I still think we should add the method under Storage and not create a separate helper utility.

dmitry-fa:
Interface extension is backward incompatible change.

frankyn:
We've been adding new methods to the interface for several features outside of this request. Could you clarify what you mean? 

If we move forward with the feature added in #214 it will create a fracture in developer experience. We can try to do something else, but I'd prefer `Storage.upload()` over a separate helper class. 

cc: @JesseLovelace @elharo @crwilcox

dmitry-fa:
Extending interface breaks API and requires a major version update since this API post 1.0.
Adding upload functionality is rather a small feature it just add a few helper methods. I don't think it deserves a major version update. The helper class will allow to add small feature like this without breaking API.

frankyn:
I agree that this doesn't substantiate a major version bump.
I don't agree that we should do this as a separate utility class. It is a fracture in the experience. 

What is the failure that you're hitting by updating the Storage API surface?

dmitry-fa:
What about introducing new `StorageIOException` extending `IOException`?

elharo:
You can't add a method to an interface without breaking the API.

frankyn:
I misunderstood. damn, we have breaking the API without major version bumps for a couple of years then. 

What other alternatives do we have to continue supporting requests in Storage.java? It really changes the way we execute work so I'm a bit confused.

elharo:
There are at least 3 ways forward:

1. Use concrete classes for new functionality.
2. Add a new subinterface.
3. Increment the major version.

There's a fourth once we drop Java 7 support. Maybe some others I haven't thought of.

frankyn:
[...]

PR #190, opened 2020-03-17
Revert "feat: add blob.uploadfrom(inputstream)"

frankyn:
Reverts googleapis/java-storage#162

Need to revert this for now. The issue is that the [Blob constructor is private](https://github.com/googleapis/java-storage/blob/1f53baa969331a94b5a73319df59711157ef2307/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java#L545) and it requires creating the object as an [empty object before uploading with an InputStream](https://github.com/googleapis/java-storage/pull/162/files#diff-8c18a40b13cc5c5d90547c273470f5eeR2965-R2966).

The implementation should be derived from the Storage class instead of Blob class to reduce the number of requests.

cc: @dmitry-fa, could you recreate the PR? I haven't prioritized this, apologies for the delay.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -38,11 +38,9 @@ import com.google.common.base.Function;
 import com.google.common.io.BaseEncoding;
 import com.google.common.io.CountingOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.OutputStream;
 import java.net.URL;
-import java.nio.ByteBuffer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.Key;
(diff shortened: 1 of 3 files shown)


dmitry-fa:
@frankyn ,

#179 will add upload method to Storage (without a Blob instance creating)

About #162: you agreed to leave upload methods in the Blob class:
> Additionally, as you recommended, I agree the existing helpers should stay where they are located in Blob.

https://github.com/googleapis/java-storage/pull/116#discussion_r379606946

Taking into account #149 upload methods returning the actual Blob instance will be helpful.

Would it make sense to re-integrated #162?

#179 requires public interface extension, so I decided to separate the feature.

frankyn:
> About #162: you agreed to leave upload methods in the Blob class:
>> Additionally, as you recommended, I agree the existing helpers should stay where they are located in Blob.

I should have been more specific, and called out Blob download helpers and *not* upload helpers. 

Please support Storage level upload helpers only.

PR #205, opened 2020-03-23
feat: add storage.upload(path)

### LINKED ISSUE #179: add Storage.upload(Path)

dmitry-fa:
Fixes #179

### DIFF
--- a/google-cloud-storage/clirr-ignored-differences.xml
+++ b/google-cloud-storage/clirr-ignored-differences.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/storage/Storage</className>
+    <method>void upload(*.BlobInfo, java.nio.file.Path, *.Storage$BlobWriteOption[])</method>
+  </difference>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/storage/Storage</className>
+    <method>void upload(*.BlobInfo, java.io.InputStream, *.Storage$BlobWriteOption[])</method>
+  </difference>
+</differences>
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -38,9 +38,11 @@ import com.google.common.base.Function;
 import com.google.common.io.BaseEncoding;
 import com.google.common.io.CountingOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.OutputStream;
 import java.net.URL;
+import java.nio.ByteBuffer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.Key;
@@ -73,7 +75,8 @@ public class Blob extends BlobInfo {
         }
       };
 
-  private static final int DEFAULT_CHUNK_SIZE = 2 * 1024 * 1024;
+  private static final int DEFAULT_BUFFER_SIZE = 15 * 1024 * 1024;
+  private static final int MIN_BUFFER_SIZE = 256 * 1024;
 
   /** Class for specifying blob source options when {@code Blob} methods are used. */
   public static class BlobSourceOption extends Option {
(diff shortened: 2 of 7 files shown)


dmitry-fa:
Note: This PR extends the public interface.

google-cloud-storage/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- see http://www.mojohaus.org/clirr-maven-plugin/examples/ignored-differences.html -->
+<differences>
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/storage/Storage</className>
+    <method>void upload(*.BlobInfo, java.nio.file.Path, *.Storage$BlobWriteOption[])</method>
```
elharo: An incompatible change like this requires a major version update since this API post 1.0

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,48 @@ public void downloadTo(Path path) {
...
+   *   // your handler
+   * }
+   * }</pre>
+   *
+   * @param input content to upload
+   * @param writer channel
+   * @param bufferSize size of the buffer to read from input and send over writer
+   * @throws IOException on I/O error
+   */
+  public static void upload(InputStream input, WriteChannel writer, int bufferSize)
```
elharo: How is this different from writing to any  writable byte channel? e.g. one that has nothing to do with GCS? Unless there's something special about GCS, I'm not sure this utility belongs here.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -260,6 +263,48 @@ public void downloadTo(Path path) {
...
+      throws IOException {
+    bufferSize = Math.max(bufferSize, MIN_BUFFER_SIZE);
+    byte[] buffer = new byte[bufferSize];
+    int length;
+    while ((length = input.read(buffer)) >= 0) {
```
elharo: This is close to reinventing ByteStreams.copy from guava, aside from the mix of a stream with a channel

google-cloud-storage/clirr-ignored-differences.xml
dmitry-fa: Yes, I understand, what is the procedure for such kind of changes?

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: The purpose of this method to allow user to vary the buffer size value. ByteStreams.copy does not provide such option. This is only reason why this method is made public.
Humongous files could be uploaded faster with lager buffer size, parallel upload might require smaller buffer sizes.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: The key difference here is the buffer size value and chunk size of the WriterChannel.
User can obtain easily InputStream for a file and WriteChannel returned by Storage.wrtie().
User can get benefit from calling setChunkSize on the WriterChannel instance and setting the buffer size for Blob.upload(input, writer, bufferSize)

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
elharo: Thanks. I see that now. I'm still not convinced it belongs in the Blob class. There might be a utility class somewhere where it fits better.

google-cloud-storage/clirr-ignored-differences.xml
elharo: I think at a minimum you'd need to update the version in the root pom.xml in this PR.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -1809,6 +1811,53 @@ Blob create(
   @Deprecated
   Blob create(BlobInfo blobInfo, InputStream content, BlobWriteOption... options);
 
+  /**
+   * Uploads the given {@code path} to the blob using {@link #writer}. By default any md5 and crc32c
```
elharo: MD5 and CRC32

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -204,6 +208,24 @@ public StorageObject call() {
     }
   }
 
+  @Override
+  public void upload(BlobInfo blobInfo, Path path, BlobWriteOption... options) throws IOException {
```
elharo: If these could be static you don't need to break the API.

dmitry-fa: To upload data we need to get either a WriterChannel instance of a Storage instance.
To not brake the API we can add:
void Blob.upload(path) // this requires a Blob instance to be created before
or/and
static void Blob.upload(storage, path)

Perhaps this adding just static Blob.upload(storage, path/input, options) will be a compromise between the convenient  API (void Storage.upload(path/input)  and incompatible change.

dmitry-fa: I think you're absolutely right, adding helper methods should not break compatibility.
Would it be okay to add a class like:
```
public final class StorageUtils {
    public final Storage storage;
   
    private StorageUtils(Storage storage) {
        this.storage = storage;
    }
   
    public static StorageUtils getInstance(Storage storage) {
        return new StorageUtils(storage);
    }

    public void upload(BlobInfo blobInfo, Path path, Storage.BlobWriteOption... options) throws IOException {
    }

    public void upload(BlobInfo blobInfo, InputStream content, Storage.BlobWriteOption... options) throws IOException {
    }

    static void upload(InputStream input, WriteChannel writer) throws IOException {
    }

    public static void upload(InputStream input, WriteChannel writer, int bufferSize) throws IOException  {
    }
}
```
- appropriate place 
- no API break
- Blob instance is not required
- new features could implemented here: folder upload, parallel_composite_upload,

elharo: It's certainly OK from an API compatibility perspective. You can almost always add new non-abstract methods to classes, just not to interfaces.

I do find the mix of channels and input streams in a single method call surprising. Most APIs use old I/o (streams) or new I/O (channels) but not both at the same time. Not sure it's bad, but it's worth a close look. 

Honestly, channels and nio never really lived up to the hype. They were too late to the party. By the time they arrived, operating systems had improved to the point where running a bunch of threads with streams ran as fast or faster.

dmitry-fa: This is a good point about the mix, I never thought this aspect. 
Probably instead of the mix it would be more elegant to introduce:
`upload(ReadableByteChannel reader, WritableByteChannel writer, int bufferSize)`

And provide the example in the comment how to obtain instances of reader/writer from file/storage.

PR #214, opened 2020-03-30
feat: add upload functionality

### LINKED ISSUE #179: add Storage.upload(Path)

dmitry-fa:
Fixes #179

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage;
+
+import com.google.cloud.WriteChannel;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
+import java.nio.channels.ReadableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Utility methods to perform various operations with the Storage such as upload.
+ *
+ * <p>Example of uploading files from a folder:
+ *
+ * <pre>{@code
+ * File folder = new File("pictures/");
+ * StorageUtils utils = StorageUtils.create(storage);
+ * for (File file: folder.listFiles()) {
+ *     if (!file.isDirectory()) {
+ *         BlobInfo blobInfo = BlobInfo.newBuilder(BUCKET, file.getName()).build();
+ *         try {
+ *             utils.upload(blobInfo, file.toPath());
+ *         } catch (IOException e) {
+ *             System.err.println("Unable to upload " + file + ": " + e.getMessage());
+ *         }
+ *     }
+ * }
+ * }</pre>
+ */
+public final class StorageUtils {
+
+  /** The instance of the Storage the utilities are associated with. */
+  public final Storage storage;
+
+  private static final int DEFAULT_BUFFER_SIZE = 15 * 1024 * 1024;
+  private static final int MIN_BUFFER_SIZE = 256 * 1024;
+
+  private StorageUtils(Storage storage) {
+    this.storage = storage;
+  }
+
+  /**
+   * Creates a new utility object associated with the given storage.
+   *
+   * @param storage the Storage
+   * @return an instance which refers to {@code storage}
+   */
+  public static StorageUtils create(Storage storage) {
+    return new StorageUtils(storage);
+  }
+
+  /**
+   * Uploads the given {@code path} to the blob using {@link Storage#writer}. By default any MD5 and
+   * CRC32C values in the given {@code blobInfo} are ignored unless requested via the {@link
+   * Storage.BlobWriteOption#md5Match()} and {@link Storage.BlobWriteOption#crc32cMatch()} options.
+   * Folder upload is not supported.
+   *
+   * <p>Example of uploading a file:
+   *
+   * <pre>{@code
+   * String bucketName = "my-unique-bucket";
+   * String fileName = "readme.txt";
+   * BlobId blobId = BlobId.of(bucketName, fileName);
+   * BlobInfo blobInfo = BlobInfo.newBuilder(blobId).setContentType("text/plain").build();
+   * StorageUtils.create(storage).upload(blobInfo, Paths.get(fileName));
+   * }</pre>
+   *
+   * @param blobInfo blob to create
+   * @param path file to upload
+   * @param options blob write options
+   * @throws IOException on I/O error
+   * @throws StorageException on failure
+   * @see #upload(BlobInfo, Path, int, Storage.BlobWriteOption...)
+   */
+  public void upload(BlobInfo blobInfo, Path path, Storage.BlobWriteOption... options)
+      throws IOException {
+    upload(blobInfo, path, DEFAULT_BUFFER_SIZE, options);
+  }
+
+  /**
+   * Uploads the given {@code path} to the blob using {@link Storage#writer} and the given {@code
+   * bufferSize}. By default any MD5 and CRC32C values in the given {@code blobInfo} are ignored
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+ *         try {
+ *             utils.upload(blobInfo, file.toPath());
+ *         } catch (IOException e) {
+ *             System.err.println("Unable to upload " + file + ": " + e.getMessage());
+ *         }
+ *     }
+ * }
+ * }</pre>
+ */
+public final class StorageUtils {
```
elharo: "Utils" classes usually don't have object state or non-static methods. Perhaps there's a better name here? Uploader or StorageUploader perhaps?

dmitry-fa: This class could be used not only of uploading, it could contain downloading or another functionality as well. Would it be okay to rename it to StorageHelper?

elharo: That works. Another possibility: StorageTransmitter. 

However, also consider whether the Storage field works better as a method argument.

dmitry-fa: I think it's a very rare case when someone has more than one instance of the storage. So, passing this instance to every method doesn't seem to be convenient. Methods with 4 arguments doesn't look as easy to understand.
Having a class above the storage will certainly give some freedom in extending API without breaking compatibility. To me 'Transmitter' sounds like something closed to the physical layer. 'Helper'  is very generic. May be StorageOperations or StorageFunctions fits better?

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
+import java.nio.channels.ReadableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Utility methods to perform various operations with the Storage such as upload.
+ *
+ * <p>Example of uploading files from a folder:
```
elharo: "directory" might or might not be clearer here. Depending on whether the reader is coming from Mac, Windows, or Linux. However Java does call this directory, so we should probably stick to that.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * <pre>{@code
+ * File folder = new File("pictures/");
+ * StorageUtils utils = StorageUtils.create(storage);
+ * for (File file: folder.listFiles()) {
+ *     if (!file.isDirectory()) {
+ *         BlobInfo blobInfo = BlobInfo.newBuilder(BUCKET, file.getName()).build();
```
elharo: two space indents in example

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
elharo: StorageMover? StorageTransporter?

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+ *             System.err.println("Unable to upload " + file + ": " + e.getMessage());
+ *         }
+ *     }
+ * }
+ * }</pre>
+ */
+public final class StorageUtils {
+
+  /** The instance of the Storage the utilities are associated with. */
+  public final Storage storage;
```
elharo: important: private

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+    this.storage = storage;
+  }
+
+  /**
+   * Creates a new utility object associated with the given storage.
+   *
+   * @param storage the Storage
+   * @return an instance which refers to {@code storage}
+   */
+  public static StorageUtils create(Storage storage) {
```
elharo: This is no better than a constructor. Remove it.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+   * @see #upload(BlobInfo, Path, int, Storage.BlobWriteOption...)
+   */
+  public void upload(BlobInfo blobInfo, Path path, Storage.BlobWriteOption... options)
+      throws IOException {
+    upload(blobInfo, path, DEFAULT_BUFFER_SIZE, options);
+  }
+
+  /**
+   * Uploads the given {@code path} to the blob using {@link Storage#writer} and the given {@code
+   * bufferSize}. By default any MD5 and CRC32C values in the given {@code blobInfo} are ignored
```
elharo: delete "the given"

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+   * unless requested via the {@link Storage.BlobWriteOption#md5Match()} and {@link
+   * Storage.BlobWriteOption#crc32cMatch()} options. Folder upload is not supported.
+   *
+   * <p>{@link #upload(BlobInfo, Path, Storage.BlobWriteOption...)} invokes this one with a buffer
```
elharo: what is "this one"?

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+    if (Files.isDirectory(path)) {
+      throw new StorageException(0, path + " is a directory");
+    }
+    try (InputStream input = Files.newInputStream(path)) {
+      upload(blobInfo, input, bufferSize, options);
+    }
+  }
+
+  /**
+   * Uploads the given {@code content} to the blob using {@link Storage#writer}. By default any MD5
```
elharo: "Uploads the given {@code content}" --> Reads bytes from an input stream and uploads those bytes to the blob

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+   * and CRC32C values in the given {@code blobInfo} are ignored unless requested via the {@link
+   * Storage.BlobWriteOption#md5Match()} and {@link Storage.BlobWriteOption#crc32cMatch()} options.
+   *
+   * <p>Example of uploading data with CRC32C checksum:
+   *
+   * <pre>{@code
+   * BlobId blobId = BlobId.of(bucketName, blobName);
+   * byte[] content = "Hello, world".getBytes(UTF_8);
```
elharo: StandardCharsets.UTF_8

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * Utility methods to perform various operations with the Storage such as upload.
+ *
+ * <p>Example of uploading files from a folder:
+ *
+ * <pre>{@code
+ * File folder = new File("pictures/");
+ * StorageUtils utils = StorageUtils.create(storage);
+ * for (File file: folder.listFiles()) {
```
elharo: I'm not sure why you have this example at all. It doesn't handle subdirectories.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
dmitry-fa: > StorageMover? StorageTransporter?

This class can be used not only for upload/download operation, but for any operation with the storage. That will allow to add new functionality without extending the Storage interface and breaking API. 

May be StorageOperations or StorageFunctions feet better?

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
dmitry-fa: This example demonstrates that:
- StorageUtils instance can be reused
- Upload of folders is not supported

To upload a folder with subfolders one should case about nuances like correct handling of symbolic links. Getting the blob name for files in subfolders requires an extra code. So the example will be too complicated.
I can remove this one.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageUtils.java
elharo: I like StorageOperations, actually.

dmitry-fa:
@elharo thanks a lot for your comments

PR #218, opened 2020-03-31
fix: Blob.downloadTo() methods do not wrap RetryHelper$RetryHelperException

### LINKED ISSUE #100: Blob.downloadTo() methods do not wrap RetryHelper$RetryHelperException

dmitry-fa:
Fixes #100

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
@@ -32,6 +32,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import com.google.api.core.ApiClock;
 import com.google.api.gax.retrying.RetrySettings;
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
```diff
@@ -662,4 +661,25 @@ public Long answer() throws Throwable {
     byte actual[] = Files.readAllBytes(file.toPath());
     assertArrayEquals(expected, actual);
   }
+
+  @Test
```
elharo, line 665: Per Google practices, new tests should use Mockito, not EasyMock. Yes, even if this means some tests use Mockito and others use EasyMock.

dmitry-fa: Is this rule applied to the tests within a single file or withing a package? 
Is it possible to make an exception?

frankyn:
I think we will need to make an exception for now but may be good to file
an issue to track it as future work.

PR #224, opened 2020-04-02
Revert "feat: add upload functionality"

frankyn:
Reverts googleapis/java-storage#214

Let's revert this and talk through it more before we finalize and merge again.

Once we release this change we will end with having to support it long term. I want a solution quickly but don't want to create friction.

Thank you for your patience.


dmitry-fa:
@frankyn Integration test fails: com.google.cloud.storage.StorageException: The project exceeded the rate limit for creating and deleting buckets.

PR #233, opened 2020-04-08
fix: Blob API Doc is confusing

### LINKED ISSUE #176: Blob API Doc is confusing
I'm not sure if the issue is the docs, the API design, or both. However I'm having real trouble understanding the API of this class. 

In particular I'm confused about what Blob objects are and what they represent. On the one hand "A Google cloud storage object." On the other hand these objects are immutable and provide methods to modify the blob on GCS which return new Blob objects.

I suspect the Javadoc is having trouble distinguishing an object that points to the thing from the thing itself.

A Google cloud storage object.

Objects of this class are immutable. Operations that modify the blob like update(com.google.cloud.storage.Storage.BlobTargetOption...) and copyTo(com.google.cloud.storage.BlobId, com.google.cloud.storage.Blob.BlobSourceOption...) return a new object. To get a Blob object with the most recent information use reload(com.google.cloud.storage.Blob.BlobSourceOption...). Blob adds a layer of service-related functionality over BlobInfo.

dmitry-fa:
Fixes #176

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -52,19 +52,19 @@ import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 /**
- * A Google cloud storage object. A {@code Blob} object includes the {@code BlobId} instance, the
- * set of properties inherited from the {@link BlobInfo} class and the {@code Storage} instance. The
- * class provides methods to perform operations on the object. Reading a property value does not
+ * An object in Google Cloud Storage. A {@code Blob} object includes the {@code BlobId} instance,
+ * the set of properties inherited from the {@link BlobInfo} class and the {@code Storage} instance.
+ * The class provides methods to perform operations on the object. Reading a property value does not
  * issue any RPC calls. The object content is not stored within the {@code Blob} instance.
  * Operations that access the content issue one or multiple RPC calls, depending on the content
  * size.
  *
  * <p>Objects of this class are immutable. Operations that modify the blob like {@link #update} and
- * {@link #copyTo} return a new object. Any changes to a Storage object made after creation of the
- * {@code Blob} are not visible in the {@code Blob}. To get a {@code Blob} object with the most
- * recent information use {@link #reload}.
+ * {@link #copyTo} return a new object. Any changes to the object in Google Cloud Storage made after
+ * creation of the {@code Blob} are not visible in the {@code Blob}. To get a {@code Blob} object
+ * with the most recent information use {@link #reload}.
  *
- * <p>Example of getting the content of a Storage object:
+ * <p>Example of getting the content of the object in Google Cloud Storage:
  *
  * <pre>{@code
  * BlobId blobId = BlobId.of(bucketName, blobName);
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -63,6 +63,15 @@ import java.util.concurrent.TimeUnit;
  * {@link #copyTo} return a new object. Any changes to a Storage object made after creation of the
  * {@code Blob} is not visible in the {@code Blob}. To get a {@code Blob} object with the most
  * recent information use {@link #reload}.
+ *
+ * <p>Example of getting the content of a Storage object:
+ *
+ * <pre>{@code
+ * BlobId blobId = BlobId.of(bucketName, blobName);
+ * Blob blob = storage.get(blobId);
+ * long size = blob.getSize(); // no RPC call is required
+ * byte[] content = blob.getContent(); // one or multiple RPC calls will be issued
+ * }</pre>
  */
 public class Blob extends BlobInfo {
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -52,12 +52,17 @@ import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 /**
- * A Google cloud storage object.
+ * A Google cloud storage object. A {@code Blob} object includes the {@code BlobId} instance, the
+ * set of properties inherited from the {@link BlobInfo} class and the {@code Storage} instance. The
+ * class provides methods to perform operations on the object. Reading a property value does not
+ * issue any RPC calls. The object content is not stored within the {@code Blob} instance,
+ * operations on obtaining the content will issue one or multiple RPC calls, depending on the
+ * content size.
  *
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -52,12 +52,26 @@
 import java.util.concurrent.TimeUnit;
 
 /**
- * A Google cloud storage object.
+ * A Google cloud storage object. A {@code Blob} object includes the {@code BlobId} instance, the
+ * set of properties inherited from the {@link BlobInfo} class and the {@code Storage} instance. The
+ * class provides methods to perform operations on the object. Reading a property value does not
+ * issue any RPC calls. The object content is not stored within the {@code Blob} instance,
```
elharo, line 58: instance.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -52,12 +52,26 @@
...
+ * operations on obtaining the content will issue one or multiple RPC calls, depending on the
```
elharo, line 59: Operations that access the content issue...

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -52,12 +52,26 @@
...
+ * content size.
  *
  * <p>Objects of this class are immutable. Operations that modify the blob like {@link #update} and
- * {@link #copyTo} return a new object. To get a {@code Blob} object with the most recent
- * information use {@link #reload}. {@code Blob} adds a layer of service-related functionality over
- * {@link BlobInfo}.
+ * {@link #copyTo} return a new object. Any changes to a Storage object made after creation of the
+ * {@code Blob} is not visible in the {@code Blob}. To get a {@code Blob} object with the most
```
elharo, line 64: is not --> are not

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: done

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: done

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: done.  But _access the content issue_ might sound confusing.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -52,12 +52,26 @@
 import java.util.concurrent.TimeUnit;
 
 /**
- * A Google cloud storage object.
+ * A Google cloud storage object. A {@code Blob} object includes the {@code BlobId} instance, the
```
elharo, line 55: Be very careful with the word "object". This is very easy to confuse wit an instance of the Storage class. I suggest using the phrase "an object in Google Cloud Storage" when you mean the bytes on the server rather than "storage object" which sounds like an instance of a class.

The overloading of the word "object" is really annoying. GCS should never have called these things "objects" but that ship, I am afraid, has sailed and we can't fix it here. :-(

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -52,12 +52,26 @@
...
+ * class provides methods to perform operations on the object. Reading a property value does not
+ * issue any RPC calls. The object content is not stored within the {@code Blob} instance.
+ * Operations that access the content issue one or multiple RPC calls, depending on the content
+ * size.
  *
  * <p>Objects of this class are immutable. Operations that modify the blob like {@link #update} and
- * {@link #copyTo} return a new object. To get a {@code Blob} object with the most recent
- * information use {@link #reload}. {@code Blob} adds a layer of service-related functionality over
- * {@link BlobInfo}.
+ * {@link #copyTo} return a new object. Any changes to a Storage object made after creation of the
```
elharo, line 63: Storage object -- the object in Google Cloud Storage

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -52,12 +52,26 @@
...
+ * {@code Blob} are not visible in the {@code Blob}. To get a {@code Blob} object with the most
+ * recent information use {@link #reload}.
+ *
+ * <p>Example of getting the content of a Storage object:
```
elharo, line 67: Storage object -- the object in Google Cloud Storage

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -44,7 +44,21 @@
 import java.util.Set;
 
 /**
- * Google Storage object metadata.
+ * Google Storage object information. A {@code BlobInfo} object includes the {@code BlobId} instance
```
elharo, line 47: Information about an object in Google Cloud Storage.

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -44,7 +44,21 @@
...
+ * and the set of properties, such as the blob's access control configuration, user provided
+ * metadata, the CRC32C checksum, etc. Instances of this class are used to create a new Storage
```
elharo, line 49: new object in Google Cloud Storage

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: agree

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -44,7 +44,21 @@
...
- * Google Storage object metadata.
+ * Information about an object in Google Cloud Storage. A {@code BlobInfo} object includes the
+ * {@code BlobId} instance and the set of properties, such as the blob's access control
+ * configuration, user provided metadata, the CRC32C checksum, etc. Instances of this class are used
+ * to create a new object in Google Cloud Storage or update the properties of an existing object. To
+ * deal with existing Storage objects the API includes the {@link Blob} class which extends {@code
+ * BlobInfo} and declares methods to perform operations on the object. Neither {@code BlobInfo} nor
+ * {@code Blob} instances keep the object content, just the object properties.
+ *
+ * <p>Example of usage {@code BlobInfo} to crate an object in Google Cloud Storage:
```
elharo, line 55: crate --> create

dmitry-fa: Thanks a lot!

PR #236, opened 2020-04-10
fix: deprecate Blob.update() and modify documentation of Storage.update methods

### LINKED ISSUE #234: Blob.update() misses BlobInfo argument
The Blob class declares method: `public Blob update(BlobTargetOption... options)`
This method is useless, it doesn't have the necessary BlobInfo parameter.
Documentation to this method mentions Storage.update(), not Blob.update().

This method should be declared as Deprecated (and deleted as soon possible)

A new  method with BlobInfo parameter should be added.

dmitry-fa:
Fixes #234

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -532,46 +532,11 @@ public class Blob extends BlobInfo {
   }
 
   /**
-   * Updates the blob's information. Bucket or blob's name cannot be changed by this method. If you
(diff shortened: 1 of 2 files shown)


frankyn:
Please wait to merge this until I have a chance to review next week.

dmitry-fa:
@frankyn 
> Please wait to merge this until I have a chance to review next week.

Sure!

frankyn:
Updated, please don't deprecate this method. It's a reasonable convenience method. I added an example in the associated issue. 

Users may be dependent on this method and deprecating it can cause friction as we've seen before with the deprecation of `Storage.create()` for `InputStream`.

PR #237, opened 2020-04-13
update libraries-bom

elharo:
@chingor13 What needs to be done to make this happen automatically?


chingor13:
We explicitly have the README excluded for generation so the libraries-bom is not automatically updated. Opened #238 to convert the README to use our template.

frankyn:
Deduping this PR on #238

PR #238, opened 2020-04-13
chore: generate README from templates

chingor13:
Custom content is now in `.readme-partials.yaml` so any updates need to go there for now. This will allow automatic updates to the install sections for the latest libraries-bom versions and keep our README format in sync with the other client libraries.

Also fixes a few broken links to the google-cloud-examples that still live in the monorepo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,53 +1,52 @@
-Google Cloud Java Client for Storage
-====================================
+# Google Google Cloud Storage Client for Java
 
-Java idiomatic client for [Google Cloud Storage][cloud-storage].
+Java idiomatic client for [Google Cloud Storage][product-docs].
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -186,63 +191,85 @@ for (Bucket bucket : storage.list().iterateAll()) {
...
 
-It is currently in major version one (``1.y.z``), which means that the public API should be considered stable.
+## Java Versions
 
-Contributing
-------------
+Java 7 or above is required for using this client.
 
-Contributions to this library are always welcome and highly encouraged.
+## Versioning
```
frankyn, line 218: @chingor13 how can I update versioning information for this library based on discussions around the Storage interface?

chingor13: I'll need to add an opt-out in the shared template configuration

frankyn: Gotcha, is this something you want to do after this PR?

PR #241, opened 2020-04-13
fix: surface storage interface expectations correctly.

frankyn:
cc: @dmitry-fa , @athakor based on pending PRs.

We will continue to update the Storage interface and mark it as InternalExtensionOnly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,8 @@ See [TESTING] to read more about testing.
 Versioning
 ----------
 
-This library follows [Semantic Versioning](http://semver.org/).
+This library follows [Semantic Versioning](http://semver.org/), but does update [Storage interface](src/main/java/com.google.cloud.storage/Storage.java)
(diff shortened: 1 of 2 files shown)


dmitry-fa:
This fix is a good solution for the situation with extending the Storage interface. My question, is this change allowed?
Quoting: [InternalExtensionOnly](https://googleapis.github.io/api-common-java/1.6.0/apidocs/com/google/api/core/InternalExtensionOnly.html)
> Adding this annotation to an API is considered API-breaking.

frankyn:
Hi @dmitry-fa, +1 to being allowed. I went through internal review/discussion with folks. The annotation is only documentation and doesn't change how the interface can be used. 

Also acknowledging that this library doesn't follow Semantic versioning strictly but looking for input from @JesseLovelace and @crwilcox on this point. I may end up removing it until later.

PR #261, opened 2020-04-16
fix: Documentation for Blob.update() and Storage.update() methods is confusing/incorrect

### LINKED ISSUE #252: Documentation for Blob.update() and Storage.update() methods is confusing/incorrect.
Documentation for Blob.update(Options...) doesn't include example of usage, the provided examples demonstrate how to use `Storage.update(blobInfo)`. The doc also mentions non existing parameter `blobInfo`: `Original metadata are merged with metadata in the provided blobInfo. `
Taking into account the fact that `Objects of this class are immutable.` this method seems useless. 
The documentation should be updated to clarify usage.

Documentation for Storage.update() methods is better but still confusing. The important aspects are not covered:
- The term `metadata` could be read as `blob's information` or as `blob's user provided metadata`
- It's not clarify if blob's generation is updated
- What happens if  the preconditions specified by the options fail

Documentation should be aligned with the new Blob/BlobInfo API Doc introduced by https://github.com/googleapis/java-storage/issues/176

dmitry-fa:
Fixes #252

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -563,8 +563,10 @@ public class Blob extends BlobInfo {
    * }</pre>
    *
    * @param options preconditions to apply the update
-   * @return the updated blob
+   * @return the updated {@code Blob}
    * @throws StorageException upon failure
+   * @see <a
+   *     href="https://cloud.google.com/storage/docs/json_api/v1/objects/update">https://cloud.google.com/storage/docs/json_api/v1/objects/update</a>
    */
   public Blob update(BlobTargetOption... options) {
     return storage.update(this, options);
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -546,9 +546,9 @@ public class Blob extends BlobInfo {
   }
 
   /**
-   * Updates the blob properties. Optional {@code options} parameter can contain the preconditions
-   * for applying the update. To update the properties you need to call {@link #toBuilder()}, set
-   * the properties you want to change, build the new {@code Blob} instance and then call {@link
+   * Updates the blob properties. The {@code options} parameter contains the preconditions for
+   * applying the update. To update the properties call {@link #toBuilder()}, set the properties you
+   * want to change, build the new {@code Blob} instance, and then call {@link
    * #update(BlobTargetOption...)}.
    *
    * <p>The property update details are described in {@link Storage#update(BlobInfo)}. {@link
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -546,44 +546,24 @@ public class Blob extends BlobInfo {
   }
 
   /**
-   * Updates the blob's information. Bucket or blob's name cannot be changed by this method. If you
-   * want to rename the blob or move it to a different bucket use the {@link #copyTo} and {@link
-   * #delete} operations. A new {@code Blob} object is returned. By default no checks are made on
-   * the metadata generation of the current blob. If you want to update the information only if the
-   * current blob metadata are at their latest version use the {@code metagenerationMatch} option:
-   * {@code newBlob.update(BlobTargetOption.metagenerationMatch())}.
+   * Updates the blob properties. Optional {@code options} parameter can contain the preconditions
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -546,44 +546,24 @@ public Blob reload(BlobSourceOption... options) {
   }
 
   /**
-   * Updates the blob's information. Bucket or blob's name cannot be changed by this method. If you
-   * want to rename the blob or move it to a different bucket use the {@link #copyTo} and {@link
-   * #delete} operations. A new {@code Blob} object is returned. By default no checks are made on
-   * the metadata generation of the current blob. If you want to update the information only if the
-   * current blob metadata are at their latest version use the {@code metagenerationMatch} option:
-   * {@code newBlob.update(BlobTargetOption.metagenerationMatch())}.
+   * Updates the blob properties. Optional {@code options} parameter can contain the preconditions
```
elharo, line 549: Optional --> The
can contain the --> contains

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -546,44 +546,24 @@ public Blob reload(BlobSourceOption... options) {
...
+   * for applying the update. To update the properties you need to call {@link #toBuilder()}, set
```
elharo, line 550: delete "you need to"

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -546,44 +546,24 @@ public Blob reload(BlobSourceOption... options) {
...
+   * the properties you want to change, build the new {@code Blob} instance and then call {@link
```
elharo, line 551: instance,

dmitry-fa: Thanks!

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -546,44 +546,24 @@ public Blob reload(BlobSourceOption... options) {
...
-   *     .build());
+   * BlobId blobId = BlobId.of(bucketName, blobName);
+   * Blob blob = storage.get(blobId);
+   * blob.toBuilder().setContentType("text/plain").build().update();
    * }</pre>
    *
-   * @param options update options
-   * @return a {@code Blob} object with updated information
+   * @param options preconditions to apply the update
+   * @return the updated blob
```
frankyn, line 566: `the updated {@code Blob}`

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -546,44 +546,24 @@ public Blob reload(BlobSourceOption... options) {
   }
 
   /**
-   * Updates the blob's information. Bucket or blob's name cannot be changed by this method. If you
-   * want to rename the blob or move it to a different bucket use the {@link #copyTo} and {@link
-   * #delete} operations. A new {@code Blob} object is returned. By default no checks are made on
-   * the metadata generation of the current blob. If you want to update the information only if the
-   * current blob metadata are at their latest version use the {@code metagenerationMatch} option:
-   * {@code newBlob.update(BlobTargetOption.metagenerationMatch())}.
+   * Updates the blob properties. The {@code options} parameter contains the preconditions for
```
frankyn, line 549: I'd state `{@code options}` is `BlobTargetOption` and instead of `preconditions` use `request options` and link to update API documentation: https://cloud.google.com/storage/docs/json_api/v1/objects/update

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: okay

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: > and link to update API documentation: https://cloud.google.com/storage/docs/json_api/v1/objects/update

done

> I'd state {@code options} is BlobTargetOption

I think this is clear from the parameter type

> and instead of preconditions use request options

Personally for me it was hard to understand the purspose of this parameter. So, I tried to rephrase this statement to help users not very familiar with the Storage details to get the meaning. I came up to:  `The options parameter contains the preconditions for  applying the update`

frankyn: Thanks, sgtm, I think having the link will help provide information.

PR #267, opened 2020-04-17
docs: label legacy storage classes in documentation

### LINKED ISSUE #254: docs: label legacy Storage Classes in documentation
Cloud Storage storage classes are listed in [library javadocs](https://googleapis.dev/java/google-cloud-storage/latest/com/google/cloud/storage/StorageClass.html) and they need to be updated to label legacy storage with [current guidance](https://cloud.google.com/storage/docs/storage-classes#legacy) on using "STANDARD" for new buckets.

Legacy Storage Classes:

* DURABLE_REDUCED_AVAILABILITY
* MULTI_REGIONAL
* REGIONAL

@dmitry-fa can you pick up this change?

dmitry-fa:
Fixes #254

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
@@ -20,8 +20,9 @@ import com.google.cloud.StringEnumType;
 import com.google.cloud.StringEnumValue;
 
 /**
- * Enums for the storage classes. See https://cloud.google.com/storage/docs/storage-classes for
- * details.
+ * Enums for the storage classes. See <a
+ * href="https://cloud.google.com/storage/docs/storage-classes">https://cloud.google.com/storage/docs/storage-classes</a>
+ * for details.
  */
 public final class StorageClass extends StringEnumValue {
   private static final long serialVersionUID = -6938125060419556331L;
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/StorageClass.java
```diff
@@ -42,42 +43,64 @@ public StorageClass apply(String constant) {
...
+   * @see <a
+   *     href="https://cloud.google.com/storage/docs/storage-classes#legacy">https://cloud.google.com/storage/docs/storage-classes#legacy</a>
    */
   public static final StorageClass REGIONAL = type.createAndRegister("REGIONAL");
 
   /**
    * Multi-regional storage class. This is supported as a legacy storage class and will be
-   * deprecated in the future. See: https://cloud.google.com/storage/docs/storage-classes for
-   * details
+   * deprecated in the future. This class cannot be set. Unless you already are using it, you should
```
frankyn, line 89: Legacy storage classes can still be set, please remove `This class cannot be set` for all legacy storage classes.

Additionally, the issue that didn't across in my issue is that a developer at first only see's `Mult-regional storage class.` and not the rest. 

ySdN1VJtAXu

The additional information isn't available until scrolling down:

FAuvGNOgqhx

If it's being truncated because of the period `.`. then I recommend this:

> Legacy Multi-regional storage class, use REGIONAL storage class for new buckets.

dmitry-fa: > Legacy Multi-regional storage class, use REGIONAL storage class for new buckets.

1) Do you really mean REGIONAL?

https://cloud.google.com/storage/docs/storage-classes#legacy recommends STANDARD class:
> Unless you already are using one of these additional classes, you should use Standard Storage instead.

2) A storage class can be set not only for buckets, but for object as well. And not only for new, but for existing. Right? May be it's not correct to say `for new buckets`

frankyn: :man_facepalming: I meant STANDARD. 

Good catch (blobs vs buckets), then use:

> Legacy {Storage class} storage class, use STANDARD storage class instead.

dmitry-fa: legacyStorageClasses1

...

legacyStorageClasses2

Issue #138, opened 2020-02-19
Expose the methods of Notifications


athakor:
Expose the following methods of Storage Notifications.

- [x] [createNotification](https://cloud.google.com/storage/docs/json_api/v1/notifications/insert)

- [x] [getNotification](https://cloud.google.com/storage/docs/json_api/v1/notifications/get)

- [x] [listNotification](https://cloud.google.com/storage/docs/json_api/v1/notifications/list)

- [x] [deleteNotification](https://cloud.google.com/storage/docs/json_api/v1/notifications/delete)

/cc: @frankyn @JesseLovelace

frankyn:
You can pull most of the implementation to complete this task from:

https://github.com/Googleapis/java-notification

It was implemented in a separate artifact and customers would benefit from having it in the same artifact.

The reason why it was split off was to add a helper method when defining a pubsub topic generated with the PubSub client. It would be great to merge and get deprecate the other artifact.

PR #269, opened 2020-04-17
feat: add storage.upload(path)

### LINKED ISSUE #179: add Storage.upload(Path)

dmitry-fa:
Fixes #179


google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
```diff
@@ -3072,4 +3078,183 @@ public void testWriterWithSignedURL() throws MalformedURLException {
...
+      Assert.fail();
+    } catch (IOException e) {
+      assertEquals(NoSuchFileException.class, e.getClass());
+      assertEquals(fileName, e.getMessage());
+    }
+  }
+
+  @Test
+  public void testUploadDirectory() throws IOException {
+    EasyMock.replay(storageRpcMock);
```
elharo, line 3098: Please use Mockito for all new tests. It really is better.

dmitry-fa: @elharo I was going to use Mockito for these new tests, but found this idea not very productive.
I submitted #270 and assigned to myself to fix this.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -1811,6 +1813,113 @@ Blob create(
...
+   * }</pre>
+   *
+   * @param blobInfo blob to create
+   * @param path file to upload
+   * @param options blob write options
+   * @throws IOException on I/O error
+   * @throws StorageException on failure
+   * @see #upload(BlobInfo, Path, int, BlobWriteOption...)
+   */
+  void upload(BlobInfo blobInfo, Path path, BlobWriteOption... options) throws IOException;
```
frankyn, line 1839: Blob should be returned from upload() methods similar to create() methods.

dmitry-fa: returning Blob requires an extra RPC request to be issued, that is not always needed.
I can add uploadFrom() method to the Blob class (symmetric to downloadTo)  to return an updated Blob.

dmitry-fa: A month ago I compared upload/download implemented in various ways for several types of file. This is the [summary](https://docs.google.com/document/d/19u1wYn-8kDdz9bv2m106ApGmZV3KF8__EGuQSn1g2gU/edit#)

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -204,6 +214,54 @@ public StorageObject call() {
...
+    try (WriteChannel writer = writer(blobInfo, options)) {
+      upload(Channels.newChannel(content), writer, bufferSize);
+    }
+  }
+
+  /*
+   * Uploads the given content to the storage using specified write channel and the given buffer
+   * size. This method does not close any channels.
+   */
+  private static void upload(ReadableByteChannel reader, WriteChannel writer, int bufferSize)
```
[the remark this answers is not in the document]
dmitry-fa: transmit() ?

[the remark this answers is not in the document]
frankyn: `uploadHelper()`

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
dmitry-fa: Do you mean we can update the StorageRpc interface to make void write(uploadId, ...) to return a StorageObject?

dmitry-fa: > 1 nit so far,  added @JesseLovelace for surface review as well.
> Blob should be returned from upload() methods similar to create() methods.

It's not a nit at all :) It's a significant change. I agree with you, upload() should return Blob.
I see the following ways how to implement:

1) Update the StorageRpc interface and make write() to return a StorageObject instance or null in  write is not final request for the given upload. Now StorageRpc .write() returns null.

2) Update HttpStorageRpc.write() to remember StorageObject  and return this object in a separate method.
In this case BlobWriterChannel.fluchBuffer() will look like:
```
public void run() {
         StorageRpc  rpc = getOptions().getStorageRpcV1();
         rpc.write(getUploadId(), getBuffer(), 0, getPosition(), length, last);
         if (last && rpc instnceof HttpStorageRpc) {
             this.uploadedBlob = ((HttpStorageRpc)rpc.getUplodedBlob())
         } else {
             this.uploadedBlob = null;  // let storage.upload() to resolve it
         }
}
```
3) Issue an extra rpc per upload to return a Blob

I believe 2. is not a Google way, so we should go 1. As a temporary solution 3. is also considered, because it could be improved to 1.

WDYT?

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
dmitry-fa: @elharo, new tests for upload() converted to mockito.
It's not a final version yet. upload() method will be updated to return Blob (now they return void), that will require test update as well.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
dmitry-fa: Hi @frankyn,

> When the upload is complete GCS responds with object metadata so we don't need to perform an additional get request.

I managed to implement upload based on your hint. I modified `StorageRpc.write()` to return `StorageObject` instead of `void`.
Two issues were found:
- GCS does not respond with object metadata if writer was created by `storage.writer(signedUrl)`
Not sure if it's a bug or a feature, looks like a bug: [spec](https://developers.google.com/gdata/docs/resumable_upload#Upload) says nothing about it. I worked it around.

- Change StrorageRpc affects storage-nio, Kokoro - Test: Linkage Monitor fails:
`(com.google.cloud:google-cloud-nio:0.120.0-alpha) com.google.cloud.storage.contrib.nio.testing.FakeStorageRpc's method write(String arg1, byte[] arg2, int arg3, long arg4, int arg6, boolean arg7) is not implemented in the class referenced from com.google.cloud.storage.spi.v1.StorageRpc (com.google.cloud:google-cloud-storage:1.107.1-SNAPSHOT)
`

cc: @JesseLovelace

google-cloud-storage/pom.xml
```diff
@@ -156,6 +156,11 @@
       <artifactId>easymock</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
```
frankyn, line 163: Can you split out mockito tests into a separate PR because it got a bit confusing in review? Apologies for churn on this.

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
```diff
@@ -47,6 +48,13 @@
     super(options, null, uploadId);
   }
 
+  // Contains metadata of the updated object or null if upload is not completed.
+  private StorageObject objectProto;
+
+  StorageObject getObjectProto() {
```
frankyn, line 54: Use: `getStorageObject()`

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
```diff
@@ -47,6 +48,13 @@
     super(options, null, uploadId);
   }
 
+  // Contains metadata of the updated object or null if upload is not completed.
+  private StorageObject objectProto;
```
frankyn, line 52: Use: `storageObject`

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -1805,12 +1807,123 @@ Blob create(
...
+   * is noticeable when uploading many objects in parallel. Buffer sizes less than 256 KiB are
+   * treated as 256 KiB.
+   *
+   * @param blobInfo blob to create
+   * @param content input stream to read from
+   * @param bufferSize size of the buffer I/O operations
+   * @param options blob write options
+   * @return a {@code Blob} with complete information
+   * @throws IOException on I/O error
+   * @throws StorageException on failure
```
frankyn, line 1922: StorageException on server side error.

google-cloud-storage/pom.xml
dmitry-fa: Agree. Done #284

dmitry-fa:
Hi @frankyn,

Thanks for you comments.

> The more I think about this you may want to use StorageRpc.open() and StorageRpc.write() directly instead of through the writer.
> 
> That is an alternative solution there, but it does introduce duplication of code. I'm not convinced that there a clean solution other than what you've done by introducing a new value of objectProto.

I agree, `objectProto` is not elegant solution, but it's better than duplication of BaseWriteChannel code. BaseWriteChannel provides nice and not trivial functionality on buffering (`setChunkSize()`) which is already tested. Repeating that may lead to new bugs.

frankyn:
Thanks @dmitry-fa, let's move forward for now. It's not introducing new surface methods in the BlobWriteChannel class. Please resolve merge conflicts. This will be blocked until Linkage Monitor is resolved because the StorageRpc update.

dmitry-fa:
Hi @frankyn, 

thanks for the prompt response. I have one more question regarding this feature. 
Once we add upload() methods to the Storage there will be two groups of methods with similar signatures which do the same job: create() and upload():

```
  @Deprecated
  Blob create(BlobInfo blobInfo, InputStream content, BlobWriteOption... options);

  Blob upload(BlobInfo blobInfo, InputStream content, BlobWriteOption... options) throws IOException;
```

The differences between them: 
- create() use direct upload, upload() use resumable upload. Would it make sense to reflect this 
- upload() throw IOException, create() throw StorageException

Would it make sense to use `create` instead of `upload` to reflect this?

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
dmitry-fa: done

frankyn:
Thanks for raising this question @dmitry-fa, in this case I think it would help to have consistent names. In my perspective a user only wants to upload. In the case of the new upload() methods, they aren't resumable because the writer is not exposed to the user. 

It would help to unify the names, but haven't thought about how to do this well. Open to suggestions.

dmitry-fa:
Hi @frankyn,

I've come up to the conclusion that having both create() and upload() methods is okay:
- `upload` implies reading the content from file with possible IOException
- `create` implies the content to be in memory (except deprecated create(InputStream)) , IOException is never thrown

frankyn:
Can you merge master and resolve conflicts @dmitry-fa

dmitry-fa:
@frankyn 
> Can you merge master and resolve conflicts @dmitry-fa

Sure, this is my plan for tomorrow. There are many changes made to the Storage, I'd like to review them carefully.

frankyn:
Sgtm, take your time, thanks for your patience @dmitry-fa, I just want to make sure you don't need another libraries-bom release.

dmitry-fa:
@frankyn, thanks for your attention to this issue.

We need at least to release the Nio to make the linkage-monitor happy, but I can't say for sure if it is enough or not.

frankyn:
I'm working on ubnlcoking releases for storage-nio in https://github.com/googleapis/java-storage-nio/issues/147. We should need one more release of libraries-bom then. 

Thanks for clarifying that @dmitry-fa, cheers.

frankyn:
@dmitry-fa could you address the conflict added after merging #138 PR?

dmitry-fa:
@frankyn I think it's not needed to be merged.

PR #308, opened 2020-05-12
fix: blob.reload() does not work as intuitively expected

### LINKED ISSUE #149: Blob.reload() does not work as intuitively expected

dmitry-fa:
Fixes #149

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -525,24 +525,38 @@ public class Blob extends BlobInfo {
   }
 
   /**
-   * Fetches current blob's latest information. Returns {@code null} if the blob does not exist.
+   * Fetches the latest blob properties. Returns {@code null} if the blob no longer exists.
    *
-   * <p>Example of getting the blob's latest information, if its generation does not match the
-   * {@link Blob#getGeneration()} value, otherwise a {@link StorageException} is thrown.
+   * <p>{@code options} parameter can contain the preconditions. E.g. user may need to get the blob
+   * properties only if the content has not been updated externally. {@code StorageException} with
+   * the code {@code 412} is thrown if preconditions fail.
+   *
+   * <p>Example of retrieving the blob's latest information only if the content is not updated
+   * externally:
    *
    * <pre>{@code
-   * Blob latestBlob = blob.reload(BlobSourceOption.generationNotMatch());
-   * if (latestBlob == null) {
-   *   // the blob was not found
+   * Blob blob = storage.get(BlobId.of(bucketName, blobName));
+   *
+   * doSomething();
+   *
+   * try {
+   *   blob = blob.reload(Blob.BlobSourceOption.generationMatch());
+   * } catch (StorageException e) {
+   *   if (e.getCode() == 412) {
+   *     // the content was updated externally
+   *   } else {
+   *     throw e;
+   *   }
    * }
    * }</pre>
    *
-   * @param options blob read options
-   * @return a {@code Blob} object with latest information or {@code null} if not found
+   * @param options preconditions to fetch
+   * @return a {@code Blob} object with latest information or {@code null} if no longer exists.
    * @throws StorageException upon failure
    */
   public Blob reload(BlobSourceOption... options) {
-    return storage.get(getBlobId(), toGetOptions(this, options));
+    BlobId id = getBlobId();
+    return storage.get(BlobId.of(id.getBucket(), id.getName()), toGetOptions(this, options));
   }
 
   /**
@@ -732,6 +746,7 @@ public class Blob extends BlobInfo {
    *     // handle exception
    *   }
    * }
+   * blob = blob.reload();
    * }</pre>
    *
    * @param options target blob options
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -550,7 +550,7 @@ public class Blob extends BlobInfo {
    * }
    * }</pre>
    *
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -525,24 +525,38 @@ public boolean exists(BlobSourceOption... options) {
   }
 
   /**
-   * Fetches current blob's latest information. Returns {@code null} if the blob does not exist.
+   * Fetches the latest blob properties. Returns {@code null} if the blob no longer exists.
    *
-   * <p>Example of getting the blob's latest information, if its generation does not match the
-   * {@link Blob#getGeneration()} value, otherwise a {@link StorageException} is thrown.
+   * <p>{@code options} parameter can contain the preconditions. E.g. user may need to get the blob
```
elharo, line 530: E.g. --> For example,

Google style prohibits Latin abbreviations

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -525,24 +525,38 @@ public boolean exists(BlobSourceOption... options) {
...
-   * @param options blob read options
-   * @return a {@code Blob} object with latest information or {@code null} if not found
+   * @param options preconditions to fetch
+   * @return a {@code Blob} object with latest information or {@code null} if no longer exists.
    * @throws StorageException upon failure
    */
   public Blob reload(BlobSourceOption... options) {
-    return storage.get(getBlobId(), toGetOptions(this, options));
+    BlobId id = getBlobId();
+    return storage.get(BlobId.of(id.getBucket(), id.getName()), toGetOptions(this, options));
```
elharo, line 559: Why do you make a copy of the ID here?

dmitry-fa: BlobId includes three components: bucket, name and generation.  I form BlobId without generation to request the latest version of the blob.

elharo: OK, please add a comment explaining this

dmitry-fa: done

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
dmitry-fa: E.g. is widely used, perhaps we need to collect such things and do some code clean up.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -525,24 +525,39 @@ public boolean exists(BlobSourceOption... options) {
   }
 
   /**
-   * Fetches current blob's latest information. Returns {@code null} if the blob does not exist.
+   * Fetches the latest blob properties. Returns {@code null} if the blob no longer exists.
    *
-   * <p>Example of getting the blob's latest information, if its generation does not match the
-   * {@link Blob#getGeneration()} value, otherwise a {@link StorageException} is thrown.
+   * <p>{@code options} parameter can contain the preconditions. For example, user may need to get
```
elharo, line 530: user may need --> the user might want

dmitry-fa: done

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -525,24 +525,39 @@ public boolean exists(BlobSourceOption... options) {
...
+   *     // the content was updated externally
+   *   } else {
+   *     throw e;
+   *   }
    * }
    * }</pre>
    *
-   * @param options blob read options
-   * @return a {@code Blob} object with latest information or {@code null} if not found
+   * @param options preconditions to fetch
```
frankyn, line 553: preconditions to use, see https://cloud.google.com/storage/docs/json_api/v1/objects/get for more information.

dmitry-fa: reload-jdoc

frankyn: Thanks @dmitry-fa,

use: `preconditions to use on reload, see https://cloud.google.com/storage/docs/json_api/v1/objects/get for more information` instead.

dmitry-fa: done

PR #311, opened 2020-05-12
Add doc for equals method in blob

### LINKED ISSUE #310: Document Blob.equals
https://googleapis.dev/java/google-cloud-storage/latest/com/google/cloud/storage/Blob.html

equals method should explain when two blobs are and are not equal

michiboo:

Fixes #310 ☕️

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -900,6 +900,10 @@ public class Blob extends BlobInfo {
     return new Builder(this);
   }
 
+  /**
+   * Returns true if obj instance {@code Blob.toPb()} metadata representation and {@code
+   * Blob.options} instance of StorageOptions are both equal.
+   */
   @Override
   public final boolean equals(Object obj) {
     if (obj == this) {


google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -900,6 +900,7 @@ public Builder toBuilder() {
     return new Builder(this);
   }
 
+  /** Returns true if obj's class, toPb and options are equals */
```
frankyn, line 903: Please change to:

```
Returns true if obj instance `<pre>{@code Blob.toPb()}</pre>` metadata representation and `<pre>{@code Blob.options}</pre>` instance of StorageOptions are both equal.
```

Also generate javadocs and take a screenshot of the resulting change.

michiboo: image

Hi please see the screen shot above

frankyn: Please inline `<pre>{@code Blob.toPb()}</pre>` and `<pre>{@code Blob.options}</pre>`.

michiboo: @frankyn If I put it inline it won't pass the lint test.... mvn com.coveo:fmt-maven-plugin:format auto format like this.

frankyn: Ah, do this instead:

```
Returns true if obj instance {@code Blob.toPb()} metadata representation and {@code Blob.options} instance of StorageOptions are both equal.
```

michiboo: image

michiboo:
@frankyn  There is error on Kokoro integration test Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 400 Bad Request
"Bucket is requester pays bucket but no user project provided."

PR #338, opened 2020-05-26
feat: auto content-type on blob creation

### LINKED ISSUE #47: Feature suggestion: auto content-type on blob creation
What if the API saves the blob setting it's content-type automatically? I guess it would make sense at least for public blobs. Do you have an opinion on that?

Cheers

dmitry-fa:
Fixes #47 

Proposed fix of the auto detection of the content type based on the blob name, like other clients do (php, nodejs, go).

The class javax.activation.MimetypesFileTypeMap from JDK is used to detect the content type. Using this class with Java 11+ requires the dependency on Java Activation Framework.


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -81,6 +81,7 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import javax.activation.MimetypesFileTypeMap;
```
elharo, line 84: Use com.google.common.net.MediaType so you don't introduce another dependency, and ping the guava team to get this out of beta

dmitry-fa: com.google.common.net.MediaType doesn't provide mapping extension to the meida type. Its parse method recognizes strings like "text/plain".

elharo: What about Files.probeContentType(Path)?

elharo: or URLConnection

google-cloud-storage/pom.xml
```diff
@@ -88,6 +88,11 @@
       <groupId>org.threeten</groupId>
       <artifactId>threetenbp</artifactId>
     </dependency>
+    <dependency>
```
elharo, line 91: isn't this included in java 6 and later?

dmitry-fa: it works with jdk7,8. This dependency is required for java9+

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
dmitry-fa: Both Files.probeContentType(Path) and URLConnection work! I will use it instead. Thanks for the hint.

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -376,9 +377,12 @@ public StorageObject create(
...
   private static String detectContentType(StorageObject object) {
     String contentType = object.getContentType();
-    return contentType != null
-        ? contentType
-        : MIMETYPES_FILE_TYPE_MAP.getContentType(object.getName().toLowerCase());
+    if (contentType != null) {
+      return contentType;
+    }
+    return firstNonNull(
+        FILE_NAME_MAP.getContentTypeFor(object.getName().toLowerCase()),
```
elharo, line 384: toLowerCase always should have a locale

dmitry-fa: done

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3273,4 +3273,45 @@ public void testBlobReload() throws Exception {
     updated.delete();
     assertNull(updated.reload());
   }
+
+  private Blob createBlob(String method, BlobInfo blobInfo) throws IOException {
+    switch (method) {
+      case "create":
+        return storage.create(blobInfo);
+      case "writer":
+        {
```
elharo, line 3282: why the braces here?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3273,4 +3273,45 @@ public void testBlobReload() throws Exception {
...
+    for (int i = 0; i < names.length; i++) {
+      BlobId blobId = BlobId.of(BUCKET, names[i]);
+      BlobInfo blobInfo = BlobInfo.newBuilder(blobId).build();
+      blob = createBlob(method, blobInfo);
+      assertEquals(types[i], blob.getContentType());
+    }
+    String customType = "custom/type";
+    BlobId blobId = BlobId.of(BUCKET, names[0]);
+    BlobInfo blobInfo = BlobInfo.newBuilder(blobId).setContentType(customType).build();
+    blob = createBlob(method, blobInfo);
```
elharo, line 3304: don't reuse local variable. There are two separate blob objects here that coincidentally share a variable.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3273,4 +3273,45 @@ public void testBlobReload() throws Exception {
...
+        }
+      default:
+        throw new IllegalArgumentException("Unknown method " + method);
+    }
+  }
+
+  private void testAutoContentType(String method) throws IOException {
+    String[] names = {"file1.txt", "dir with spaces/Pic.Jpg", "no_extension"};
+    String[] types = {"text/plain", "image/jpeg", "application/octet-stream"};
+    Blob blob = null;
```
elharo, line 3294: I think you can push this into the for loop

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
dmitry-fa: I had a reason to reuse the variable but after rewriting the test that reason is not actual anymore. Thanks for catching that.

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -372,6 +375,16 @@ public StorageObject create(
     }
   }
 
+  private static String detectContentType(StorageObject object) {
```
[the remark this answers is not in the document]
dmitry-fa: I was thinking about making this feature optional, but I didn't find an appropriate place to store this setting. It could be a property of a project/client/bucket/object.

I only found: https://cloud.google.com/storage/docs/metadata#content-type
which states:  `If the Content-Type is not specified by the uploader and cannot be determined, it is set to application/octet-stream or application/x-www-form-urlencoded, depending on how you uploaded the object.`

If I read this correctly, the Storage should attempt to determine the content type if not explicitly set, so this feature should be on by default.

BTW, right now the content type of blobs created with Storage.create() methods is `null`, but should be `application/octet-stream`.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3273,4 +3273,45 @@ public void testBlobReload() throws Exception {
...
+          storage.writer(blobInfo).close();
+          return storage.get(BlobId.of(blobInfo.getBucket(), blobInfo.getName()));
+        }
+      default:
+        throw new IllegalArgumentException("Unknown method " + method);
+    }
+  }
+
+  private void testAutoContentType(String method) throws IOException {
+    String[] names = {"file1.txt", "dir with spaces/Pic.Jpg", "no_extension"};
```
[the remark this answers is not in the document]
dmitry-fa: It's not always possible. In case of resumeable upload a blob is created first and the content is available later:

```
writer = storage.writer(blobInfo); // creating a blob, detecting content-type, no contect
writer.write(content);
writer.close();              
```

I did an experiment:
I created a file `PDF.txt` with a pdf content. Then I uploaded it as `PDF.jpg` with various clients and checked the resulting content type. I got the following:

gsutil:  text/plain
NodeJS:  image/jpeg
Go: application/pdf

To achieve consistent behavior across all the clients it should be implemented on the server-side at the moment when the upload is finished.

Until this feature is implemented I suggest type detecting based on file extension, as it has been done recently in nodejs (https://github.com/googleapis/nodejs-storage/pull/1190)

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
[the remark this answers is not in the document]
frankyn: Storage doesn't provide contentType detection in the API.

If contentType is set at creation then `application/octet-stream` is used.

Please make this optional otherwise it's a behavior change not originally intended on uploads with the Java client.

dmitry-fa:
@frankyn, 

For the time being, objects created with Storage.create() method have content-type set to null if not explicitly set, not `application/octet-stream`. 

Is it possible to store custom properties on the project level? 
I mean something like: `storage.setProperty('java.contentTypeDetection', 'extension')`

frankyn:
> If contentType is set at creation then application/octet-stream is used.
Whoops I meant, If contentType is *not* set at creation then application/octet-stream is used by the service.

dmitry-fa:
Not sure if it will be convenient to specify 'I want to detect the content type' each time when a new object is created. The content type could be provided explicitly instead. 

Okay, I'll think about that. If the user specifies explicitly that he wants to detect the content type, he can also select a strategy for that: 'by extension', 'by content' or 'by both'.

PR #348, opened 2020-06-04
chore: generate CODEOWNERS from templates

chingor13:
CODEOWNERS is now managed by templates and we can configure the default team in the .repo-metadata.json file.


czahedi:
This is super cool! LGTM though Kokoro is grumpy (seems unrelated)

PR #351, opened 2020-06-05
feat: stub implementation of StorageRpc for the sake of testing

### LINKED ISSUE #242: Merge Storage-NIO library into this repository.
Currently have two repositories that must  be merged into one library.

https://github.com/googleapis/java-storage-nio needs to be merged into this repository to unblock changes to the underlying internal only StorageRpc interface. Otherwise we can accidentally break users.

dmitry-fa:
Allows extending of the StorageRpc interface without breaking cross library dependencies.
Tests outside the Storage library might extend the StorageRpcTestBase class instead of implementing the StorageRpc interface.

Fixes #242

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/testing/StorageRpcTestBase.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/testing/StorageRpcTestBase.java
@@ -1,0 +1,505 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage.testing;
+
+import com.google.api.services.storage.model.Bucket;
+import com.google.api.services.storage.model.BucketAccessControl;
+import com.google.api.services.storage.model.HmacKey;
+import com.google.api.services.storage.model.HmacKeyMetadata;
+import com.google.api.services.storage.model.Notification;
+import com.google.api.services.storage.model.ObjectAccessControl;
+import com.google.api.services.storage.model.Policy;
+import com.google.api.services.storage.model.ServiceAccount;
+import com.google.api.services.storage.model.StorageObject;
+import com.google.api.services.storage.model.TestIamPermissionsResponse;
+import com.google.cloud.Tuple;
+import com.google.cloud.storage.spi.v1.RpcBatch;
+import com.google.cloud.storage.spi.v1.StorageRpc;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/testing/StorageRpcTestBase.java
```diff
@@ -0,0 +1,505 @@
...
+import com.google.cloud.Tuple;
+import com.google.cloud.storage.spi.v1.RpcBatch;
+import com.google.cloud.storage.spi.v1.StorageRpc;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A stub implementation of {@link StorageRpc} which could be used outside of the Storage module for
```
elharo, line 38: could --> can

google-cloud-storage/src/main/java/com/google/cloud/storage/testing/StorageRpcTestBase.java
```diff
@@ -0,0 +1,505 @@
...
+ * testing purposes. All the methods are implemented either to return the default value or to throw
```
elharo, line 39: "are implemented either to return the default value or to throw" --> "either return the default value or throw"

google-cloud-storage/src/main/java/com/google/cloud/storage/testing/StorageRpcTestBase.java
```diff
@@ -0,0 +1,505 @@
...
+  /** Creates an instance with methods returning the default value. */
+  public StorageRpcTestBase() {
+    this(true);
+  }
+
+  /**
+   * Creates an instance with methods either returning the default value or throwing {@code
+   * UnsupportedOperationException}, depending on the {@code isImplemented} parameter.
+   *
+   * @param isImplemented {@code true} to return a value, {@code false} to throw an exception.
```
elharo, line 58: nit: no period since not a complete sentence

google-cloud-storage/src/main/java/com/google/cloud/storage/testing/StorageRpcTestBase.java
```diff
@@ -0,0 +1,505 @@
...
+      return null;
+    } else {
+      throw new UnsupportedOperationException("Not implemented yet");
+    }
+  }
+
+  @Override
+  public StorageObject compose(
+      Iterable<StorageObject> sources, StorageObject target, Map<Option, ?> targetOptions) {
+    if (isImplemented) {
```
elharo, line 166: Why is "isImplemented" needed? If you're just returning null here and elsewhere and not expecting these methods to be called, unconditionally throwing the exception might be better.

dmitry-fa: Yes, that was my original intention to throw the exception unconditionally. Then I decided that adding "isImplemented" feature might simplify the work on extending that class.

elharo: It doesn't seem like returning null helps a lot in this case. Subclassers who invoke a method like this will still need to override this.

dmitry-fa: I tend to agree with you, but I will think more about cases when returning null is useful.

dmitry-fa: Fixed as you suggested. If someone needs an implementation returning null, an IDE can help to produce such class.

PR #340, opened 2020-05-27
deps: bump shared-deps version and add back certain test deps

stephaniewang526:

PR #366, opened 2020-06-15
fix: update dependencies

frankyn:


frankyn:
Missed: #340 going to use that instead.

PR #390, opened 2020-06-23
feat: Add support to disable logging from bucket

### LINKED ISSUE #389: Disable logging from bucket is not working
I was successfully able to add the logging to a bucket.

```javascript
BucketInfo.Logging logging =
          BucketInfo.Logging.newBuilder()
              .setLogBucket(logsBucket)
              .setLogObjectPrefix("test-logs")
              .build();
      Bucket bucket =
          storage.create(
              BucketInfo.newBuilder(loggingBucket).setLocation("us").setLogging(logging).build());
```
But seems there is no way to disable the logging from bucket

`Bucket updatedBucket = bucket.toBuilder().setLogging(null).build().update();`

Looks like there is no support in SDK for PATCH update where as Rest API for Disabling logging.

https://cloud.google.com/storage/docs/access-logs#disabling

```
PATCH /example-bucket?logging HTTP/1.1
Host: storage.googleapis.com

{
 "logging": null
}
```

~cc: @frankyn

athakor:
Fixes #389

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -1310,7 +1310,7 @@ public class BucketInfo implements Serializable {
 
     @Override
     public Builder setLogging(Logging logging) {
-      this.logging = logging;
+      this.logging = logging != null ? logging : null;
       return this;
     }
 
@@ -1758,11 +1758,8 @@ public class BucketInfo implements Serializable {
     if (iamConfiguration != null) {
       bucketPb.setIamConfiguration(iamConfiguration.toPb());
     }
-    if (logging != null) {
-      bucketPb.setLogging(logging.toPb());
-    } else {
-      bucketPb.setLogging(Data.nullOf(Bucket.Logging.class));
-    }
+    Bucket.Logging loggingPb = logging != null ? logging.toPb() : Data.nullOf(Bucket.Logging.class);
+    bucketPb.setLogging(loggingPb);
     return bucketPb;
   }
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -983,6 +983,14 @@ public class BucketInfo implements Serializable {
     public abstract Builder deleteLifecycleRules();
 
     /**
+     * Disable access logging for this bucket.
+     *
+     * @see <a href="https://cloud.google.com/storage/docs/access-logs#disabling">Disabling
+     *     logging</a>
+     */
+    public abstract Builder disableLogging();
+
+    /**
      * Sets the bucket's storage class. This defines how blobs in the bucket are stored and
      * determines the SLA and the cost of storage. A list of supported values is available <a
      * href="https://cloud.google.com/storage/docs/storage-classes">here</a>.
@@ -1200,6 +1208,12 @@ public class BucketInfo implements Serializable {
     public Builder deleteLifecycleRules() {
       setDeleteRules(null);
       setLifecycleRules(null);
+      return this;
+    }
+
+    @Override
+    public Builder disableLogging() {
+      setLogging(null);
       return this;
     }
 
@@ -1760,6 +1774,8 @@ public class BucketInfo implements Serializable {
     }
     if (logging != null) {
       bucketPb.setLogging(logging.toPb());
+    } else {
+      bucketPb.setLogging(Data.nullOf(Bucket.Logging.class));
     }
     return bucketPb;
   }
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1760,6 +1760,8 @@ public Rule apply(LifecycleRule lifecycleRule) {
     }
     if (logging != null) {
       bucketPb.setLogging(logging.toPb());
+    } else {
+      bucketPb.setLogging(Data.nullOf(Bucket.Logging.class));
```
frankyn, line 1764: This can have unexpected behavior. Instead could you update setLogging() builder method to include this if else condition when `null` is provided?

athakor: done

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1310,7 +1310,7 @@ public Builder setIamConfiguration(IamConfiguration iamConfiguration) {
 
     @Override
     public Builder setLogging(Logging logging) {
-      this.logging = logging;
+      this.logging = logging != null ? logging : null;
```
frankyn, line 1313: Instead do:

```
this.logging = logging != null ? logging : Data.nullOf(Bucket.Logging.class);
```

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1758,9 +1758,8 @@ public Rule apply(LifecycleRule lifecycleRule) {
     if (iamConfiguration != null) {
       bucketPb.setIamConfiguration(iamConfiguration.toPb());
     }
-    if (logging != null) {
-      bucketPb.setLogging(logging.toPb());
-    }
+    Bucket.Logging loggingPb = logging != null ? logging.toPb() : Data.nullOf(Bucket.Logging.class);
```
frankyn, line 1761: Revert this change:
```
 Bucket.Logging loggingPb = logging != null ? logging.toPb() : Data.nullOf(Bucket.Logging.class);
```

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
athakor: @frankyn 
I think we have to go with the `null ` instead of `Data.nullOf(Bucket.Logging.class)` because if we do as you said then the following error occurs :
`unable to create new instance of class com.google.cloud.storage.BucketInfo$Logging because it has no accessible default constructor`

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
athakor: I think we also have to keep this check because when the object of `logging `is `null`, the `logging.toPb()` will throw the nullPointer exception.

frankyn: https://github.com/googleapis/java-storage/compare/qlogics?expand=1

Here's alternative solution. If you have concerns with it always open to discuss further. 

I'd like to address the issue with logging field is not selected in Fields. In which case it comes back null and this PR can corrupt the logging field when not selected.

For example this will corrupt logging metadata:
```
      BucketInfo.Logging logging =
          BucketInfo.Logging.newBuilder()
              .setLogBucket(logsBucket)
              .setLogObjectPrefix("test-logs")
              .build();
      Bucket bucket =
          storage.create(
              BucketInfo.newBuilder(loggingBucket).setLocation("us").setLogging(logging).build());
      // Perform Get Request without selecting logging field
      Bucket remoteBucket = storage.get(loggingBucket);
      // Don't update logging and only perform an update.
      Bucket updatedBucket = bucket.toBuilder().build().update();
      // Logging is now null.
```


Additionally could you remove setting IAM policy to allow writes from allAuthenticatedUsers. This is a security risk:
```
          storage.setIamPolicy(
              logsBucket,
              policy
                  .toBuilder()
                  .addIdentity(StorageRoles.legacyBucketWriter(), Identity.allAuthenticatedUsers())
                  .build()));

```

athakor: @frankyn good catch and thanks for the detailed description. 

Both the comments have been addressed PTAL when get chance.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -286,9 +286,14 @@ public String getLogObjectPrefix() {
     }
 
     Bucket.Logging toPb() {
-      Bucket.Logging logging = new Bucket.Logging();
-      logging.setLogBucket(logBucket);
-      logging.setLogObjectPrefix(logObjectPrefix);
+      Bucket.Logging logging;
+      if (logBucket != null && logObjectPrefix != null) {
```
frankyn, line 290: This should be an ||  an or.

athakor: done

PR #398, opened 2020-06-29
fix: PostPolicyV4.PostFieldsV4.Builder.addCustomMetadataField() allows to add prefixed an not prefixed custom fields

### LINKED ISSUE #394: Support: Accept x-goog-meta-* as metadata value in PostPolicyV4
Allow field parameter to pass in x-goog-meta-* as field name:

https://github.com/googleapis/java-storage/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java#L139

It would support both x-goog-meta-* prefixed fields and not prefixed fields.

dmitry-fa:
Fixes #394

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
@@ -143,13 +143,13 @@ public final class PostPolicyV4 {
         return this;
       }
 
-      /** @deprecated use {@link #addCustomMetadataField(String, String)} */
+      /** @deprecated use {@link #setCustomMetadataField(String, String)} */
       @Deprecated
       public Builder AddCustomMetadataField(String field, String value) {
-        return addCustomMetadataField(field, value);
+        return setCustomMetadataField(field, value);
       }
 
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
```diff
@@ -136,8 +143,17 @@ public Builder setSuccessActionStatus(int successActionStatus) {
...
       }
 
+      /** @deprecated use {@link #addCustomMetadataField(String, String)} */
+      @Deprecated
       public Builder AddCustomMetadataField(String field, String value) {
-        fieldsMap.put("x-goog-meta-" + field, value);
+        return addCustomMetadataField(field, value);
+      }
+
+      public Builder addCustomMetadataField(String field, String value) {
```
frankyn, line 152: use setCustomMetadataField

dmitry-fa: @frankyn 
done

dmitry-fa: @frankyn: gentle ping

PR #399, opened 2020-06-30
feat: expose the methods of Notifications

### LINKED ISSUE #138: Expose the methods of Notifications

athakor:
Addressed the feedback of https://github.com/googleapis/java-storage/pull/393

Fixes #138


google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,139 @@
...
+
+  private final StorageOptions options;
+  private transient Storage storage;
+
+  /** Builder for {@code Notification}. */
+  public static class Builder extends NotificationInfo.Builder {
+    private final Storage storage;
+    private final NotificationInfo.BuilderImpl infoBuilder;
+
+    Builder(Notification notification) {
```
dmitry-fa, line 39: this ctor could be made private

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,139 @@
...
+      return this;
+    }
+
+    @Override
+    public Notification build() {
+      return new Notification(storage, infoBuilder);
+    }
+  }
+
+  Notification(Storage storage, NotificationInfo.BuilderImpl infoBuilder) {
```
dmitry-fa, line 98: this ctor could be made private as well

google-cloud-storage/src/main/java/com/google/cloud/storage/NotificationInfo.java
```diff
@@ -0,0 +1,342 @@
...
+    public abstract Builder setEtag(String etag);
+
+    public abstract Builder setCustomAttributes(Map<String, String> customAttributes);
+
+    /** Creates a {@code NotificationInfo} object. */
+    public abstract NotificationInfo build();
+  }
+
+  public static final class BuilderImpl extends Builder {
+
```
dmitry-fa, line 96: Why do you need ` public abstract class Builder` and `public static final class BuilderImpl extends Builder`?
I guess just `public static final class Builder` will be enough

google-cloud-storage/src/main/java/com/google/cloud/storage/NotificationInfo.java
```diff
@@ -0,0 +1,342 @@
...
+    generatedId = builder.generatedId;
+    etag = builder.etag;
+    selfLink = builder.selfLink;
+    topic = builder.topic;
+    eventTypes = builder.eventTypes;
+    customAttributes = builder.customAttributes;
+    payloadFormat = builder.payloadFormat;
+    objectNamePrefix = builder.objectNamePrefix;
+  }
+
```
dmitry-fa, line 188: This ctor should be private,
you need `public static Builder newBuilder()`

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,139 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Google Storage Notification object;
+ *
+ * @see <a href="https://cloud.google.com/storage/docs/json_api/v1/notifications">notifications</a>
+ */
```
dmitry-fa, line 28: `Google Storage Notification object;` - is not clear at all. Storage does not have Notifications, Pub/Sub does.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,36 @@ HmacKeyMetadata updateHmacKeyState(
...
+  Notification createNotification(String bucket, NotificationInfo notificationInfo);
+
+  /**
+   * Get the notification with the specified name on the specified object.
+   *
+   * @return the notification object that exist on the bucket.
+   * @throws StorageException upon failure
+   */
+  Notification getNotification(String bucket, String notification);
+
```
dmitry-fa, line 3566: Very confusing doc. I guess you mean bucket, not object. You mention 'the specified name' and parameter 'notification' without any documentation. 
It's not clear what is 'notification name', may be you mean 'topic'?

`the notification object that exist on the bucket.`  What happens if the object does not exist? 
A good spec should cover that.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,36 @@ HmacKeyMetadata updateHmacKeyState(
...
    */
   ServiceAccount getServiceAccount(String projectId);
+
+  /**
+   * Creates a notification with the specified entity on the specified bucket.
+   *
+   * @return the notification that was created.
+   * @throws StorageException upon failure
+   */
+  Notification createNotification(String bucket, NotificationInfo notificationInfo);
```
dmitry-fa, line 3557: Does `creates a notification` mean subscription?

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,36 @@ HmacKeyMetadata updateHmacKeyState(
...
+  /**
+   * Get the notification with the specified name on the specified object.
+   *
+   * @return the notification object that exist on the bucket.
+   * @throws StorageException upon failure
+   */
+  Notification getNotification(String bucket, String notification);
+
+  /**
+   * List the notifications for the provided bucket.
```
dmitry-fa, line 3568: List -> Lists

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,36 @@ HmacKeyMetadata updateHmacKeyState(
...
+   * @return a list of {@link Notification} objects that exist on the bucket.
+   * @throws StorageException upon failure
+   */
+  List<Notification> listNotifications(String bucket);
+
+  /**
+   * Deletes the notification with the specified name on the specified object.
+   *
+   * @return {@code true} if the notification was deleted, {@code false} if it was not found
+   * @throws StorageException upon failure
```
dmitry-fa, line 3579: Does `delete` mean unsubscribe?

the specified object -> the bucket
was deleted -> has been deleted
if it was not found -> if not found

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/StorageRpc.java
```diff
@@ -553,6 +553,14 @@ TestIamPermissionsResponse testIamPermissions(
...
 
+  /**
+   * Get the notification with the specified name on the specified object.
+   *
+   * @return the notification object that exist on the bucket.
+   * @throws StorageException upon failure
+   */
+  Notification getNotification(String bucket, String notification);
+
   /**
```
dmitry-fa, line 564: Get -> Gets
clarify what is notification name
what happens if no notification exists

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
athakor: yes

dmitry-fa: It should be reflected in the spec.

google-cloud-storage/src/main/java/com/google/cloud/storage/NotificationInfo.java
elharo: Overuse of abstract classes and interfaces is why CLIRR is giving you troubles. Prefer concrete classes.

athakor:
@dmitry-fa all the comments have been addressed PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,313 @@
...
+
+    return notificationPb;
+  }
+
+  /**
+   * Creates a {@code Notification} object for the provided topic name.
+   *
+   * @param topic The name of the topic. It must have the format
+   *     "projects/{project}/topics/{topic}".
+   */
```
dmitry-fa, line 263: provided topic name -> provided topic
@param topic a string in the format "projects/{project}/topics/{topic}"
(no . at the end)

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  /**
+   * Returns a {@code Notification} builder where the topic's name is set to the provided name.
+   *
+   * @param topic The name of the topic. It must have the format
+   *     "projects/{project}/topics/{topic}".
+   */
+  public static Builder newBuilder(String topic) {
+    PATH_TEMPLATE.validatedMatch(topic, "topic name must be in valid format");
```
dmitry-fa, line 281: topic must be in the valid format

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,313 @@
...
+    public Builder setSelfLink(String selfLink) {
+      this.selfLink = selfLink;
+      return this;
+    }
+
+    /** The name of the topic. It must have the format "projects/{project}/topics/{topic}". */
+    public Builder setTopic(String topic) {
+      this.topic = topic;
+      return this;
+    }
```
dmitry-fa, line 114: You need to to validate that the given topic meets the format

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,313 @@
...
+   * Creates a {@code Notification} object for the provided topic name.
+   *
+   * @param topic The name of the topic. It must have the format
+   *     "projects/{project}/topics/{topic}".
+   */
+  public static Notification of(String topic) {
+    PATH_TEMPLATE.validatedMatch(topic, "topic name must be in valid format");
+    return newBuilder(topic).build();
+  }
+
```
dmitry-fa, line 268: Please consider creating a private static function to validate the format. This function will be called each time the topic is set.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,43 @@ HmacKeyMetadata updateHmacKeyState(
...
+
+  /**
+   * Creates a notification subscription for a given bucket.
+   *
+   * @param bucket name of the bucket
+   * @param notification a {@code Notification} object
+   * @return the notification that was created
+   * @throws StorageException upon failure
+   */
+  Notification createNotification(String bucket, Notification notification);
```
dmitry-fa, line 3559: I can't understand the logic here: what is the difference between the input notification and the returned one.
Intuitively I think that this method is about to subscribe. If so it should look like:

`Notification subscribe(String bucket, SubscriptionRequest request)` or may be:
`Notification subscribe(String bucket, String topic)`

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,43 @@ HmacKeyMetadata updateHmacKeyState(
...
+
+  /**
+   * Deletes the notification subscription with the specified name on the bucket.
+   *
+   * @param bucket name of the bucket
+   * @param notification ID of the notification to delete
+   * @return {@code true} if the notification has been deleted, {@code false} if not found
+   * @throws StorageException upon failure
+   */
+  boolean deleteNotification(String bucket, String notification);
```
dmitry-fa, line 3588: please rename `notification` param to make it clear what it stands for. id, name, or topic.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
athakor: @dmitry-fa most of the implementation pull out from this artifact https://github.com/Googleapis/java-notification.
It was implemented in a separate artifact and customers would benefit from having it in the same artifact.

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
athakor: same thing we are doing here [see](https://googleapis.github.io/api-common-java/1.1.0/apidocs/com/google/api/pathtemplate/PathTemplate.html) and i don't think so we need to create an extra function here.

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
athakor: yes. we can validate in` build()` instead of here.

athakor:
@dmitry-fa PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
dmitry-fa: I meant here that instead of calling `PATH_TEMPLATE.validatedMatch(topic, "topic name must be in valid format");` multiple times it's better to introduce:

private static void checkTopicFormat(String topic) {
   PATH_TEMPLATE.validatedMatch(topic, "topic name must be in valid format");
}

and call this function.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
dmitry-fa: The major problem here not in the method name, but in the description. I can't understand how I can use this new API.

New introduced entity `Notification` is described as:

```
/**
 * Google Storage Notification metadata;
 *
 * @see <a href="https://cloud.google.com/storage/docs/concepts-techniques#concepts">Concepts and
 *     Terminology</a>
 */
```
The link points to a page which doesn't mention 'Notification', so this link is not very helpful.

`createNotification` name sounds confusing to me. Probably it will sound confusing to others.  A few more words describing how to use this method will be useful. An example of use would be awesome.

athakor:
@dmitry-fa PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,320 @@
...
+/**
+ * A {@code Notification} object sends the information about the changes of objects in your buckets,
+ * where the information is added to the topic of your choice in the form of messages. For example,
+ * you can track objects that are created and deleted in your bucket. Each notification contains
+ * information describing both the event that triggered it and the object that changed.
+ *
+ * <p>You can send notifications to any topic in any project for which you have sufficient
+ * permissions. Once received by the topic, the resulting message can be sent to any number of
+ * subscribers to the topic.
+ */
```
dmitry-fa, line 39: This is much much better!

Would it make sense to say: 

```
/**
  * The class representing Pub/Sub notifications for the Storage.
  * See <a
 * href="https://cloud.google.com/storage/docs/pubsub-notifications">https://cloud.google.com/storage/docs/pubsub-notifications</a>
 * for details.
 */
```

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
dmitry-fa: I think the right name for the method is `addNotification()`
Actually this method adds a new notification to the list of bucket's notifications.
I guess the difference between input and output notifications - the output has the server-generated id set.

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,320 @@
...
+
+    public Builder setObjectNamePrefix(String objectNamePrefix) {
+      this.objectNamePrefix = objectNamePrefix;
+      return this;
+    }
+
+    public Builder setEventTypes(Iterable<String> eventTypes) {
+      this.eventTypes = eventTypes != null ? ImmutableList.copyOf(eventTypes) : null;
+      return this;
+    }
```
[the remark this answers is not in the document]
athakor: Here we cannot do that because `setEventTypes `requires the list parameter you can see  [Notification#setEventTypes](https://developers.google.com/resources/api-libraries/documentation/storage/v1/java/latest/com/google/api/services/storage/model/Notification.html#setEventTypes-java.util.List-) of underlying Google API client object model.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
athakor: we could give that name as well but according to api doc description `createNotification `looks more informative here.

dmitry-fa: createNotification is the internal name for an RPC call. Methods in the Storage interface could be named independently.

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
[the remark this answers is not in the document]
dmitry-fa: `EventType...`  is an array of EventType. You can covert it to the list of String inside of the method. It's much convenient for the user to call `setEventTypes(TYPE1, TYPE2)` rather than create Iterable<String>. Because the set of values is limited, please use an enum to list all the valid values.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,83 @@ HmacKeyMetadata updateHmacKeyState(
...
+   *
+   * @param bucket name of the bucket
+   * @param notification a {@code Notification} object
+   * @return the notification that was created
+   * @throws StorageException upon failure
+   */
+  Notification createNotification(String bucket, Notification notification);
+
+  /**
+   * Gets the notification with the specified name on the bucket.
```
dmitry-fa, line 3574: name -> id

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,83 @@ HmacKeyMetadata updateHmacKeyState(
    * @throws StorageException upon failure
    */
   ServiceAccount getServiceAccount(String projectId);
+
+  /**
+   * Creates a notification subscription for a given bucket.
```
dmitry-fa, line 3552: Adds a new notification to the bucket.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,83 @@ HmacKeyMetadata updateHmacKeyState(
...
+   *
+   * <p>Example of creating a notification subscription:
```
dmitry-fa, line 3554: Example of adding a notification:

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,83 @@ HmacKeyMetadata updateHmacKeyState(
...
+   *
+   * <pre>{@code
+   * String bucketName = "my-unique-bucket";
+   * String topic = "projects/myProject/topics/topic"
```
dmitry-fa, line 3558: /topic -> /myTopic

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,83 @@ HmacKeyMetadata updateHmacKeyState(
...
+   * Notification bucketNotification = Notification.newBuilder(topic)
+   *  .setCustomAttributes(ImmutableMap.of("label1", "value1"))
+   *  .setPayloadFormat(Notification.PayloadFormat.JSON_API_V1.JSON_API_V1)
+   *  .build();
+   * Notification notification = storage.createNotification(bucketName, bucketNotification);
+   * }</pre>
```
dmitry-fa, line 3564: please use setEventType() in the example, it's very informative method.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,83 @@ HmacKeyMetadata updateHmacKeyState(
...
+   *
+   * @param bucket name of the bucket
+   * @param notification a {@code Notification} object
+   * @return the notification that was created
```
dmitry-fa, line 3568: @return the created notification

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,83 @@ HmacKeyMetadata updateHmacKeyState(
...
+   *
+   * @param bucket name of the bucket
+   * @param notificationId notification ID
+   * @return the notification object that exist on the bucket or {@code null} if not found
+   * @throws StorageException upon failure
+   */
+  Notification getNotification(String bucket, String notificationId);
+
+  /**
+   * Retrieves a list of notification subscriptions for a given bucket.
```
dmitry-fa, line 3592: Retrieves the list of notifications associated with the bucket.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3547,4 +3547,83 @@ HmacKeyMetadata updateHmacKeyState(
...
+   * }</pre>
+   *
+   * @param bucket name of the bucket
+   * @return a list of {@link Notification} objects that exist on the bucket.
+   * @throws StorageException upon failure
+   */
+  List<Notification> listNotifications(String bucket);
+
+  /**
+   * Deletes the notification subscription with the specified name on the bucket.
```
dmitry-fa, line 3608: Deletes the notification with the specified id.

athakor:
@dmitry-fa all the comments have been addressed PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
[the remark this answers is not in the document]
athakor: done

dmitry-fa:
@athakor, thanks for addressing my all my comments. At first glance everything looks good. I'm going to perform an overall review of the entire changes.

google-cloud-storage/src/main/java/com/google/cloud/storage/Notification.java
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  /**
+   * Returns the list of event types that this notification will apply to. If empty, notifications
+   * will be sent on all event types.
+   *
+   * @see <a href="https://cloud.google.com/storage/docs/cross-origin">Cross-Origin Resource Sharing
+   *     (CORS)</a>
+   */
+  public EventType[] getEventTypes() {
```
frankyn, line 209: What about a List return instead of array?

athakor:
@frankyn all the nits have been addressed PTAL when you get chance.

athakor:
> I guess @frankyn meant converting the array to list inside of the setEventTypes(EventType...) method, not replacing argument type.

@dmitry-fa done

google-cloud-storage/src/test/java/com/google/cloud/storage/NotificationTest.java
```diff
@@ -0,0 +1,105 @@
...
+          .build();
+
+  @Test
+  public void testToBuilder() {
+    compareBucketsNotification(NOTIFICATION, NOTIFICATION.toBuilder().build());
+    Notification notification = NOTIFICATION.toBuilder().setGeneratedId("id").build();
+    assertEquals("id", notification.getGeneratedId());
+    notification = notification.toBuilder().setGeneratedId(GENERATED_ID).build();
+    compareBucketsNotification(NOTIFICATION, notification);
+  }
```
dmitry-fa, line 57: the generated id should be never set manually.
Tests should verify that this field is set after StorageRpc.createNotification() is invoked.

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplMockitoTest.java
```diff
@@ -360,6 +361,39 @@ public long millisTime() {
           .put('~', "~")
           .build();
 
+  // Notification
+  private static final String ETAG = "0xFF00";
+  private static final String GENERATED_ID = "B/N:1";
```
dmitry-fa, line 366: GENERATED_ID should be assigned after addNotification() method is invoked.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -143,6 +148,8 @@
 
   private static RemoteStorageHelper remoteStorageHelper;
   private static Storage storage;
+  private static Notification notification;
+  private static TopicAdminClient topicAdminClient;
```
dmitry-fa, line 152: these fields could be non static

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -214,6 +229,10 @@ public static void beforeClass() throws IOException {
 
     // Prepare KMS KeyRing for CMEK tests
     prepareKmsKeys();
+
+    // Add notification.
+    topicAdminClient = TopicAdminClient.create();
+    addNotification(topicAdminClient);
```
dmitry-fa, line 235: more clear would be:
notification = addNotification(topicAdminClient);

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -382,6 +406,36 @@ private static String ensureKmsKeyExistsForTests(
...
+
+    // Add a new notification to the bucket.
+    Notification bucketNotification =
+        Notification.newBuilder(TOPIC)
+            .setCustomAttributes(CUSTOM_ATTRIBUTES)
+            .setPayloadFormat(PAYLOAD_FORMAT)
+            .build();
+    notification = storage.addNotification(BUCKET, bucketNotification);
+    return notification;
+  }
```
dmitry-fa, line 431: it's better:
return storage.addNotification(BUCKET, bucketNotification);

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3364,4 +3418,25 @@ public void testUploadWithEncryption() throws Exception {
...
+  public void testListNotification() {
+    List<Notification> notifications = storage.listNotifications(BUCKET);
+    for (Notification actualNotification : notifications) {
+      if (actualNotification.getGeneratedId().equals(notification.getGeneratedId())) {
+        assertEquals(CUSTOM_ATTRIBUTES, actualNotification.getCustomAttributes());
+        assertEquals(PAYLOAD_FORMAT.name(), actualNotification.getPayloadFormat().name());
+        assertTrue(actualNotification.getTopic().contains(TOPIC));
+      }
+    }
+  }
```
dmitry-fa, line 3441: Perhaps the following test scenario will cover the notification behavior better:

- check that listNotification() returns empty list
- add new notification and check:
   * id is not null
   * other fields are expected
   * listNotification includes just created one
   * getNotification(id) returns it
- deleteNotification() and then check:
   * listNotification() returns empty list
   * getNotification(id) returns null

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
athakor: It should be static due to  
```
@BeforeClass
  public static void beforeClass() throws IOException {}
```

dmitry-fa: yes, agree

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
athakor: done

athakor:
@frankyn @dmitry-fa all the comments have been addressed PTAL

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3420,23 +3407,37 @@ public void testUploadWithEncryption() throws Exception {
...
-      }
+  public void testNotification() throws InterruptedException, ExecutionException {
+    String bucketName = RemoteStorageHelper.generateBucketName();
+    Bucket bucket = storage.create(BucketInfo.newBuilder(bucketName).setLocation("us").build());
+    Notification bucketNotification =
+        Notification.newBuilder(TOPIC)
+            .setCustomAttributes(CUSTOM_ATTRIBUTES)
+            .setPayloadFormat(PAYLOAD_FORMAT)
+            .build();
+    try {
```
dmitry-fa, line 3418: you can check that notificationList is empty at this point.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3420,23 +3407,37 @@ public void testUploadWithEncryption() throws Exception {
...
+
+      // Gets the notification with the specified id.
+      Notification actualNotification =
+          storage.getNotification(bucketName, notification.getGeneratedId());
+      assertEquals(notification, actualNotification);
+
+      // Retrieves the list of notifications associated with the bucket.
+      List<Notification> notifications = storage.listNotifications(bucketName);
+      assertEquals(1, notifications.size());
+
```
dmitry-fa, line 3433: please check that notifications.get(0).id() == actualNotification.id()

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3420,23 +3407,37 @@ public void testUploadWithEncryption() throws Exception {
...
+      // Deletes the notification with the specified id.
+      boolean isDeleted = storage.deleteNotification(bucketName, notification.getGeneratedId());
+      assertThat(isDeleted).isTrue();
+      assertThat(storage.getNotification(bucketName, notification.getGeneratedId())).isNull();
+      assertThat(storage.listNotifications(bucketName)).isEmpty();
+    } finally {
```
dmitry-fa, line 3439: add check that storage.getNotification(bucketName, notification.getGeneratedId()) returns null.

athakor: its already there `assertThat(storage.getNotification(bucketName, notification.getGeneratedId())).isNull();`

dmitry-fa: yes, I missed that

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
athakor: done

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
athakor: done

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3420,4 +3461,42 @@ public void testAutoContentTypeCreateFrom() throws IOException {
...
+
+      // Retrieves the list of notifications associated with the bucket.
+      List<Notification> notifications = storage.listNotifications(bucketName);
+      assertThat(notifications.size()).isEqualTo(1);
+      assertThat(notifications.get(0).getGeneratedId())
+          .isEqualTo(actualNotification.getGeneratedId());
+
+      // Deletes the notification with the specified id.
+      boolean isDeleted = storage.deleteNotification(bucketName, notification.getGeneratedId());
+      assertThat(isDeleted).isTrue();
```
dmitry-fa, line 3495: The final nit from me:
      boolean isDeleted = storage.deleteNotification(bucketName, notification.getGeneratedId());
      assertThat(isDeleted).isTrue();
->
      assertThat(storage.deleteNotification(bucketName, notification.getGeneratedId())).isTrue();
      assertThat(storage.deleteNotification(bucketName, notification.getGeneratedId())).isFalse();

athakor: done

athakor:
@frankyn  please take a final look when you get chance

google-cloud-storage/clirr-ignored-differences.xml
```diff
@@ -21,4 +21,29 @@
         <method>com.google.cloud.storage.BucketInfo$Builder deleteLifecycleRules()</method>
         <differenceType>7013</differenceType>
     </difference>
+    <difference>
+        <className>com/google/cloud/storage/Storage</className>
```
elharo, line 25: This requires a major version bump.

athakor: > This requires a major version bump.

@frankyn what do you think on this??

frankyn: We've done it for others and updated the README.md to reflect this quirk in updates of the Interface. We haven't hit customer issues with these interface changes.

We are going to move forward with similar changes.

frankyn:
Thanks for your patience folks, I want @JesseLovelace to have a look before merging this PR into master.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3589,4 +3589,84 @@ HmacKeyMetadata updateHmacKeyState(
...
+   * <p>Example of adding a notification:
+   *
+   * <pre>{@code
+   * String bucketName = "my-unique-bucket";
+   * String topic = "projects/myProject/topics/myTopic"
+   * Notification bucketNotification = Notification.newBuilder(topic)
+   *  .setCustomAttributes(ImmutableMap.of("label1", "value1"))
+   *  .setEventTypes(Notification.EventType.OBJECT_FINALIZE)
+   *  .setPayloadFormat(Notification.PayloadFormat.JSON_API_V1)
+   *  .build();
```
frankyn, line 3605: The issue with the current flow is that Notification instances are initialized using a topic but not the id. 

Additionally, I can't tell when I have an object that's returned by the Storage API or the library. For example BlobInfo is user generated while Blob is returned by the Storage API. The same pattern is used for BucketInfo and Bucket. 

I think the path forward would be to do something similar. Pending input @JesseLovelace.

athakor: okay so do you like me to add the similar kind of pattern here correct?  for example `NotificationInfo `is user generated while `Notification `is returned by the Storage API.

google-cloud-storage/src/main/java/com/google/cloud/storage/NotificationInfo.java
```diff
@@ -0,0 +1,362 @@
...
+      topic = notificationInfo.topic;
+      eventTypes = notificationInfo.eventTypes;
+      customAttributes = notificationInfo.customAttributes;
+      payloadFormat = notificationInfo.payloadFormat;
+      objectNamePrefix = notificationInfo.objectNamePrefix;
+    }
+
+    @Override
+    Builder setGeneratedId(String generatedId) {
+      this.generatedId = generatedId;
```
frankyn, line 130: use `notificationId` instead.
Is this meant to be a String or int?

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3589,4 +3589,84 @@ HmacKeyMetadata updateHmacKeyState(
...
+   * Notification notification = storage.addNotification(bucketName, notificationInfo);
+   * }</pre>
+   *
+   * @param bucket name of the bucket
+   * @param notificationInfo notification to create
+   * @return the created notification
+   * @throws StorageException upon failure
+   */
+  Notification addNotification(String bucket, NotificationInfo notificationInfo);
```
frankyn, line 3614: `createNotification` instead

google-cloud-storage/src/main/java/com/google/cloud/storage/NotificationInfo.java
athakor: its String

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java

athakor:
@frankyn @JesseLovelace PTAL

athakor:
@frabkyn @JesseLovelace  gentle ping

athakor:
@frankyn @JesseLovelace friendly ping

athakor:
@frankyn @JesseLovelace friendly ping

frankyn (approved):

Issue #395, opened 2020-06-25
Storage: support customTime metadata and custom timestamp based OLM


jkwlui:
Request to support a new `customTime` metadata and associated custom timestamp based OLM conditions:

```java
DateTime customTimeBefore = DateTime.parseRfc3339("2020-01-01T00:00:00Z");

bucket
    .toBuilder()
    .setLifecycleRules(
        ImmutableList.of(
            new LifecycleRule(
                LifecycleRule.LifecycleAction.newDeleteAction(),
                LifecycleRule.LifecycleCondition.newBuilder()
                    .setDaysSinceCustomTime(30)
                    .setCustomTimeBefore(customTimeBefore)
                    .build())))
    .build()
    .update();

// New metadata field:
Map<String, String> newMetadata = new HashMap<>();
newMetadata.put(
    "customTime", DateTime.parseRfc3339("2020-01-01T00:00:00Z"));
Blob blob = storage.get(bucketName, objectName);
blob.toBuilder().setMetadata(newMetadata).build().update();
```

PR #413, opened 2020-07-01
feat: add support of customTime metadata

athakor:
Towards #395

### DIFF
--- a/google-cloud-storage/clirr-ignored-differences.xml
+++ b/google-cloud-storage/clirr-ignored-differences.xml
@@ -26,9 +26,4 @@
         <method>com.google.cloud.storage.BucketInfo$Builder setUpdateTime(java.lang.Long)</method>
         <differenceType>7013</differenceType>
     </difference>
-    <difference>
-        <className>com/google/cloud/storage/BlobInfo$Builder</className>
-        <method>com.google.cloud.storage.BlobInfo$Builder setCustomTime(java.lang.Long)</method>
-        <differenceType>7013</differenceType>
-    </difference>
 </differences>
--- a/google-cloud-storage/clirr-ignored-differences.xml
+++ b/google-cloud-storage/clirr-ignored-differences.xml
@@ -26,4 +26,9 @@
         <method>com.google.cloud.storage.BucketInfo$Builder setUpdateTime(java.lang.Long)</method>
         <differenceType>7013</differenceType>
     </difference>
+    <difference>
+        <className>com/google/cloud/storage/BlobInfo$Builder</className>
+        <method>com.google.cloud.storage.BlobInfo$Builder setCustomTime(java.lang.Long)</method>
+        <differenceType>7013</differenceType>
+    </difference>
 </differences>
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
@@ -84,6 +84,7 @@ public class BlobInfo implements Serializable {
   private final String etag;
   private final String md5;
   private final String crc32c;
+  private final Long customTime;
   private final String mediaLink;
   private final Map<String, String> metadata;
   private final Long metageneration;
@@ -260,6 +261,9 @@ public class BlobInfo implements Serializable {
      */
     public abstract Builder setCrc32c(String crc32c);
 
+    /* */
+    public abstract Builder setCustomTime(Long customTime);
+
     /**
      * Sets the CRC32C checksum of blob's data as described in <a
      * href="http://tools.ietf.org/html/rfc4960#appendix-B">RFC 4960, Appendix B;</a> from hex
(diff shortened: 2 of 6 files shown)


google-cloud-storage/clirr-ignored-differences.xml
```diff
@@ -21,4 +21,9 @@
         <method>com.google.cloud.storage.BucketInfo$Builder deleteLifecycleRules()</method>
         <differenceType>7013</differenceType>
     </difference>
+    <difference>
+        <className>com/google/cloud/storage/BlobInfo$Builder</className>
+        <method>com.google.cloud.storage.BlobInfo$Builder setCustomTime(java.lang.Long)</method>
+        <differenceType>7013</differenceType>
```
elharo, line 27: Are you prepared to bump the major version? If not, please don't do this. This check is here for a reason. We should not simply ad an exclude every time it says something is wrong.

athakor:
@jkwlui

athakor:
@frankyn @jkwlui friendly ping

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -260,6 +261,9 @@ static CustomerEncryption fromPb(StorageObject.CustomerEncryption customerEncryp
      */
     public abstract Builder setCrc32c(String crc32c);
 
+    /** Sets the custom time for an object. */
+    public abstract Builder setCustomTime(Long customTime);
```
elharo, line 265: Why is this abstract? If it's concrete it wouldn't be a breaking change.

frankyn: It's the development model we've used over the years. It was designed before my time and folks who worked on the library have since left the team.

We could update it to concrete method instead but we would start changing development flow. Maybe that's a good thing, but I'd prefer to prevent changing development flow until we have a new design that fits into that model.

elharo: The pot roast principle applies:

https://www.psychologytoday.com/us/blog/thinking-makes-it-so/201402/the-pot-roast-principle

frankyn: If it's implemented, it could be implemented to raise an exception of it not being implemented. Seems a little clunky, but would remove the need to create a breaking change.

Thanks for raising the simple questions @elharo, I haven't applied this train of thought to this problem, and I appreciate the feedback.

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -260,6 +261,9 @@ static CustomerEncryption fromPb(StorageObject.CustomerEncryption customerEncryp
      */
     public abstract Builder setCrc32c(String crc32c);
 
+    /** Sets the custom time for an object. */
```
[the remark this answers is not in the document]
athakor: done

athakor:
@frankyn PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
frankyn: @athakor please implement this method instead of using abstract to prevent the breaking change. Instead raise an exception of unimplemented if called directly.

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -260,6 +261,22 @@ static CustomerEncryption fromPb(StorageObject.CustomerEncryption customerEncryp
      */
     public abstract Builder setCrc32c(String crc32c);
 
+    /**
+     * Sets the custom time for an object. Once set it can't be unset and only changed to a custom
+     * datetime in the future. If the custom_time must be unset, you must either perform a rewrite
```
JesseLovelace, line 266: Don't use "custom_time" in this context, say "custom time" instead. Replace "If the custom_time must be unset," with "To unset the custom time,"

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -260,6 +261,22 @@ static CustomerEncryption fromPb(StorageObject.CustomerEncryption customerEncryp
...
+     * operation or upload the data again.
+     *
+     * <p>Example of setting the custom_time.
```
JesseLovelace, line 269: custom time instead of custom_time

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
athakor: done

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
athakor: done

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
athakor: done

athakor:
@frankyn @JesseLovelace All the comments have been addressed PTAL

PR #430, opened 2020-07-13
feat: add support of startOffset and endOffset

### LINKED ISSUE #424: storage: support startOffset and endOffset query param during object listing
Currently we don't have support to set the `startOffset `and `endOffset ` during the object listing. 

Similar case found in go as well see https://github.com/googleapis/google-cloud-go/issues/2556 also check the[ JSON API](https://cloud.google.com/storage/docs/json_api/v1/objects/list#http-request) for more info.

~/cc: @frankyn

athakor:
Fixes #424

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
@@ -999,17 +999,14 @@ public class ITStorageTest {
     Bucket bucket =
         storage.create(BucketInfo.newBuilder(bucketName).setVersioningEnabled(true).build());
     try {
-      String[] blobNames = {
-        "test-list-blobs-start-offset-blob1",
-        "test-list-blobs-start-offset-blob2",
-        "test-list-blobs-end-offset-blob3"
-      };
+      List<String> blobNames =
+          ImmutableList.of("startOffset_blob1", "startOffset_blob2", "blob3_endOffset");
       BlobInfo blob1 =
-          BlobInfo.newBuilder(bucket, blobNames[0]).setContentType(CONTENT_TYPE).build();
+          BlobInfo.newBuilder(bucket, blobNames.get(0)).setContentType(CONTENT_TYPE).build();
       BlobInfo blob2 =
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -993,6 +993,56 @@ public void testListBlobsVersioned() throws ExecutionException, InterruptedExcep
...
+      Blob remoteBlob2 = storage.create(blob2);
+      Blob remoteBlob3 = storage.create(blob3);
+      assertNotNull(remoteBlob1);
+      assertNotNull(remoteBlob2);
+      assertNotNull(remoteBlob3);
+      Page<Blob> page =
+          storage.list(
+              bucketName,
+              Storage.BlobListOption.startOffset("test-list-blobs-start-offset-blob"),
+              Storage.BlobListOption.endOffset("test-list-blobs-end-offset-blob3"));
```
frankyn, line 1024: Recommend starting from not the first object in the new objects.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -993,6 +993,56 @@ public void testListBlobsVersioned() throws ExecutionException, InterruptedExcep
...
+      // Listing blobs is eventually consistent, we loop until the list is of the expected size.
+      while (Iterators.size(page.iterateAll().iterator()) != 3) {
```
frankyn, line 1026: I can't tell if this is working because it's using all the objects created during the test.

athakor: @frankyn yes you are right, i have wrote new test instead of that. Now you can easily verify it.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
athakor: Replace this test with new one.

athakor:
@frankyn PTAL

PR #438, opened 2020-07-20
feat: add support of null to remove the CORS configuration from bucket 

### LINKED ISSUE #437: Unable to remove CORS configuration from bucket by setting null
```javascript
        String bucketName = RemoteStorageHelper.generateBucketName();
        try{
            Cors cors = Cors.newBuilder()
                    .setOrigins(ImmutableList.of(Cors.Origin.of("*")))
                    .setMethods(ImmutableList.of(HttpMethod.GET))
                    .setResponseHeaders(ImmutableList.of("Content-Type"))
                    .setMaxAgeSeconds(100)
                    .build();
            Bucket bucket =
                    storage.create(BucketInfo.newBuilder(bucketName).setCors(ImmutableList.of(cors)).build());
            System.out.println("Prior size : " + bucket.getCors().size());
            Bucket updatedBucket  = bucket.toBuilder().setCors(null).build().update();
            System.out.println("After updating operation size must be zero instead of : " + updatedBucket.getCors().size());

        }finally {
            RemoteStorageHelper.forceDelete(storage, bucketName, 5, TimeUnit.SECONDS);
        }
```
**Please check the unexpected response :**
```
Prior size : 1
After updating operation size must be zero instead of : 1

```

athakor:
Fixes #437

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -1240,7 +1240,7 @@ public class BucketInfo implements Serializable {
 
     @Override
     public Builder setCors(Iterable<Cors> cors) {
-      this.cors = cors != null ? ImmutableList.copyOf(cors) : null;
+      this.cors = cors != null ? ImmutableList.copyOf(cors) : ImmutableList.<Cors>of();
       return this;
     }
 
@@ -1509,7 +1509,7 @@ public class BucketInfo implements Serializable {
    *     (CORS)</a>
    */
   public List<Cors> getCors() {
-    return cors;
+    return cors != null ? cors : ImmutableList.<Cors>of();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1509,7 +1509,7 @@ public StorageClass getStorageClass() {
    *     (CORS)</a>
    */
   public List<Cors> getCors() {
-    return cors;
+    return cors != null ? cors : ImmutableList.<Cors>of();
```
frankyn, line 1512: Please leave this field as is and revert change. There are potentially 4 cases which are considered here and we don't want to mess them up.

1. Cors are set and field selector is selected then returns not-null.
2. Cors are set but field selector isn't selected then returns null.
3. Cors are not set and field selector is selected then returns null.
4. Cors are not set and field selector isn't selected then returns null.

athakor: Done

athakor:
@frankyn PTAL

Issue #410, opened 2020-06-30
Classes defined in PostPolicyV4 are not covered by unit tests


dmitry-fa:
Public classes defined in PostPolicyV4 are not properly covered by tests. There is a group of conformance tests to verify the functionality but this is not enough for public API.

Unit tests should be developed to check that all properties appear correctly in the corresponding RPC calls.

PR #442, opened 2020-07-29
fix: PostPolicyV4 classes could be improved

### LINKED ISSUE #440: PostPolicyV4 classes could be improved
Classes defined in PostPolictyV4.java are not as perfect as they can be.
Enhancements and bugs are listed below:

class `PostPolicyV4`
- the class description is stingy, need to put more words and point to usage example
- public methods miss Javadoc
- no verification for given URL string is a valid URL
- BUG:  `getFields()` returns the map as is, so one can modify PostPolicyV4 objects after creation
(suggested fix:  make `fields` immutable)

class `PostPolicyV4.PostFieldsV4`
- BUG: public method `getFieldsMap()` returns map that allows to alter PostFieldsV4 instances without Builder (no control)
  (suggested fix: deprecate method getFieldsMap(), make the map immutable, define getters)

class `PostPolicyV4.PostConditionsV4`
- BUG `getConditions()` returns mutable Set
- BUG `getConditions()` exposes non public class `PostPolicyV4.ConditionV4`
  (suggested fix: make `PostPolicyV4.ConditionV4` public, make `conditions` immutable)

dmitry-fa:
Fixes #440

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
@@ -180,7 +180,7 @@ public final class PostPolicyV4 {
       }
 
       /**
-       * @deprecated Invocation of this method has no effect, because All valid HTML form fields
+       * @deprecated Invocation of this method has no effect, because all valid HTML form fields
        *     except Content-Length can use exact matching. Use {@link
        *     PostPolicyV4.PostConditionsV4.Builder#addContentLengthRangeCondition(int, int)} to
        *     specify a range for the content-length.
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
@@ -84,7 +84,7 @@ public final class PostPolicyV4 {
    * A helper class to define fields to be specified in a V4 POST request. Instance of this class
    * helps to construct {@code PostPolicyV4} objects. Used in: {@link
    * Storage#generateSignedPostPolicyV4(BlobInfo, long, TimeUnit, PostPolicyV4.PostFieldsV4,
-   * PostPolicyV4.PostConditionsV4, Storage.PostPolicyV4Option...)}
+   * PostPolicyV4.PostConditionsV4, Storage.PostPolicyV4Option...)}.
    *
    * @see <a href="https://cloud.google.com/storage/docs/xml-api/post-object#form_fields">POST
    *     Object Form fields</a>
@@ -131,9 +131,9 @@ public final class PostPolicyV4 {
     /**
      * Constructs {@code PostPolicyV4.PostFieldsV4} object of the given field map.
      *
-     * @param fields a map
+     * @param fields a map the HTML form fields
      * @return constructed object
-     * @throws IllegalArgumentException if unsupported field is specified.
+     * @throws IllegalArgumentException if an unsupported field is specified
      */
     public static PostFieldsV4 of(Map<String, String> fields) {
       return new PostFieldsV4(fields);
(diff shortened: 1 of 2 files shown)


dmitry-fa:
@elharo 
The issue for new tests: https://github.com/googleapis/java-storage/issues/410

The model code was integrated without proper coverage by tests. For the time being it's covered by 10 conformance tests. This just adds javadoc and fixes minor bugs found.

Before developing new tests I'd like to get feedback on the proposed changes.

google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
```diff
@@ -18,56 +18,123 @@
...
  */
 public final class PostPolicyV4 {
   private final String url;
   private final Map<String, String> fields;
 
   private PostPolicyV4(String url, Map<String, String> fields) {
+    try {
+      new URL(url);
+    } catch (MalformedURLException e) {
+      throw new IllegalArgumentException(e);
```
elharo, line 53: This needs a test

google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
```diff
@@ -18,56 +18,123 @@
...
+    }
+    PostFieldsV4.validateFields(fields);
+
     this.url = url;
-    this.fields = fields;
+    this.fields = Collections.unmodifiableMap(fields);
```
elharo, line 58: This needs a test

google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
```diff
@@ -18,56 +18,123 @@
...
+            "x-goog-algorithm",
+            "x-goog-credential",
+            "x-goog-date",
+            "x-goog-signature");
+
+    private static void validateFields(Map<String, String> fields) {
+      for (String key : fields.keySet()) {
+        if (!VALID_FIELDS.contains(key.toLowerCase())
+            && !key.startsWith(Builder.CUSTOM_FIELD_PREFIX)) {
+          throw new IllegalArgumentException("Invalid key: " + key);
```
elharo, line 117: This needs a test.

google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
dmitry-fa: The first portion of planned tests which cover the changed code has been added.

dmitry-fa: @elharo all the planned tests are pushed

dmitry-fa:
@elharo 
> Comment changes don't need tests but this PR changes behavior. Every behavioral change should be proceeded by a test that fails and only passes after the model code is added.

All the tests have been developed. Will much appreciate if you take a look.

google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
```diff
@@ -18,56 +18,123 @@
...
+ * PostPolicyV4.PostFieldsV4, PostPolicyV4.PostConditionsV4, Storage.PostPolicyV4Option...)} for
+ * example of usage.
  */
 public final class PostPolicyV4 {
   private final String url;
   private final Map<String, String> fields;
 
   private PostPolicyV4(String url, Map<String, String> fields) {
+    try {
+      new URL(url);
```
elharo, line 51: What URLs do you really want to allow and not allow here? java.net.URL is based on java protocol handlers, and is likely to be either too lenient or too strict depending on what you need.

dmitry-fa: I want to allow strings that could be used to construct an URL object and reject strings like 'google.com' given by mistake instead of 'https://google.com'. Catching 'ftp://google.com' is not the goal.

PR #449, opened 2020-08-04
feat: expose updateTime field of the bucket

### LINKED ISSUE #448: Expose updateTime field of the bucket's
Expose `updateTime `field of the [bucket's](https://github.com/googleapis/google-api-java-client-services/blob/b39bb16a44aa99647d56b39054817ee6493ef8f5/clients/google-api-services-storage/v1/1.30.1/com/google/api/services/storage/model/Bucket.java#L226) it's look like missing here, [see](https://github.com/googleapis/java-storage/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java#L75-L101).

~//cc: @frankyn

athakor:
Fixes #448

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
@@ -56,7 +56,7 @@ public class BucketInfoTest {
   private static final User OWNER = new User("user@gmail.com");
   private static final String SELF_LINK = "http://storage/b/n";
   private static final Long CREATE_TIME = System.currentTimeMillis();
-  private static final Long UPDATE_TIME = CREATE_TIME - 1L;
+  private static final Long UPDATE_TIME = CREATE_TIME;
   private static final List<Cors> CORS = Collections.singletonList(Cors.newBuilder().build());
   private static final List<Acl> DEFAULT_ACL =
       Collections.singletonList(Acl.of(User.ofAllAuthenticatedUsers(), Role.WRITER));
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
@@ -17,7 +17,9 @@
 package com.google.cloud.storage;
 
 import static com.google.cloud.storage.Acl.Project.ProjectRole.VIEWERS;
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 import com.google.api.services.storage.model.Bucket;
 import com.google.api.services.storage.model.Bucket.Lifecycle.Rule;
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
@@ -17,9 +17,7 @@
 package com.google.cloud.storage;
 
 import static com.google.cloud.storage.Acl.Project.ProjectRole.VIEWERS;
-import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -56,6 +56,7 @@
   private static final User OWNER = new User("user@gmail.com");
   private static final String SELF_LINK = "http://storage/b/n";
   private static final Long CREATE_TIME = System.currentTimeMillis();
+  private static final Long UPDATE_TIME = CREATE_TIME - 1L;
```
dmitry-fa, line 59: Why `CREATE_TIME - 1L` not just `CREATE_TIME`?

athakor: we could do that as well but for the different value i would prefer this and one more thing these is only for testing purpose.

dmitry-fa: If the update time of an object differs from creating time, it means that the object has been changed since creating.

athakor: done. @dmitry-fa PTAL

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3535,4 +3535,22 @@ public void testRemoveBucketCORS() throws ExecutionException, InterruptedExcepti
...
+      assertThat(bucket).isNotNull();
+      assertThat(bucket.getCreateTime()).isNotNull();
+
+      Bucket updatedBucket = bucket.toBuilder().setVersioningEnabled(false).build().update();
+      assertThat(updatedBucket.versioningEnabled()).isFalse();
+      assertThat(updatedBucket.getUpdateTime()).isNotNull();
+    } finally {
+      RemoteStorageHelper.forceDelete(storage, bucketName, 5, TimeUnit.SECONDS);
+    }
+  }
```
dmitry-fa, line 3555: To make sure that the new feature is working and useful I would check the following by this test: 

1) after bucket creation:
createTime != null 
updateTime == createTime

2) after bucket updated:
createTime has not changed
updateTime is set to time after createTime

athakor: done

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3535,4 +3535,27 @@ public void testRemoveBucketCORS() throws ExecutionException, InterruptedExcepti
...
+      assertThat(bucket.versioningEnabled()).isTrue();
+      assertThat(bucket.getCreateTime()).isNotNull();
+      assertThat(bucket.getUpdateTime()).isEqualTo(bucket.getCreateTime());
+
+      Bucket updatedBucket = bucket.toBuilder().setVersioningEnabled(false).build().update();
+      assertThat(updatedBucket.versioningEnabled()).isFalse();
+      assertThat(updatedBucket.getUpdateTime()).isNotNull();
+      assertThat(updatedBucket.getCreateTime()).isEqualTo(bucket.getCreateTime());
+      assertThat(updatedBucket.getUpdateTime()).isNotEqualTo(bucket.getCreateTime());
+      assertThat(updatedBucket.getUpdateTime()).isNotEqualTo(bucket.getUpdateTime());
```
dmitry-fa, line 3556: one of two last assertions is redundant (because you checked: bucket.getCreateTime() == bucket.getUpdateTime())
I would replace them with:
assertThat(updatedBucket.getUpdateTime() > updatedBucket.getCreateTime() );

athakor: done

PR #454, opened 2020-08-05
test: framework for testing rpc

### LINKED ISSUE #450: Need a test framework for Storage RPC tests
Google requires to cover all new code and all changes in the code by tests. The majority of the Storage code is covered by unit tests, but not `com.google.cloud.storage.spi.v1.HttpStorageRpc`. Testing of  `HttpStorageRpc` is a bit tricky, so having a framework allowing us to write laconic tests for RPC would help to meet the Google requirements.

// cc @frankyn

dmitry-fa:
The suggested approach allows us to add new Storage RPC calls covered by unit tests.
The existing RPC calls could be also covered by unit tests.

Fixes #450

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
@@ -102,16 +102,12 @@ public class HttpStorageRpcTest {
     return holder;
   }
 
-  private static String getUnzippedContent(LowLevelHttpRequest request) {
-    try {
-      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-      request.getStreamingContent().writeTo(outputStream);
-      GZIPInputStream gzipInputStream =
-          new GZIPInputStream(new ByteArrayInputStream(outputStream.toByteArray()));
-      return new String(ByteStreams.toByteArray(gzipInputStream), UTF_8);
-    } catch (IOException e) {
-      throw new AssertionError("Unexpected exception", e);
-    }
+  private static String getUnzippedContent(LowLevelHttpRequest request) throws IOException {
+    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+    request.getStreamingContent().writeTo(outputStream);
+    GZIPInputStream gzipInputStream =
+        new GZIPInputStream(new ByteArrayInputStream(outputStream.toByteArray()));
+    return new String(ByteStreams.toByteArray(gzipInputStream), UTF_8);
   }
 
   @Before
@@ -138,7 +134,7 @@ public class HttpStorageRpcTest {
   }
 
   @Test
-  public void testCreateBucket() {
+  public void testCreateBucket() throws IOException {
     byte[] content = "{\"name\":\"yyy\"}".getBytes(UTF_8);
     String url = BASE_URL + '?' + URL_PROJECT;
     RpcRequestHolder holder = mockResponse("POST", url, new TestResponse(200, content));
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
@@ -277,7 +277,6 @@ public class HttpStorageRpcTest {
   public void testCreateBucket() throws Exception {
     byte[] content = "{\"name\":\"yyy\"}".getBytes(UTF_8);
     String url = BASE_URL + '?' + URL_PROJECT;
-    mockResponse("POST", url, new TestResponse(401, null));
     RpcRequestHolder holder = mockResponse("POST", url, new TestResponse(200, content));
 
     Bucket created = rpc.create(new Bucket().setName("xxx"), new HashMap());
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
@@ -34,13 +34,14 @@ import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.storage.StorageException;
 import com.google.cloud.storage.StorageOptions;
 import com.google.cloud.storage.spi.StorageRpcFactory;
+import com.google.common.io.ByteStreams;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.Map;
 import java.util.Queue;
 import java.util.zip.GZIPInputStream;
@@ -57,7 +58,7 @@ public class HttpStorageRpcTest {
 
   // Objects for a test case: one object per one RPC mock.
   // A test creates as many MockData objects as many RPC call it issues.
-  private static final Queue<MockData> TEST_MOCK_DATA = new LinkedList<>();
+  private static final Queue<MockData> TEST_MOCK_DATA = new ArrayDeque<>();
 
   private static final String BASE_URL = "https://storage.googleapis.com/storage/v1/b";
   private static final String URL_PROJECT = "project=projectId&projection=full";
@@ -252,11 +253,9 @@ public class HttpStorageRpcTest {
     try {
       ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
       request.getStreamingContent().writeTo(outputStream);
-      GZIPInputStream is =
+      GZIPInputStream gzipInputStream =
           new GZIPInputStream(new ByteArrayInputStream(outputStream.toByteArray()));
-      byte[] buffer = new byte[1000000];
-      int size = is.read(buffer);
-      return new String(buffer, 0, size, UTF_8);
+      return new String(ByteStreams.toByteArray(gzipInputStream), UTF_8);
     } catch (IOException e) {
       throw new Error(e);
     }
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
@@ -23,12 +23,10 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
 
-import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.LowLevelHttpRequest;
 import com.google.api.client.http.LowLevelHttpResponse;
 import com.google.api.core.ApiClock;
 import com.google.api.services.storage.model.Bucket;
-import com.google.auth.http.HttpTransportFactory;
 import com.google.cloud.ServiceOptions;
 import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.storage.StorageException;
@@ -38,7 +36,6 @@ import com.google.common.io.ByteStreams;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,314 @@
...
+import org.mockito.stubbing.Answer;
+
+public class HttpStorageRpcTest {
+
+  // object under test, created before each test case
+  private HttpStorageRpc rpc;
+
+  // Objects for a test case: one object per one RPC mock.
+  // A test creates as many MockData objects as many RPC call it issues.
+  private static final Queue<MockData> TEST_MOCK_DATA = new LinkedList<>();
```
elharo, line 60: ArrayDeque instead of LinkedList

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,314 @@
...
+    return holder;
+  }
+
+  private static String getUnzippedContent(LowLevelHttpRequest request) {
+    try {
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      request.getStreamingContent().writeTo(outputStream);
+      GZIPInputStream is =
+          new GZIPInputStream(new ByteArrayInputStream(outputStream.toByteArray()));
+      byte[] buffer = new byte[1000000];
```
elharo, line 257: This isn't safe or correct. Use ByteStreams.toByteArray(is)

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
dmitry-fa: done

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
dmitry-fa: Yes, agree. Forgot to fix this before PR. Done.

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,313 @@
...
+        @Override
+        public long millisTime() {
+          return 42_000L;
+        }
+      };
+
+  private static final StorageRpcFactory RPC_FACTORY_MOCK =
+      mock(StorageRpcFactory.class, UNEXPECTED_CALL_ANSWER);
+
+  private static class TestRequest extends LowLevelHttpRequest {
```
elharo, line 94: make this an outer class for clarity

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,313 @@
...
+      }
+    }
+
+    @Override
+    public LowLevelHttpResponse execute() throws IOException {
+      return response;
+    }
+  };
+
+  private static class TestResponse extends LowLevelHttpResponse {
```
elharo, line 115: make this an outer class for clarity

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,313 @@
...
+import org.mockito.stubbing.Answer;
+
+public class HttpStorageRpcTest {
+
+  // object under test, created before each test case
+  private HttpStorageRpc rpc;
+
+  // Objects for a test case: one object per one RPC mock.
+  // A test creates as many MockData objects as many RPC call it issues.
+  private static final Queue<MockData> TEST_MOCK_DATA = new ArrayDeque<>();
```
elharo, line 61: Why is this static? That's likely to cause coupling between tests. This can probably be an instance field.

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,313 @@
...
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+public class HttpStorageRpcTest {
+
+  // object under test, created before each test case
+  private HttpStorageRpc rpc;
+
+  // Objects for a test case: one object per one RPC mock.
+  // A test creates as many MockData objects as many RPC call it issues.
```
elharo, line 60: as it issues RPC calls

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,313 @@
...
+    void setRequest(TestRequest request) {
+      this.request = request;
+    }
+
+    TestRequest getRequest() {
+      return request;
+    }
+  }
+
+  static class MockData {
```
elharo, line 222: outer class

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,313 @@
...
+      return new String(ByteStreams.toByteArray(gzipInputStream), UTF_8);
+    } catch (IOException e) {
+      throw new Error(e);
+    }
+  }
+
+  @Before
+  public void setUp() throws Exception {
+    rpc = new HttpStorageRpc(STORAGE_OPTIONS);
+    TEST_MOCK_DATA.clear();
```
elharo, line 267: yes, this really should be an instance variable so you don't have to worry about clearing it

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -51,13 +51,122 @@
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
 
+class TestRequest extends LowLevelHttpRequest {
```
elharo, line 54: I meant these should be in new files.

dmitry-fa: Are you sure? These classes unlikely will be used by anyone else but HttpStorageRpcTest. 
I can move them to separate files if you insist, but IMHO it will only make understanding harder. 
If it is just a matter of taste, I would prefer to see them as inner classes.

elharo: Not sure if there's an explicit style rule but Google almost never uses non-public outer classes in the same .java file. 

In general inner classes encourage violation of data encapsulation and the single responsibility principle. 

And perhaps most importantly here, the sheer size of a file becomes a problem. I found this harder to read as one large class than as several smaller classes.

dmitry-fa: Thanks for the explanation. I'll move the classes with code to new files and, if you don't mind, keep the trivial classes  (MockData and RpcRequestHolder) as inner.

dmitry-fa: All inner classes moved to new files

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
dmitry-fa: refactored to be an instance variable

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private static String getUnzippedContent(LowLevelHttpRequest request) {
+    try {
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      request.getStreamingContent().writeTo(outputStream);
+      GZIPInputStream gzipInputStream =
+          new GZIPInputStream(new ByteArrayInputStream(outputStream.toByteArray()));
+      return new String(ByteStreams.toByteArray(gzipInputStream), UTF_8);
+    } catch (IOException e) {
+      throw new Error(e);
```
elharo, line 113: RuntimeException, or better yet just bubble the IOException

dmitry-fa: I selected Error because `An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch. Most such errors are abnormal conditions.`
IOException thrown from this try-catch block will signal about problems in the core Java libraries, not in the Google libraries under test.

dmitry-fa: I replaced `Error` with `AssertionError` which `JUnit.fail()` throws. I can't replace with `fail()` because the method returns a value.

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private static String getUnzippedContent(LowLevelHttpRequest request) {
+    try {
+      ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+      request.getStreamingContent().writeTo(outputStream);
+      GZIPInputStream gzipInputStream =
+          new GZIPInputStream(new ByteArrayInputStream(outputStream.toByteArray()));
+      return new String(ByteStreams.toByteArray(gzipInputStream), UTF_8);
+    } catch (IOException e) {
+      throw new AssertionError("Unexpected exception", e);
```
elharo, line 113: just bubble the IOException, no need to catch it at all

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,174 @@
...
+    RpcRequestHolder holder = new RpcRequestHolder();
+    testMockData.add(new MockData(method, url, response, holder));
+    return holder;
+  }
+
+  private static String getUnzippedContent(LowLevelHttpRequest request) throws IOException {
+    ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+    request.getStreamingContent().writeTo(outputStream);
+    GZIPInputStream gzipInputStream =
+        new GZIPInputStream(new ByteArrayInputStream(outputStream.toByteArray()));
```
elharo, line 109: since you already have a byte array use an inflater, not a stream

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.mockito.stubbing.Answer;
+
+public class HttpStorageRpcTest {
+
+  // object under test, created before each test case
+  private HttpStorageRpc rpc;
+
+  // Objects for a test case: one object per one RPC mock.
+  // A test creates as many MockData objects as it issues RPC calls.
+  private Queue<MockData> testMockData;
```
elharo, line 58: can initialize the queue here

dmitry-fa: are you sure it's necessary? This queue is re-initialized before each test case.

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
dmitry-fa: Would you mind if I kept using GZIPInputStream? java.util.zip.Inflater introduces new DataFormatException the test should handle. The inflater requires a byte buffer of a certain size, but the size is unknown. And I think, I can't use the java.util.zip.Inflater from the box to decompress GZIP content: 
```
 To fully decode the gzip format with this class, you will need to process the gzip header as described in RFC 1952, use inflater to decompress the raw deflate data, compute the crc32 of the uncompressed data using that class, and then verify the crc and length (modulo 2^32) in the gzip trailer, again as specified in the RFC.
```

Or do you mean another inflater?

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
elharo: not necessary, just a little cleaner. When possible, it's good habit to initialize and declare in the same line to avoid unexpected failure initialize or reuse of variables

google-cloud-storage/src/test/java/com/google/cloud/storage/spi/v1/HttpStorageRpcTest.java
elharo: Hmm, it's weird to copy this into a  byte array first. You should be able to go straight from the LowLevelHttpRequest to the decompressor, but LowLevelHttpRequest doesn't offer an input stream top read from, which is really strange. Maybe use a piped stream but that would need an extra thread.

elharo (commented):
LGTM

frankyn:
Closing for now, will consider in the future at a later time.

PR #456, opened 2020-08-06
feat: expose timeStorageClassUpdated property of blob's

### LINKED ISSUE #455: Expose timeStorageClassUpdated property of the blob's
Add support of `timeStorageClassUpdated ` property of the [blob's](https://github.com/googleapis/google-api-java-client-services/blob/b39bb16a44aa99647d56b39054817ee6493ef8f5/clients/google-api-services-storage/v1/1.30.1/com/google/api/services/storage/model/StorageObject.java#L273). The purpose of exposing these field's is to get the time at which the object's storage class was last changed.

see the api doc : https://cloud.google.com/storage/docs/json_api/v1/objects#resource

~//cc: @frankyn

athakor:
Fixes #455

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
@@ -298,7 +298,15 @@ public class BlobInfo implements Serializable {
     /** Sets the blob's storage class. */
     public abstract Builder setStorageClass(StorageClass storageClass);
 
-    abstract Builder setTimeStorageClassUpdated(Long timeStorageClassUpdated);
+    /**
+     * Sets the modification time of an object's storage class. Once set it can't be unset directly,
+     * the only way is to rewrite the object with the desired storage class.
+     */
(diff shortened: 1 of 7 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -276,6 +277,8 @@ static CustomerEncryption fromPb(StorageObject.CustomerEncryption customerEncryp
     /** Sets the blob's storage class. */
     public abstract Builder setStorageClass(StorageClass storageClass);
 
+    abstract Builder setTimeStorageClassUpdated(Long timeStorageClassUpdated);
```
frankyn, line 280: Follow the same pattern in custom time to raise an exception and implement this method to prevent breaking changes.

athakor: done

athakor:
@frankyn PTAL

frankyn (approved):

frankyn (approved):

PR #460, opened 2020-08-10
feat: Add abort() to writer

### LINKED ISSUE #202: Add abort() to writer
Feature request filed in google-cloud-java: https://github.com/googleapis/google-cloud-java/issues/1232

dmitry-fa:
Fixes #202
The current implementation is not elegant, because one has to call `storage.abort(writer)` to cancel resumable upload. 
Once https://github.com/googleapis/java-core/issues/245 is implemented (I hope it will be implemented sooner or later), we need just to rename private `cancelUpload()` method of `BlobWriteChannel` to `public void abort()`.

### DIFF
--- a/google-cloud-storage/clirr-ignored-differences.xml
+++ b/google-cloud-storage/clirr-ignored-differences.xml
@@ -22,6 +22,16 @@
         <differenceType>7013</differenceType>
     </difference>
     <difference>
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/storage/Storage</className>
+        <method>void abort(com.google.cloud.WriteChannel)</method>
+    </difference>
+    <difference>
+        <differenceType>7012</differenceType>
+        <className>com/google/cloud/storage/spi/v1/StorageRpc</className>
+        <method>void abort(java.lang.String)</method>
+    </difference>
+    <difference>
         <className>com/google/cloud/storage/BucketInfo$Builder</className>
         <method>com.google.cloud.storage.BucketInfo$Builder setUpdateTime(java.lang.Long)</method>
         <differenceType>7013</differenceType>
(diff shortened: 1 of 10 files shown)


dmitry-fa:
New tests covering RPC code will be added after https://github.com/googleapis/java-storage/pull/454 is merged.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -696,6 +696,15 @@ public BlobWriteChannel writer(URL signedURL) {
     return new BlobWriteChannel(getOptions(), signedURL);
   }
 
+  @Override
+  public void abort(WriteChannel channel) {
+    if (!(channel instanceof BlobWriteChannel)) {
```
elharo, line 701: why not make the method argument a BlobWriteChannel then?

google-cloud-storage/clirr-ignored-differences.xml
```diff
@@ -21,6 +21,16 @@
         <method>com.google.cloud.storage.BucketInfo$Builder deleteLifecycleRules()</method>
         <differenceType>7013</differenceType>
     </difference>
+    <difference>
```
elharo, line 24: Yet again, we're getting stuck because of excessive use of interfaces. Either there are multiple implementations of Storage and we shouldn't make this change or we should bite the bullet and combine Storage and StorageImpl once and for all. Ditto StorageRpc and HttpStorageRpc.

elharo (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
dmitry-fa: @elharo BlobWriteChannel is a package-private class. If it were public we could invoke channel.abort() directly.

google-cloud-storage/clirr-ignored-differences.xml
dmitry-fa: @frankyn what do you think about combining Storage and StorageImpl?

PR #484, opened 2020-08-28
fix: add missing Bucket and Blob FieldSelector

athakor:


google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -151,7 +153,8 @@ public String getSelector() {
     EVENT_BASED_HOLD("eventBasedHold"),
     TEMPORARY_HOLD("temporaryHold"),
     RETENTION_EXPIRATION_TIME("retentionExpirationTime"),
-    UPDATED("updated");
+    UPDATED("updated"),
+    CUSTOM_TIME("customTime");
```
frankyn, line 157: Definitely missed, thanks @athakor.

PR #604, opened 2020-11-09
fix: retry using remote offset

### LINKED ISSUE #603: Retry Based on Remote Offset
If your request is terminated prior to receiving a response from the server or if you receive an HTTP 503 response from the server, you can query the current status of the upload by issuing an empty PUT request on the unique upload URI.

The storage client should poll the server to determine which bytes it has received.

```java
// Variables:
// chunk = getBuffer()
// localNextByteOffset == getPosition()
// chunkSize = getChunkSize()
//
// Case 1: localNextByteOffset == 0 && remoteNextByteOffset == 0:
// we are retrying from first chunk start from 0 offset.
//
// Case 2: localNextByteOffset == remoteNextByteOffset:
// Special case of Case 1 when a chunk is retried.
//
// Case 3: localNextByteOffset < remoteNextByteOffset
//             && driftOffset < chunkSize:
// Upload progressed and localNextByteOffset is not in-sync with
// remoteNextByteOffset and driftOffset is less than chunkSize.
// driftOffset must be less than chunkSize for it to retry using
// chunk maintained in memory.
// Find the driftOffset by subtracting localNextByteOffset from
// remoteNextByteOffset.
// Use driftOffset to determine where to restart from using the chunk in
// memory.
//
// Case 4: localNextByteOffset < remoteNextByteOffset
//            && driftOffset == chunkSize:
// Special case of Case 3.
// If chunkSize is equal to driftOffset then remoteNextByteOffset has moved on
// to the next chunk.
//
// Case 5: localNextByteOffset < remoteNextByteOffset
//            && driftOffset > chunkSize:
// Throw exception as remoteNextByteOffset has drifted beyond the retriable
// chunk maintained in memory. This is not possible unless there's multiple
// clients uploading to the same resumable upload session.
//
// Case 6: localNextByteOffset > remoteNextByteOffset:
// For completeness, this case is not possible because it would require retrying
// a 400 status code which is not allowed.
```

frankyn:
Note: @JesseLovelace will be driving this PR to completion as I'll be OOO over the next two weeks. Thanks Jesse!
- [ ] Code coverage does not decrease

Fixes #603 ☕️

### DIFF
--- a/.github/readme/synth.metadata/synth.metadata
+++ b/.github/readme/synth.metadata/synth.metadata
@@ -4,14 +4,14 @@
       "git": {
         "name": ".",
         "remote": "https://github.com/googleapis/java-storage.git",
-        "sha": "1f5f9e07d36eb79b524bc7c9a8e514e4260be17a"
+        "sha": "e1c221d06ef8ed55384b498573827c3cca863591"
       }
     },
     {
       "git": {
         "name": "synthtool",
         "remote": "https://github.com/googleapis/synthtool.git",
-        "sha": "f68649c5f26bcff6817c6d21e90dac0fc71fef8e"
+        "sha": "fd3584b01cedd6c9f79d08be4e2365085f955aa5"
       }
     }
   ]
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ If you are using Maven with [BOM][libraries-bom], add this to your pom.xml file
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>libraries-bom</artifactId>
-      <version>13.2.0</version>
+      <version>13.3.0</version>
       <type>pom</type>
       <scope>import</scope>
     </dependency>
--- a/google-cloud-storage/clirr-ignored-differences.xml
+++ b/google-cloud-storage/clirr-ignored-differences.xml
@@ -26,4 +26,9 @@
         <method>com.google.cloud.storage.BucketInfo$Builder setUpdateTime(java.lang.Long)</method>
         <differenceType>7013</differenceType>
     </difference>
+    <difference>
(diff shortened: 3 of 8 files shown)


JesseLovelace:
The coverage miss seems to be due to getCurrentUploadOffset in HttpStorageRpc not being tested. This is part of way bigger issue, that entire file doesn't seem to have any coverage. I'm going to go ahead and merge this as is and it should be a follow up issue to get coverage for HttpStorageRpc

JesseLovelace (approved):

wildraid:
I suspect, this fix did go so well.

I am now getting following exceptions sometimes:

```
com.google.cloud.storage.StorageException: java.lang.NullPointerException
Stack trace:
com.google.cloud.storage.StorageException.translateAndThrow(StorageException.java:81)
com.google.cloud.storage.BlobWriteChannel.flushBuffer(BlobWriteChannel.java:180)
com.google.cloud.BaseWriteChannel.flush(BaseWriteChannel.java:112)
com.google.cloud.BaseWriteChannel.write(BaseWriteChannel.java:139)
java.base/java.nio.channels.Channels.writeFullyImpl(Channels.java:74)
java.base/java.nio.channels.Channels.writeFully(Channels.java:97)
java.base/java.nio.channels.Channels$1.write(Channels.java:172)
```

I suspect, we should have at least a better error message.

I use version 1.113.4. I'll try to go back to 1.113.3 and see how it goes.

Thank you.

frankyn:
Thanks for letting  us know, @wildraid!

Could you share a snippet of code that you used which caused this error?

frankyn:
@wildraid released a new version with a fix for the issue you raised. Thanks for calling it out!

wildraid:
@frankyn , great news! Thanks for your help!

PR #636, opened 2020-11-26
feat: add support of public access prevention

### LINKED ISSUE #616: Add support for PublicAccessPrevention
We need to add support for a new field `publicAccessPrevention` in the bucket IAM configuration for storage. Full details are available in [this doc](https://docs.google.com/document/d/1o_ElQg4c8enEx8OYa3irrUYFKADmHFbRn7dC9g287sI/). This should include ability to create buckets with this field set, get the field, and set the field. We'd also like an integration test that verifies that the feature works as expected.

Please wait for approval from myself or @tritone before merging or releasing this change so we can coordinate around release dates.

Based on looking at the code, I _think_ the field should go here:
https://github.com/googleapis/java-storage/blob/ac65e5b0bd324d5726504bb3405c758675a56ddc/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java#L110

athakor:
Fixes #616


athakor:
@shaffeeullah @tritone

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -112,6 +112,7 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
 
     private Boolean isUniformBucketLevelAccessEnabled;
     private Long uniformBucketLevelAccessLockedTime;
+    private String publicAccessPrevention;
```
frankyn, line 115: @shaffeeullah @tritone are other languages introducing Enums for this value or just keeping it as a string?

google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -72,10 +72,12 @@
               LifecycleAction.newDeleteAction(),
               LifecycleCondition.newBuilder().setAge(5).build()));
   private static final String INDEX_PAGE = "index.html";
+  private static final String PUBLIC_ACCESS_PREVENTION_ENFORCED = "enforced";
```
frankyn, line 75: @athakor can you add a system test even if it doesn't pass right now?

frankyn (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
tritone: In python and node we used strings with provided constants for "enforced" and "unspecified". So at the very least, we should provide these constants in the library surface (this is not done currently looks like).

If enums are more typical in this library (vs. node or python) we should feel free to use an enum here as well.

google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
tritone: Yes, we'll definitely need a system test here.

tritone: Most functionality should work currently for allowlisted projects.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
frankyn: Thanks @tritone. It would be helpful @athakor to introduce enums to reduce need for string literals.

google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
athakor: @shaffeeullah @tritone I provided our  java CI project IDs for the allow list in a comment on the design doc, thank you.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
athakor: done

elharo:
```
Dec 03, 2020 7:29:41 AM com.google.cloud.tools.dependencies.linkagemonitor.LinkageMonitor run
SEVERE: Newly introduced problems:
(com.google.apis:google-api-services-storage:v1-rev20200927-1.30.10) com.google.api.services.storage.model.Bucket$IamConfiguration's method setPublicAccessPrevention(String) is not found
  referenced from com.google.cloud.storage.BucketInfo (com.google.cloud:google-cloud-storage:1.113.5-SNAPSHOT)
(com.google.apis:google-api-services-storage:v1-rev20200927-1.30.10) com.google.api.services.storage.model.Bucket$IamConfiguration's method getPublicAccessPrevention() is not found
  referenced from com.google.cloud.storage.BucketInfo (com.google.cloud:google-cloud-storage:1.113.5-SNAPSHOT)

com.google.apis:google-api-services-storage:v1-rev20200927-1.30.10 is at:
  com.google.cloud:google-cloud-nio:0.122.1 (compile) / com.google.apis:google-api-services-storage:v1-rev20200927-1.30.10 (compile)
  and 2 other dependency paths.
com.google.cloud:google-cloud-storage:1.113.5-SNAPSHOT is at:
  com.google.cloud:google-cloud-storage:1.113.5-SNAPSHOT (compile)
```

frankyn:
For the linker issues, we need to cut a new release of storage with the updated Apiary client (https://github.com/googleapis/java-storage/pull/613)

tritone:
> For the linker issues, we need to cut a new release of storage with the updated Apiary client (#613) and a new release of libraries-bom as well.

Looks like this has been completed. @athakor can you merge in master?

athakor:
> > For the linker issues, we need to cut a new release of storage with the updated Apiary client (#613) and a new release of libraries-bom as well.
> 
> Looks like this has been completed. @athakor can you merge in master?

done

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -235,6 +269,21 @@ Builder setUniformBucketLevelAccessLockedTime(Long uniformBucketLevelAccessLocke
...
+      /**
+       * Sets the bucket's Public Access Prevention configuration. Currently supported options are
+       * 'unspecified' (default) or 'enforced'.
+       *
+       * @see <a
+       *     href="https://cloud.google.com/storage/docs/public-access-prevention">public-access-prevention</a>
+       */
+      public Builder setPublicAccessPrevention(PublicAccessPrevention publicAccessPrevention) {
+        this.publicAccessPrevention =
+            publicAccessPrevention == null
```
frankyn, line 281: It's possible for the value to be `null` if not selected in a GET request.

frankyn (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
athakor: > It's possible for the value to be `null` if not selected in a GET request.

yes, done. PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -176,6 +202,8 @@ public Long getUniformBucketLevelAccessLockedTime() {
               : new DateTime(uniformBucketLevelAccessLockedTime));
 
       iamConfiguration.setUniformBucketLevelAccess(uniformBucketLevelAccess);
+      iamConfiguration.setPublicAccessPrevention(
+          publicAccessPrevention == null ? null : publicAccessPrevention.getValue());
```
shaffeeullah, line 206: I'm not sure this case will ever be relevant, as the default value for `publicAccessPrevention` is "unspecified". This field should never be null.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -184,17 +212,23 @@ static IamConfiguration fromPb(Bucket.IamConfiguration iamConfiguration) {
       Bucket.IamConfiguration.UniformBucketLevelAccess uniformBucketLevelAccess =
           iamConfiguration.getUniformBucketLevelAccess();
       DateTime lockedTime = uniformBucketLevelAccess.getLockedTime();
+      String publicAccessPrevention = iamConfiguration.getPublicAccessPrevention();
 
       return newBuilder()
           .setIsUniformBucketLevelAccessEnabled(uniformBucketLevelAccess.getEnabled())
           .setUniformBucketLevelAccessLockedTime(lockedTime == null ? null : lockedTime.getValue())
+          .setPublicAccessPrevention(
+              publicAccessPrevention == null
```
shaffeeullah, line 221: publicAccessPrevention should never be null

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3235,6 +3235,85 @@ public void testEnableAndDisableUniformBucketLevelAccessOnExistingBucket() throw
...
+    String papBucket = RemoteStorageHelper.generateBucketName();
+    try {
+      BucketInfo.IamConfiguration iamConfiguration =
+          BucketInfo.IamConfiguration.newBuilder()
+              .setIsUniformBucketLevelAccessEnabled(true)
+              .build();
+      Bucket bucket =
+          storage.create(
+              Bucket.newBuilder(papBucket).setIamConfiguration(iamConfiguration).build());
+      assertNull(bucket.getIamConfiguration().getPublicAccessPrevention());
```
shaffeeullah, line 3249: publicAccessPrevention should never be null

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
frankyn: If a request uses a field selector which does not include iamConfiguration, then this value will be null because it's not returned by the service.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
frankyn: Same here, if the value is never set/used then it stays null.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
frankyn: This looks like a bug, will follow-up!

frankyn: Blocked for now as this is a backend bug as discussed offline. Thanks @shaffeeullah and @tritone

shaffeeullah: @frankyn Can you please remove this check in the integration test? Public access prevention should not be null, but also I think this is validating the API functionality rather than the client library functionality. I don't think we should be checking the public access prevention default.

frankyn: Gotcha, so just validate that public data is not accessible when enabled and is accessible when disabled. Will follow-up

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -101,6 +101,21 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
   private final String locationType;
   private final Logging logging;
 
+  public enum PublicAccessPrevention {
```
BenWhitehead, line 104: Do we have some docs we can point to to provide some context for why this enum exists? Maybe something like is on `IamConfiguration` below.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -112,6 +127,7 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
 
     private Boolean isUniformBucketLevelAccessEnabled;
     private Long uniformBucketLevelAccessLockedTime;
+    private PublicAccessPrevention publicAccessPrevention;
```
BenWhitehead, line 130: Can this field be final?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -184,17 +212,23 @@ static IamConfiguration fromPb(Bucket.IamConfiguration iamConfiguration) {
...
       DateTime lockedTime = uniformBucketLevelAccess.getLockedTime();
+      String publicAccessPrevention = iamConfiguration.getPublicAccessPrevention();
 
       return newBuilder()
           .setIsUniformBucketLevelAccessEnabled(uniformBucketLevelAccess.getEnabled())
           .setUniformBucketLevelAccessLockedTime(lockedTime == null ? null : lockedTime.getValue())
+          .setPublicAccessPrevention(
+              publicAccessPrevention == null
+                  ? null
+                  : PublicAccessPrevention.valueOf(publicAccessPrevention.toUpperCase()))
```
BenWhitehead, line 223: `valueOf` isn't very safe for direct use without some pre validation as it will throw an `IllegalArgumentException` if the value passed in doesn't match.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -101,6 +101,21 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
   private final String locationType;
   private final Logging logging;
 
+  public enum PublicAccessPrevention {
+    ENFORCED("enforced"),
+    UNSPECIFIED("unspecified");
```
BenWhitehead, line 106: Is one of these two values the "default" value? Can we add a comment calling that out?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3235,6 +3235,116 @@ public void testEnableAndDisableUniformBucketLevelAccessOnExistingBucket() throw
...
+      }
+
+      // Making object public via ACL should fail.
+      try {
+        // Create a public object
+        bucket.create(
+            "pap-test-object",
+            "".getBytes(),
+            Bucket.BlobTargetOption.predefinedAcl(Storage.PredefinedAcl.PUBLIC_READ));
+        fail("pap: expected adding AllUsers ACL to object should fail");
```
BenWhitehead, line 3275: nit: Typo `s/AllUsers/allUsers/`

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3235,6 +3235,116 @@ public void testEnableAndDisableUniformBucketLevelAccessOnExistingBucket() throw
...
+      } catch (StorageException storageException) {
```
BenWhitehead, line 3276: Is there anything distict/informative in the StorageException we should verify?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3235,6 +3235,116 @@ public void testEnableAndDisableUniformBucketLevelAccessOnExistingBucket() throw
...
+
+      // Now, making object public or making bucket public should succeed.
+      try {
+        // Create a public object
+        bucket.create(
+            "pap-test-object",
+            "".getBytes(),
+            Bucket.BlobTargetOption.predefinedAcl(Storage.PredefinedAcl.PUBLIC_READ));
+      } catch (StorageException storageException) {
+        fail("pap: expected adding AllUsers ACL to object to succeed");
```
BenWhitehead, line 3306: nit: Typo `s/AllUsers/allUsers/`

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3235,6 +3235,116 @@ public void testEnableAndDisableUniformBucketLevelAccessOnExistingBucket() throw
...
+                .setVersion(3)
+                .setBindings(
+                    ImmutableList.<com.google.cloud.Binding>of(
+                        com.google.cloud.Binding.newBuilder()
+                            .setRole("roles/storage.objectViewer")
+                            .addMembers("allUsers")
+                            .build()))
+                .build());
+      } catch (StorageException storageException) {
+        fail("pap: expected adding AllUsers policy to bucket to succeed");
```
BenWhitehead, line 3323: nit: Typo `s/AllUsers/allUsers/`

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -3235,6 +3235,116 @@ public void testEnableAndDisableUniformBucketLevelAccessOnExistingBucket() throw
     }
   }
 
+  @Test
+  public void testUnspecifiedAndEnforcedPublicAccessPreventionOnBucket() throws Exception {
```
BenWhitehead, line 3239: nit: This test feels like 3 tests to me
1. When ENFORCED 
    * chaging bucket policy to enforced should fail
    * creating a public object should fail
2. When UNSPECIFIED 
    * chaging bucket policy to enforced should succeed
    * creating a public object should succeed
3. Updating UBLA should not modify PublicAccessPrevention

BenWhitehead (changes requested):
A few minor questions

frankyn (commented):
@BenWhitehead can you take a look when you have a moment?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -101,17 +101,42 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
 
+  /**
+   * Public Access Prevention enum with expected values.
+   *
+   * @see <a
+   *     href="https://cloud.google.com/storage/docs/public-access-prevention">public-access-prevention</a>
+   */
+  public enum PublicAccessPrevention {
+    ENFORCED("enforced"),
+    // Default value for Public Access Prevention
```
BenWhitehead, line 112: A javadoc comment on the enum value will allow it to show up in the generated javadocs
```suggestion
    /**
     * Default value for Public Access Prevention
     */
```

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -235,6 +278,18 @@ Builder setUniformBucketLevelAccessLockedTime(Long uniformBucketLevelAccessLocke
         return this;
       }
 
+      /**
+       * Sets the bucket's Public Access Prevention configuration. Currently supported options are
+       * 'unspecified' or 'enforced'.
```
BenWhitehead, line 283: nit:
```suggestion
       * {@link PublicAccessPrevention#UNSPECIFIED} or {@link PublicAccessPrevention#ENFORCED}
```

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
frankyn: Apologies, missed this.

This would definitely suck. Do you have a suggestion?

BenWhitehead (changes requested):
A couple new nits, and the outstanding valueOf call

BenWhitehead (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -121,6 +122,15 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
     public String getValue() {
       return value;
     }
+
+    public static PublicAccessPrevention parse(String value) {
+      String upper = value.toUpperCase();
+      try {
+        return valueOf(upper);
+      } catch (IllegalArgumentException ignore) {
```
tritone, line 130: Do we still need to catch this exception? Where is it being thrown from now?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -110,7 +110,8 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
   public enum PublicAccessPrevention {
     ENFORCED("enforced"),
     /** Default value for Public Access Prevention */
-    UNSPECIFIED("unspecified");
+    UNSPECIFIED("unspecified"),
+    UNKNOWN(null);
```
tritone, line 114: Let's add a comment to indicate that this will be returned if we receive an unrecognized value from the API.

tritone (commented):
Thanks so much for fixing this up Ben!

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
BenWhitehead: If the string passed to valueOf doesn't match one of the defined enum values, this exception is thrown. We catch it and return unknown so that we don't have to re-implement the traversal and testing logic already implemented in valueOf.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
BenWhitehead: fixed

tritone (approved):
Looks like there is a lint failure; otherwise LGTM

shaffeeullah:
@BenWhitehead & @frankyn ,

We would like to merge and release this feature next Monday (06/14/21). Does that work for you?

shaffeeullah:
Public access prevention rollout has been delayed due to a bug surfaced during Googler preview. I will keep this PR updated as I learn new release timeline details.

shaffeeullah:
@BenWhitehead  This feature can now be merged and released.

PR #679, opened 2021-01-12
fix: unnecessary options in resumable upload URL

JesseLovelace:
Fix unnecessary options in resumable upload URL by explicitly selecting them instead of iterating though all


frankyn (approved):
LGTM, pending kokoro.

PR #813, opened 2021-05-10
feat: update ci config for 1.113.14-sp branch

BenWhitehead:


tritone (commented):
Note that the integration test failure is due to the moved/missing landsat dataset-- we fixed this in Go, C#, etc. See https://github.com/googleapis/google-cloud-go/pull/3806

PR #846, opened 2021-05-26
fix: improve error detection and reporting for BlobWriteChannel retry state

### LINKED ISSUE #839: Storage: `BlobWriteChannel` generates `NullPointerException` in some failure cases (e.g. networking issues)
#### Problem
`BlobWriteChannel` generates `NullPointerException` in some failure cases (e.g. due to a networking error).
Specifically, [BlobWriteChannel:176](https://github.com/googleapis/java-storage/blob/5e9906173ade5434ce1cecbd920445494b375652/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java#L176) might retrieve `null` `StorageObject` which causes [BlobWriteChannel:180](https://github.com/googleapis/java-storage/blob/5e9906173ade5434ce1cecbd920445494b375652/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java#L180) to generate `NullPointerException` when trying to dereference it.

This unhandled exception makes it difficult to clearly resolve issues and resume any impacted operations/scripts. As an engineer/on call, I want logs to be more clear (unnecessary exceptions not to be thrown at all if that's possible), so I can get issues resolved faster with less confusion and spend less time on manual workarounds.

#### Environment details

1. Specify the API at the beginning of the title. For example, "BigQuery: ...").
   General, Core, and Other are also allowed as types  
2. OS type and version: 
3. Java version:
4. storage version(s): 1.113.14 and 1.114.0.

#### Steps to reproduce

 1.  Simulate faulty network (dropped packets ~35%, less works too, but takes more time to catch)
 2. Call `BlobWriteChannel`
 3. See `NullPointerException`

#### Stack trace
```
java.lang.RuntimeException: java.util.concurrent.ExecutionException: ... omitted ...
Caused by: ...: Unable to write ...
... omitted ...
Caused by: java.lang.RuntimeException: Unable to write ... to gs://...
... omitted ...
... 10 more
Caused by: com.google.cloud.storage.StorageException: java.lang.NullPointerException
at com.google.cloud.storage.StorageException.translateAndThrow(StorageException.java:81)
at com.google.cloud.storage.BlobWriteChannel.flushBuffer(BlobWriteChannel.java:212)
at com.google.cloud.BaseWriteChannel.close(BaseWriteChannel.java:151)
at java.base/java.nio.channels.Channels$1.close(Channels.java:177)
at java.base/java.io.FilterOutputStream.close(FilterOutputStream.java:188)
at java.base/java.io.FilterOutputStream.close(FilterOutputStream.java:188)
at org.apache.avro.file.DataFileWriter.close(DataFileWriter.java:461)
... omitted ...
... 14 more
Caused by: java.lang.NullPointerException
at com.google.cloud.storage.BlobWriteChannel$1.run(BlobWriteChannel.java:180)
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
at com.google.api.gax.retrying.DirectRetryingExecutor.submit(DirectRetryingExecutor.java:105)
at com.google.cloud.RetryHelper.run(RetryHelper.java:76)
at com.google.cloud.RetryHelper.runWithRetries(RetryHelper.java:50)
at com.google.cloud.storage.BlobWriteChannel.flushBuffer(BlobWriteChannel.java:157)
```

BenWhitehead:
Add new checks to ensure a more informative error than NullPointerException is thrown if the StorageObject or it's size are unable to be resolved on the last chunk.

Fixes #839

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
@@ -26,6 +26,7 @@ import com.google.cloud.RetryHelper;
 import com.google.cloud.WriteChannel;
 import com.google.cloud.storage.spi.v1.StorageRpc;
 import com.google.common.collect.Maps;
+import java.math.BigInteger;
 import java.net.URL;
 import java.util.Map;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
```diff
@@ -83,14 +84,52 @@ private StorageObject getRemoteStorageObject() {
...
+      long remotePosition,
+      boolean last) {
+    return unrecoverableState(
+        uploadId,
+        chunkOffset,
+        chunkLength,
+        localPosition,
+        remotePosition,
+        last,
+        "Unable to recover in upload.\nThis may be a symptom of multiple clients uploading to the same upload session.");
```
tritone, line 101: I don't think multiple clients uploading to the same session is possible? The session should be created by the client and then only used by that client AFAIK.

tritone (commented):
Generally seems like a good improvement, one question

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
BenWhitehead: @frankyn Do you have context on why this was the original message? (I refactored it to a parameter so I could reuse the details formatting code).

frankyn: It's possible if the upload URI is shared between clients. GCS API supports it, however it's very unlikely it's a possible case.

frankyn (approved):

PR #852, opened 2021-05-27
fix: update BucketInfo translation code to properly handle lifecycle rules

### LINKED ISSUE #850: Storage#update always removes all lifecycle rules if they are not specified
Issue with `com.google.cloud.storage.Storage#update(com.google.cloud.storage.BucketInfo, com.google.cloud.storage.Storage.BucketTargetOption...)`

#### Expected behaviour:
update only fields that are specified in `BucketInfo` parameter.

#### Actual behaviour:
along the updated fields, removes lifecycle rules if they are not specified.

Example code:
```scala
    val bucketId = "[...]"

    val client = StorageOptions.getDefaultInstance.getService
    val before = client.get(bucketId)
    println(s"Before: ${before.getLifecycleRules.size()}")

    val updateRequest = BucketInfo
      .newBuilder(bucketId)
      .setLabels(singletonMap("test", "test"))
      .build
    client.update(updateRequest)

    val after = client.get(bucketId)
    println(s"After: ${after.getLifecycleRules.size()}")
```

Actual output (for bucket with 1 lifecycle rule):
```txt
Before: 1
After: 0
```

I suspect that issue is located here:
https://github.com/googleapis/java-storage/blob/accf5021e99a083aa4b3e8935d8ccd0e3668ae6f/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java#L1841-L1869

BenWhitehead:
Fixes #850

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -1843,11 +1843,9 @@ public class BucketInfo implements Serializable {
     if (deleteRules != null || lifecycleRules != null) {
       Lifecycle lifecycle = new Lifecycle();
 
-      if (
-          (deleteRules == null && lifecycleRules.isEmpty())
+      if ((deleteRules == null && lifecycleRules.isEmpty())
           || (lifecycleRules == null && deleteRules.isEmpty())
-          || (deleteRules != null && deleteRules.isEmpty() && lifecycleRules.isEmpty())
-      ) {
+          || (deleteRules != null && deleteRules.isEmpty() && lifecycleRules.isEmpty())) {
         lifecycle.setRule(Collections.<Rule>emptyList());
       } else {
         Set<Rule> rules = new HashSet<>();
@@ -1878,7 +1876,7 @@ public class BucketInfo implements Serializable {
           lifecycle.setRule(ImmutableList.copyOf(rules));
         }
       }
-      
+
       bucketPb.setLifecycle(lifecycle);
     }
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -1862,7 +1862,7 @@ public class BucketInfo implements Serializable {
               }));
     }
 
-    if (rules != null) {
+    if (!rules.isEmpty()) {
       Lifecycle lifecycle = new Lifecycle();
       lifecycle.setRule(ImmutableList.copyOf(rules));
       bucketPb.setLifecycle(lifecycle);
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -1843,6 +1843,14 @@ public class BucketInfo implements Serializable {
     if (deleteRules != null || lifecycleRules != null) {
       Lifecycle lifecycle = new Lifecycle();
 
+      // Here we determine if we need to "clear" any defined Lifecycle rules by explicitly setting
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1862,7 +1862,7 @@ public Rule apply(LifecycleRule lifecycleRule) {
               }));
     }
 
-    if (rules != null) {
+    if (!rules.isEmpty()) {
```
frankyn, line 1865: Will this support rules deletion as well? I'm wondering if rules should be set to null at construction time and only have a non-null value when rules are returned.

BenWhitehead: I'll test and update if necessary

frankyn: kokoro failed on this case

BenWhitehead: Added some unit tests and refactored things afterward to account for clearing rules as well.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1838,33 +1839,44 @@ public ObjectAccessControl apply(Acl acl) {
...
-    }
 
-    if (rules != null) {
+    if (deleteRules != null || lifecycleRules != null) {
       Lifecycle lifecycle = new Lifecycle();
-      lifecycle.setRule(ImmutableList.copyOf(rules));
+
+      if ((deleteRules == null && lifecycleRules.isEmpty())
+          || (lifecycleRules == null && deleteRules.isEmpty())
+          || (deleteRules != null && deleteRules.isEmpty() && lifecycleRules.isEmpty())) {
```
frankyn, line 1848: This got very complex, could you add a comment above this condition to clarify why it's needed?

frankyn (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
BenWhitehead: Comment added PTAL

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1843,6 +1843,14 @@ public ObjectAccessControl apply(Acl acl) {
...
+      // Here we determine if we need to "clear" any defined Lifecycle rules by explicitly setting
+      // the Rule list of lifecycle to the empty list.
+      // In order for us to clear the rules, one of the three following must be true:
+      //   1. deleteRules is null while lifecycleRules is non-null and empty
+      //   2. lifecycleRules is null while deleteRules is non-null and empty
+      //   3. lifecycleRules is non-null and empty while deleteRules is non-null and empty
+      // If none of the above three is true, we will interpret as the Lifecycle rules being
+      // updated to the defined set of DeleteRule and LifecycleRule.
       if ((deleteRules == null && lifecycleRules.isEmpty())
           || (lifecycleRules == null && deleteRules.isEmpty())
```
frankyn, line 1855: should you check that `deleteRules != null` here as well?

frankyn (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
BenWhitehead: `deleteRules != null` is already true by the time we reach this branch of the if statement, and by virtue of the outer guarding if statement. Either `lifecycleRules` or `deleteRules` is non-null, so there is no need to check it again.

frankyn (approved):

PR #872, opened 2021-06-09
feat: Add shouldReturnRawInputStream option to Get requests

### LINKED ISSUE #321: Support reading gzip files as-is with google-cloud-storage?
**Is your feature request related to a problem? Please describe.**
https://cloud.google.com/storage/docs/transcoding describes that if I upload a file with content-encoding set to gzip, it's automatically decompressed for read, unless I specify `Accept-Encoding: gzip` in my read request. However, I can't find an obvious way to set the header with `google-cloud-storage`, so I wonder if it supports to do so, or if there's another way to read gzipped file as is with the library. 
Note that sometimes I do want to get the file back unzipped, so I want to keep setting content-encoding to gzip when I save the object.

 **Describe the solution you'd like**
`google-cloud-storage` should provide an option to set Accept-Encoding when downloading a blob. For example, make it an option of `Blob.downloadTo` maybe.

JesseLovelace:
Adds support for shouldReturnRawInputStream, which will allow users to specify whether the client should auto-decompress gzipped content or not. 

Note: `read(StorageObject, Map<Option, ?>, long, OutputStream)` and `read(StorageObject, Map<Option, ?>, long, int)` in HttpStorageRpc are set up to have different defaults to preserve their original behavior. Before, returnRawInputStream was always set to true, however `read(StorageObject, Map<Option, ?>, long, OutputStream)` had a bug that would cause that to be lost, so it would always turn out to be false (this is why blob.downloadTo and BlobReadChannel.read had different behaviors). That bug is fixed here, but these method signatures now have different defaults to preserve the original behavior. Users can use the new flag to change that behavior 

Fixes #321

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -118,7 +118,7 @@ public class Blob extends BlobInfo {
         case USER_PROJECT:
           return Storage.BlobSourceOption.userProject((String) getValue());
         case RETURN_RAW_INPUT_STREAM:
-                return Storage.BlobSourceOption.shouldReturnRawInputStream((boolean) getValue());
+          return Storage.BlobSourceOption.shouldReturnRawInputStream((boolean) getValue());
         default:
           throw new AssertionError("Unexpected enum value");
       }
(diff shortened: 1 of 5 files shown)


frankyn (changes requested):

oripwk:
@JesseLovelace Does it work also with `getContent()` or only with `downloadTo()`?

JesseLovelace:
@oripwk Should work with `getContent`, use `getContent(Blob.BlobSourceOption.shouldReturnRawInputStream(true));`

oripwk:
@JesseLovelace when is it going to be released?

oripwk:
Good, I see it's on 1.116.0

marioaae:
Hi there, thanks for fixing this, unfortunately I haven't been able to make it work on my end. In Scala I'm trying to read files through a GZIPInputStream as follows:

```
val bucketName       = configuration.get[String]("backup.bucketName")
val filePath                = "path/to/file.gz
val storage: Storage = StorageOptions.getDefaultInstance.getService

val blob: Blob = storage.get(BlobId.of(bucketName, fullPath), Storage.BlobGetOption.shouldReturnRawInputStream(true))
val reader: ReadChannel = blob.reader()
val inputStream: InputStream = Channels.newInputStream(reader)
```

Afterwards I'm wrapping it up in a GZIPInputStream to process it as follows: `new GZIPInputStream(inputStream)` but I get the following error:

`java.util.zip.ZipException: Not in GZIP format`

Still, if I only use the InputStream it works, so it seems to me that the file is being downloaded decompressed, I thought the `shouldReturnRawInputStream(true)` would download my file compressed in the original .gz format.

When I download the file from a different source (e.g. an SFTP server) where I know for sure no decompression is going on, wrapping it up in GZIPInputStream works properly.

Has anybody had the same problem?

Thanks!

Issue #868, opened 2021-06-07
Add fromStorageUrl and toStorageUrl methods to BlobId


yihanzhen:
porting over an old issue in an old repo over to have it triaged: https://github.com/googleapis/google-cloud-java/issues/2397

PR #888, opened 2021-06-18
feat: Add from and to storage url options for BlobId

JesseLovelace:
Adds options for creating a blob ID from a storage url and vice versa. Fixes https://github.com/googleapis/java-storage/issues/868

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobId.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobId.java
@@ -58,7 +58,7 @@ public final class BlobId implements Serializable {
   }
 
   /** Returns this blob's Storage url which can be used with gsutil */
-  public String toStorageUrl() {
+  public String toGsUtilUri() {
     return "gs://" + bucket + "/" + name;
   }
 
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobId.java
```diff
@@ -114,6 +120,21 @@ public static BlobId of(String bucket, String name, Long generation) {
     return new BlobId(checkNotNull(bucket), checkNotNull(name), generation);
   }
 
+  /**
+   * Creates a {@code BlobId} object.
+   *
+   * @param storageUrl the Storage url to create the blob from
+   */
+  public static BlobId fromStorageUrl(String storageUrl) {
```
BenWhitehead, line 128: Maybe `fromGsutilUri` instead of storage? To my knowledge, there are many different types of storage URLs
* `https://storage.googleapis.com/bucket/blob`
* `https://storage.cloud.google.com/bucket/blob`

In the cloud console these format of URI are labeled `gsutil URI`
image

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobId.java
```diff
@@ -114,6 +120,21 @@ public static BlobId of(String bucket, String name, Long generation) {
...
+    if(!Pattern.matches("gs://.*/.*", storageUrl)) {
+      throw new IllegalArgumentException(storageUrl + " is not a valid Storage URL");
+    }
+    String bucketName = storageUrl.split("/")[2];
+    String blobName = storageUrl.split(bucketName + "/")[1];
```
BenWhitehead, line 133: Possibly simplify string allocation and traversal?

```suggestion
    int blobNameStartIndex = storageUrl.indexOf('/', 5);
    String bucketName = storageUrl.substring(5, blobNameStartIndex);
    String blobName = storageUrl.substring(blobNameStartIndex);
```

BenWhitehead (approved):

PR #898, opened 2021-06-23
docs: add notice about broken OLM experience

JesseLovelace:
Adds a notice to the docs regarding the broken OLM experience in old versions of the library


danielduhh (approved):
👍🏿

PR #918, opened 2021-07-09
fix: correct lastChunk retry logic in BlobWriteChannel

BenWhitehead:
Add new method StorageRpc#queryResumableUpload which allows getting a shallow StorageObject for a resumable upload session which is complete.

Update BlobWriteChannel to use StoageRpc#queryResumableUpload instead of StorageRpc#get when attempting to validate the upload size of an object when it determines the upload is complete and is on the last chunk.

If a BlobWriteChannel is opened with a conditional like IfGenerationMatch it is not possible to simply get the object, as the object can drift generationally while the resumable upload is being performed.

Related to #839

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -821,8 +821,6 @@ public class HttpStorageRpc implements StorageRpc {
       // If the response is 200
       if (response.getStatusCode() == 200) {
         return response.parseAs(StorageObject.class);
-        // String s = response.parseAsString();
-        // throw new RuntimeException("kaboobm");
       } else {
         throw buildStorageException(response.getStatusCode(), response.getStatusMessage());
       }
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -811,7 +811,7 @@ public class HttpStorageRpc implements StorageRpc {
   }
 
   @Override
-  public StorageObject queryResumableUpload(String uploadId, long totalBytes) {
+  public StorageObject queryCompletedResumableUpload(String uploadId, long totalBytes) {
     try {
       GenericUrl url = new GenericUrl(uploadId);
       HttpRequest req = storage.getRequestFactory().buildPutRequest(url, new EmptyContent());
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -811,6 +811,27 @@ public class HttpStorageRpc implements StorageRpc {
   }
 
   @Override
+  public StorageObject queryResumableUpload(String uploadId, long totalBytes) {
+    try {
+      GenericUrl url = new GenericUrl(uploadId);
+      HttpRequest req = storage.getRequestFactory().buildPutRequest(url, new EmptyContent());
+      req.getHeaders().setContentRange(String.format("bytes */%s", totalBytes));
+      req.setParser(storage.getObjectParser());
+      HttpResponse response = req.execute();
+      // If the response is 200
+      if (response.getStatusCode() == 200) {
+        return response.parseAs(StorageObject.class);
+        // String s = response.parseAsString();
+        // throw new RuntimeException("kaboobm");
+      } else {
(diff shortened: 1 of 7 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -810,6 +810,27 @@ public long getCurrentUploadOffset(String uploadId) {
     }
   }
 
+  @Override
+  public StorageObject queryResumableUpload(String uploadId, long totalBytes) {
```
tritone, line 814: Could use a docstring. What is this used for and when during the upload should it be called? What does totalBytes represent here?

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -810,6 +810,27 @@ public long getCurrentUploadOffset(String uploadId) {
...
+    try {
+      GenericUrl url = new GenericUrl(uploadId);
+      HttpRequest req = storage.getRequestFactory().buildPutRequest(url, new EmptyContent());
+      req.getHeaders().setContentRange(String.format("bytes */%s", totalBytes));
+      req.setParser(storage.getObjectParser());
+      HttpResponse response = req.execute();
+      // If the response is 200
+      if (response.getStatusCode() == 200) {
+        return response.parseAs(StorageObject.class);
+        // String s = response.parseAsString();
```
tritone, line 824: Remove print debugging

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -810,6 +810,27 @@ public long getCurrentUploadOffset(String uploadId) {
...
+        // throw new RuntimeException("kaboobm");
+      } else {
+        throw buildStorageException(response.getStatusCode(), response.getStatusMessage());
```
tritone, line 827: If I understand correctly, this means that the caller would have made a request to do an upload, but they get an error from the request to check the upload status which is a separate RPC. This might be confusing-- is there a way to wrap this in an error that corresponds to the method they originally called?

Also, if the response is a 308, I would assume that instead of failing we should retry from the last offset, no?

(Ignore me if I'm just missing how this works).

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
JesseLovelace: +1

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
BenWhitehead: Normally, when a resumable upload completes the shallow Object will be returned in the response. If the response is not received for any reason (network failure usually), this method allows for querying for that shallow Object. This method is used as part of the retry evaluation logic to determine if a final chunk must be retransmitted or not. This method will only return cleanly if the resumable upload returns 200, if a 308 is returned due to an incomplete upload a `StorageException` will be thrown here to signal that fact. In the case of `BlobWriteChannel` which uses this method, the exception would be caught by the retry handler and be evaluated there.

I've added a javadoc to the method on there interface with an explanation of why it's there.

tritone: Cool, the javadoc really helps, thank you.

So, in the 308 case-- would we then (if in `BlobWriteChannel`) make _another_ identical PUT request to query the correct offset to retry at? If so, is there a way to preserve the information from the 308 response here instead?

BenWhitehead: In the case of `BlobWriteChannel` it performs a pre-check to try and determine the remote offset, if the remote offset indicates that the data has been written but for some reason it wasn't able to get the object metadata as part of the final `PUT`, it will then use this method to try and get that object metadata.

BenWhitehead: If a 308 happens, that would land us in the normal retry flow, where right now unfortunately we don't have the ability to rewind and retry a chunk to try and recover. 

In practice, with `BlobWriteChannel` any call to this new method is guarded by a check to see if the resumable session is complete or not (in fact I would have prefered to make this check an internal implementation detail but unfortunately `BlobWriteChannel` does not have the means of accessing the http client directly in order to be able to make this call). So, for all intents and purposes we side step the need to be able to rewind a chunk.

tritone (approved):

PR #924, opened 2021-07-20
feat: fix post policy escape bug, update conformance tests

JesseLovelace:
V4 Post Policy had an issue where quotes in the json policy would get double escaped (`"` would became `\\"` instead of `\"`), this fixes that issue and updates the conformance tests to the latest version

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
@@ -443,13 +443,21 @@ public final class PostPolicyV4 {
       StringBuilder escapedJson = new StringBuilder();
 
       // Certain characters in a policy must be escaped
-      for (char c : json.toCharArray()) {
+      char[] jsonArray = json.toCharArray();
+      for (int i = 0; i < jsonArray.length; i++) {
+        char c = jsonArray[i];
         if (c >= 128) { // is a unicode character
           escapedJson.append(String.format("\\u%04x", (int) c));
         } else {
           switch (c) {
             case '\\':
-              escapedJson.append("\\\\");
+              // The JsonObject/JsonArray operations above handle quote escapes, so leave any "/""
+              // found alone
+              if (jsonArray[i + 1] == '"') {
+                escapedJson.append("\\");
+              } else {
+                escapedJson.append("\\\\");
+              }
               break;
             case '\b':
               escapedJson.append("\\b");
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
@@ -451,7 +451,7 @@ public final class PostPolicyV4 {
         } else {
           switch (c) {
             case '\\':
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
```diff
@@ -443,13 +443,21 @@ public String toJson() {
...
+      char[] jsonArray = json.toCharArray();
+      for (int i = 0; i < jsonArray.length; i++) {
+        char c = jsonArray[i];
         if (c >= 128) { // is a unicode character
           escapedJson.append(String.format("\\u%04x", (int) c));
         } else {
           switch (c) {
             case '\\':
-              escapedJson.append("\\\\");
+              // The JsonObject/JsonArray operations above handle quote escapes, so leave any "/""
```
elharo, line 454: is forward slash meant here? Looks like it should be a backslash

google-cloud-storage/src/test/java/com/google/cloud/storage/V4PostPolicyTest.java
```diff
@@ -166,17 +154,25 @@ public void test() {
             style);
 
     String expectedPolicy = testData.getPolicyOutput().getExpectedDecodedPolicy();
+
     StringBuilder escapedPolicy = new StringBuilder();
 
     // Java automatically unescapes the unicode escapes in the conformance tests, so we need to
     // manually re-escape them
-    for (char c : expectedPolicy.toCharArray()) {
+    char[] expectedPolicyArray = expectedPolicy.toCharArray();
```
elharo, line 162: This appears to be reproducing the algorithm in the test. That risks copying bugs from one place to another, and verifying the buggy behavior. Can you make the expected output a literal?

pom.xml
```diff
@@ -93,7 +93,7 @@
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-conformance-tests</artifactId>
-        <version>0.0.11</version>
+        <version>0.1.1</version>
```
elharo, line 96: Does this verify the change?

elharo (commented):
Is there a bug reference for this?

google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java
JesseLovelace: Should be, will fix

google-cloud-storage/src/test/java/com/google/cloud/storage/V4PostPolicyTest.java
JesseLovelace: Yeah, they both happen to need to do the same thing (escaping certain characters). Sorry, I'm unsure what you mean by "make it a literal," can you elaborate?

pom.xml
JesseLovelace: Yeah, in fact the failures caused by it is what brought the bug to my attention (https://github.com/googleapis/java-storage/pull/510)

BenWhitehead (approved):
Feel free to merge after fixing the typo

PR #939, opened 2021-07-30
test: add retry conformance tests

BenWhitehead:
Adds the Retry Conformance Tests which were added in https://github.com/googleapis/conformance-tests/pull/51

This PR adds the following:
1. A JUnit Rule (`@ClassRule`) which will start and stop the storage-testbench server, while also providing a handle to the `/retry_test` REST Api of the testbench server
2. A JUnit Rule (`@Rule`) which will add the test scenario to the server, and verify completion of the test
3. The code necessary to load and bootstrap each test case based off the scenarios defined in https://github.com/googleapis/conformance-tests/pull/51
4. An initial skeleton of the RPC method to API Method binding


google-cloud-storage/src/test/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage;
+
+import com.google.cloud.storage.BucketInfo.BuilderImpl;
+
+public final class PackagePrivateMethodWorkarounds {
```
tritone, line 21: Could you add a comment about why this is needed?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
```diff
@@ -0,0 +1,124 @@
...
+          Bucket resolvedBucket = ctx.getStorage().create(bucketInfo);
+          return ctx.map(s -> s.with(resolvedBucket));
+        };
+    private static final CtxFunction object =
+        (ctx, c) -> {
+          BlobInfo blobInfo =
+              BlobInfo.newBuilder(ctx.getState().getBucket().getName(), c.getObjectName()).build();
+          Blob resolvedBlob = ctx.getStorage().create(blobInfo);
+          return ctx.map(s -> s.with(resolvedBlob));
+        };
```
tritone, line 80: Assuming HMAC key and Notification will go here as well eventually?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
```diff
@@ -0,0 +1,124 @@
...
+        (ctx, c) -> ctx.map(state -> state.with(BlobInfo.newBuilder(state.getBlobId()).build()));
+  }
+
+  static final class Local {
+    static final CtxFunction blobCopy =
+        (ctx, c) -> ctx.map(s -> s.withCopyDest(BlobId.of(c.getBucketName2(), c.getObjectName())));
+
+    static final CtxFunction bucketInfo =
+        (ctx, c) -> ctx.map(s -> s.with(BucketInfo.of(c.getBucketName())));
+    static final CtxFunction blobIdWithoutGeneration =
```
tritone, line 47: Where are these actually used?

google-cloud-storage/src/test/resources/com/google/cloud/storage/conformance/retry/testNamesWhichShouldSucceed.txt
```diff
@@ -0,0 +1,121 @@
+# Each line should be a full test name
+# Each test name present in the file will be expected to pass in a CI environment
+#
+# This list can be regenerated by running the following command in a shell:
```
tritone, line 4: This is nice.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/GracefulConformanceEnforcement.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Set;
+import org.junit.AssumptionViolatedException;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public final class GracefulConformanceEnforcement implements TestRule {
```
tritone, line 32: What does "graceful" mean in this context? Just that it will not hard-fail?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/GracefulConformanceEnforcement.java
```diff
@@ -0,0 +1,96 @@
...
+  private final String testName;
+  private final Set<String> testNamesWhichShouldSucceed;
+
+  public GracefulConformanceEnforcement(String testName) {
+    this.testName = testName;
+    this.testNamesWhichShouldSucceed = loadTestNamesWhichShouldSucceed();
+  }
+
+  @Override
+  public Statement apply(Statement base, Description description) {
```
tritone, line 43: Very clever!

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
```diff
@@ -0,0 +1,301 @@
...
+  private static final Logger LOGGER = Logger.getLogger(ITRetryConformanceTest.class.getName());
+
+  private static final String HEX_SHUFFLE_SEED_OVERRIDE =
+      System.getProperty("HEX_SHUFFLE_SEED_OVERRIDE");
+
+  private static final Set<RpcMethod> readBucketMethods =
+      ImmutableSet.of(buckets.get, buckets.list);
+  private static final List<List<String>> instructionSets =
+      ImmutableList.of(ImmutableList.of("return-503", "return-503") /*,
+      ImmutableList.of("return-400", "return-400")*/);
```
tritone, line 73: Does this need to be in code and not fetched from the proto? Or is this a TODO?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestBench.java
```diff
@@ -0,0 +1,307 @@
...
+    resp.disconnect();
+    return b.build();
+  }
+
+  @Override
+  public Statement apply(final Statement base, Description description) {
+    return new Statement() {
+      @Override
+      public void evaluate() throws Throwable {
+        Path tempDirectory = Files.createTempDirectory("retry-conformance-server");
```
tritone, line 132: What happens with this file? Is it saved for debugging or just redirected to something that will be cleaned up?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestBench.java
```diff
@@ -0,0 +1,307 @@
...
+          + '\''
+          + ", completed="
+          + completed
+          + ", instructions="
+          + instructions
+          + '}';
+    }
+  }
+
+  static final class Builder {
```
tritone, line 260: This is really interesting. I'd always envisioned the flow as something like 1. CI script spins up server then 2. CI script kicks off language specific test build. Instead, you are having the java test session itself drive launching the server. Seems like a fine approach but just wanted to call it out-- do you see any risks here?

google-cloud-storage/src/test/resources/com/google/cloud/storage/conformance/retry/hello-world.txt
```diff
@@ -0,0 +1 @@
+Hello, World!!!
```
tritone, line 1: Do we need this checked in? I think python uses a `NamedTemporaryFile`, not sure if java has an equivalent.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestRetryConformance.java
```diff
@@ -0,0 +1,202 @@
...
+    URL url = cl.getResource("com/google/cloud/storage/conformance/retry/hello-world.txt");
+    assertNotNull(url);
+    try {
+      return Paths.get(url.toURI());
+    } catch (URISyntaxException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static ServiceAccountCredentials resolveServiceAccountCredentials() {
```
tritone, line 190: Why is this needed? Do you have some WithoutAuth type option for your client constructor?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMapping.java
```diff
@@ -0,0 +1,187 @@
...
+        return test.apply(ctx, c);
+      } else {
+        try {
+          test.apply(ctx, c);
+          fail("expected failure, but succeeded");
+        } catch (StorageException e) {
+          boolean matchExpectedCode = false;
+          int code = e.getCode();
+          HashSet<String> instructions = newHashSet(c.getInstruction().getInstructionsList());
+          if (instructions.contains("return-503") && code == 503) {
```
tritone, line 85: This seems like it needs a TODO for when you have the full set of test cases from the JSON.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
```diff
@@ -0,0 +1,1942 @@
...
+    OptionalInt max =
+        funcMap.values().stream().map(RpcMethodMapping::getMappingId).mapToInt(i -> i).max();
+    if (max.isPresent()) {
+      LOGGER.info(String.format("Current max mapping index is: %d%n", max.getAsInt()));
+    } else {
+      throw new IllegalStateException("No mappings defined");
+    }
+  }
+
+  static final class Mappings {
```
tritone, line 230: Soooo many mappings.

tritone (commented):
This is a lot of code and generally looking good; some initial thoughts.

As a general comment: the functional approach to building fixtures, test cases, etc is very clever, but I'd really want to make sure you document stuff at least to the point that another java programmer (who's not you) could easily debug an issue or add a new test case. I'm not sure if I'm just out of my depth as a non-java person but it's quite a lot to wade through and grok seemingly.

google-cloud-storage/src/test/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
BenWhitehead: Yeah, I'll add a comment.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
BenWhitehead: Yup, I actually added it this morning.

BenWhitehead: Notification is turning out to be a bit of a unique case in the java library as it's not part of the `Storage` api, only the `StorageRPC` api which doesn't have any retry at all in that layer. So it'll take some thinking to figure out how we're going to manage that.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
BenWhitehead: These are used in the mappings and they function as a selective composable setup step since the API we have with JUnit is quite restrictive for this type of dynamic test generation we're doing.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/GracefulConformanceEnforcement.java
BenWhitehead: Correct, the graceful essentially means "If the test is ran in a CI environment and hasn't been indicated as succeeding in the past, turn the failure into an ignore so the build isn't failed".

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/GracefulConformanceEnforcement.java
BenWhitehead: Thanks!

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
BenWhitehead: These are some pre-filters that are hanging around for a bit while we dig into the failing tests. Since the full tests suite is ~700 tests, we use these filters to pair down the set of tests that are even emitted to the test runner. Allowing for filtering like "Only run the cases where the instructions are [503, 503]".

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestBench.java
BenWhitehead: This directly will contain two temp files `stdout` and `stderr` containing the redirected output from the testbench server.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestBench.java
BenWhitehead: That's fair, with this approach I've ensured anyway someone can start the JUnit test they will be able to run (as long as they have docker) rather than having to coordinate running a server first manually before running the tests.

google-cloud-storage/src/test/resources/com/google/cloud/storage/conformance/retry/hello-world.txt
BenWhitehead: Some of the java api accepts a file path so I did this file to satisfy that requirement. If we don't want the test resource it could be refactored.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestRetryConformance.java
BenWhitehead: One of the code paths which can be used to sign things is only able to be used when it has a `com.google.auth.ServiceAccountSigner` `ServiceAccountCredentials`implements the Signer interface and is fairly easy to construct as long as we have a json (I'm reusing the one from the signing tests). This will only have an impact on the case we need to sign a url before uploading to it, otherwise the storage client is configured with `NoCredentials`.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
BenWhitehead: Yeah, and the code formatter doesn't really help with condensing at all 😩

google-cloud-storage/src/test/resources/com/google/cloud/storage/conformance/retry/hello-world.txt
tritone: Ah yeah, I meant that python has the same requirement but they use [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) to create a valid file path which gets generated and cleaned up each time. Not sure if there is something that Java offers.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestBench.java
cojenco: Interesting way +1  Discovered package `subprocess` in python that can do something similar. It's working nice locally, but I haven't incorporated it with kokoro yet. Will we have trouble running the storage-testbench docker image within the kokoro trampoline docker container?

google-cloud-storage/src/test/resources/com/google/cloud/storage/conformance/retry/testNamesWhichShouldSucceed.txt
```diff
@@ -0,0 +1,322 @@
...
+#    mvn test && xq '//testsuite/testcase[not(./error) and not(./failure) and not(./skipped)]/@name' google-cloud-storage/target/surefire-reports/TEST-com.google.cloud.storage.conformance.retry.ITRetryConformanceTest-sponge_log.xml | tail -n+2 | head -n-1 | sed 's#  <result>test\[##g' | sed 's#\]</result>##g' | sort | tee -a google-cloud-storage/src/test/resources/com/google/cloud/storage/conformance/retry/testNamesWhichShouldSucceed.txt
+# where xq is the package from https://github.com/jeffbr13/xq
+
+TestRetryConformance/1-[return-503_return-503]-storage.bucket_acl.list-7
+TestRetryConformance/1-[return-503_return-503]-storage.bucket_acl.list-8
+TestRetryConformance/1-[return-503_return-503]-storage.bucket_acl.list-90
```
cojenco, line 10: nit: Do we have to follow the naming guidelines from the design doc? Since there are way less instruction cases than mapping library methods, perhaps it would make more sense to expose the "library method name" and use "instruction index" instead?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestRetryConformance.java
tritone: Interesting-- does the emulator even have support for signed URL uploads in any meaningful way? Is it possible to just have the test give a fake signed URL/signature to upload to?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestBench.java
BenWhitehead: The kokoro trampoline does all the necessary wiring so that docker containers can be ran from within the trampoline as well.

In fact, you can see the test results from one of the integration tests builds [here](https://source.cloud.google.com/results/invocations/2d9a7f9a-8c95-42cc-9707-1666013ff9e0/targets/github%2Fjava-storage%2Fgoogle-cloud-storage%2Ftarget%2Ffailsafe-reports%2Fintegration%2Fcom.google.cloud.storage.conformance.retry.ITRetryConformanceTest/tests).

google-cloud-storage/src/test/resources/com/google/cloud/storage/conformance/retry/hello-world.txt
BenWhitehead: It's something we'd have to supply.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestRetryConformance.java
BenWhitehead: I'm not totally sure, admittedly my understanding of signed urls is still ramping up so I'm not sure if we are able to use a fake signed url. Something we'll have to look into.

google-cloud-storage/src/test/resources/com/google/cloud/storage/conformance/retry/testNamesWhichShouldSucceed.txt
BenWhitehead: in the case of java, there are often multiple library method names which are underlying these. 

In this case for bucket_acl.list
* ` 7` -> `com.google.cloud.storage.Storage.listAcls(String bucketName)`
* ` 8` -> `com.google.cloud.storage.Storage.listAcls(String bucketName, BucketSourceOption... options)`
* `90` -> `com.google.cloud.storage.Bucket.listAcls()`

When looking at the results of a test run, I think it's easier for someone to read the lines as defined by the design doc with the understanding, there are several language specific under each of the final index but knowing which instructions were used without having to memorize or refer to indexes from the test definition.

Use the instruction names are also resilient to any re-ordering which may happen as we add more cases in the future.

JesseLovelace (commented):
I agree with chris about more documentation in general, it took me a while to get my head around this. Looks generally good though, just a few notes. Also, there's lots of TODOs in the mappings, are you planning to address those after shipping?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
```diff
@@ -0,0 +1,301 @@
...
+    RetryTests.Builder testBuilder = RetryTests.newBuilder();
+    JsonFormat.parser().merge(reader, testBuilder);
+    return testBuilder.build();
+  }
+
+  private static List<Object[]> generateTestCases(
+      RpcMethodMappings rpcMethodMappings, List<RetryTest> testCases) {
+    String host = TEST_BENCH.getBaseUri().replaceAll("https?://", "");
+
+    List<Object[]> data = new ArrayList<>(testCases.size());
```
JesseLovelace, line 203: nitpick: Why this initial capacity? Currently, testCases.size() is 5 (always_idempotent, conditionally_idempotent_retries_when_precondition_is_present, conditionally_idempotent_no_retries_when_precondition_is_absent, non_idempotent, non-retryable errors), but the actual length of this list will end up much bigger, i.e. 737 with the current mappings

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+    return data;
+  }
+
+  private static Random getRand() throws NoSuchAlgorithmException {
+    long seed;
+    if (HEX_SHUFFLE_SEED_OVERRIDE != null) {
+      LOGGER.info(
+          "Shuffling test order using Random with override seed: " + HEX_SHUFFLE_SEED_OVERRIDE);
+      // bytes = Base64.getDecoder().decode(BASE64_SHUFFLE_SEED_OVERRIDE);
```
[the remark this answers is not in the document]
BenWhitehead: yeah, I've cleaned it up

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
BenWhitehead: Left over, I've removed the explicit sizing.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
```diff
@@ -0,0 +1,1942 @@
...
+
+  private void reportMappingSummary() {
+    int mappingCount = funcMap.values().stream().mapToInt(m -> 1).sum();
+    LOGGER.info("Current total number of mappings defined: " + mappingCount);
+    String counts =
+        funcMap.asMap().entrySet().stream()
+            .map(
+                e -> {
+                  RpcMethod rpcMethod = e.getKey();
+                  Collection<RpcMethodMapping> mapings = e.getValue();
```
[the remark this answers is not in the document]
BenWhitehead: Fixed

BenWhitehead:
I've added javadocs to all classes and several methods which should help clarify things.

I also took a pass at refactoring ITRetryConformanceTest to try and help make it more clear what is happening with the test resolution. The test resolution is now factored into its own helper class and provides a builder which puts names to the inputs needed for the resolution.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethod.java
```diff
@@ -0,0 +1,167 @@
...
+
+      @Override
+      public String getFullyQualifiedMethodName() {
+        return getFullQualifiedMethodName(this);
+      }
+    }
+
+    enum serviceaccount implements RpcMethod {
+      get,
+      put;
```
frankyn, line 144: ServiceAccount API doesn't have a PUT operation, is this intended?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethod.java
```diff
@@ -0,0 +1,167 @@
...
+                Arrays.stream(bucket_acl.values()),
+                Arrays.stream(buckets.values()),
+                Arrays.stream(default_object_acl.values()),
+                Arrays.stream(hmacKey.values()),
+                Arrays.stream(notifications.values()),
+                Arrays.stream(object_acl.values()),
+                Arrays.stream(objects.values()),
+                Arrays.stream(serviceaccount.values()))
+            .flatMap(Function.identity()) // .flatten()
+            .collect(Collectors.toMap(RpcMethod::getFullyQualifiedMethodName, Function.identity()));
```
frankyn, line 165: Wait, why is there a double colon? Is this C++ in Java or did I miss this Java syntax in a newer Java version?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
```diff
@@ -0,0 +1,1942 @@
...
+final class RpcMethodMappings {
+  private static final Logger LOGGER = Logger.getLogger(RpcMethodMappings.class.getName());
+
+  static final int _2MiB = 2 * 1024 * 1024;
+  final Multimap<RpcMethod, RpcMethodMapping> funcMap;
+
+  RpcMethodMappings() {
+    ArrayList<RpcMethodMapping> a = new ArrayList<>();
+
+    BucketAcl.delete(a);
```
frankyn, line 111: Mapping seems complex because it requires a separate Model class to be introduced. Could you talk more to this?

frankyn (changes requested):
Hi @BenWhitehead! Could you split up this PR into multiple? 

There are design decisions in this PR that I'll definitely gloss over accidentally.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethod.java
JesseLovelace: Double colons were introduced in Java 8!

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestRetryConformance.java
frankyn: It does not support signed URLs.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMapping.java
```diff
@@ -0,0 +1,199 @@
...
+        } catch (StorageException e) {
+          // We expect an exception to be thrown by mapping and test retry conformance config
+          // Verify that the exception we received is actually what we expect.
+          boolean matchExpectedCode = false;
+          int code = e.getCode();
+          HashSet<String> instructions = newHashSet(c.getInstruction().getInstructionsList());
+          if (instructions.contains("return-503") && code == 503) {
+            matchExpectedCode = true;
+          }
+          if (instructions.contains("return-400") && code == 400) {
```
frankyn, line 100: Does this mean failures are valid even if they happen out of order?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
frankyn: This was addressed offline.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
```diff
@@ -0,0 +1,1942 @@
...
+    }
+  }
+
+  static final class Mappings {
+
+    static final class BucketAcl {
+
+      private static void delete(ArrayList<RpcMethodMapping> a) {
+        a.add(
+            RpcMethodMapping.newBuilder(1, bucket_acl.delete)
```
frankyn, line 236: How do we not use constants as an ID? I suspect we will get burned easily unless I'm missing something.

frankyn (changes requested):
a few comments, but overall LGTM, thanks for adding the layout of the retry test conf test code.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMapping.java
BenWhitehead: This code will only ever receive the last failure to occur in the code.

So in the case that we have a test which is expected to fail upon the first 400 returned by the server, this code will receive that 400.

In the case of something like `instructions: ["return-503", "return-503", "return-503", "return-503"]` this code doesn't know that the 503 it's receiving is from index 2 because the first two where expected to be handled by the retry logic, whereas the index 2 values would be due to retry exhaustion.

frankyn: Thanks for clarifying!

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
BenWhitehead: the id's don't need to be stable themselves, only unique. If any id is determined to be non-unique an error will be thrown and the suite will fail, for example:
```
java.lang.IllegalStateException: duplicate mapping ids present: [2, 5, 6]
	at com.google.cloud.storage.conformance.retry.RpcMethodMappings.validateMappingDefinitions(RpcMethodMappings.java:195)
	at com.google.cloud.storage.conformance.retry.RpcMethodMappings.<init>(RpcMethodMappings.java:166)
```

Additonaly, any ids which aren't used during test case resolution will be reported as well
```
Sep 15, 2021 1:21:46 PM com.google.cloud.storage.conformance.retry.ITRetryConformanceTest$RetryTestCaseResolver validateGeneratedTestCases
WARNING: Declared but unused mappings with ids: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

If we were to want named constants, we'd need to define (so far) 123 of them.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethod.java
BenWhitehead: I've cleaned this up.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/TestRetryConformance.java
BenWhitehead: Do you know if that also extends to accepting a signed url to start a resumable upload?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
frankyn: Thanks for the additional context, would we lose anything if you used a static counter which increments on each instant of a mapping instead?

BenWhitehead: With JUnit 4 we don't easily have a way of pointing to the mapping as the actual test definition, so the static id serves as a ctrl+f way of finding a test fairly easily. If we had access to JUnit 5 there are supposed to be ways which we could "point" to the mapping.

frankyn (approved):

JesseLovelace (approved):

PR #1017, opened 2021-09-14
chore: cleanup usage of runWithRetries

BenWhitehead:
* Cleanup and deduplicate error handling/exception translation logic
* Condense all callables to lambdas

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Retrying.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Retrying.java
@@ -18,13 +18,28 @@ package com.google.cloud.storage;
 
 import static com.google.cloud.RetryHelper.runWithRetries;
 
+import com.google.api.core.ApiClock;
+import com.google.api.gax.retrying.ResultRetryAlgorithm;
+import com.google.api.gax.retrying.RetrySettings;
 import com.google.cloud.BaseService;
 import com.google.cloud.RetryHelper.RetryHelperException;
 import java.util.concurrent.Callable;
 import java.util.function.Function;
 
-public final class Retrying {
+final class Retrying {
 
+  /**
+   * A convenience wrapper around {@link com.google.cloud.RetryHelper#runWithRetries(Callable, RetrySettings, ResultRetryAlgorithm, ApiClock)}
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Retrying.java
```diff
@@ -0,0 +1,38 @@
...
+import static com.google.cloud.RetryHelper.runWithRetries;
+
+import com.google.cloud.BaseService;
+import com.google.cloud.RetryHelper.RetryHelperException;
+import java.util.concurrent.Callable;
+import java.util.function.Function;
+
+public final class Retrying {
+
+  static <T, U> U run(StorageOptions options, Callable<T> c, Function<T, U> f) {
```
tritone, line 28: Since this is used across lots of different calls, it would be good to add some docs/examples about the templated pieces/args and how they're intended to be used.

tritone (commented):
Love how much boilerplate this approach allows us to remove! Since it's a bit complicated as far as logic, I'd like to get a review from the java folks on the team as well (@JesseLovelace / @frankyn ).

Also, can you discuss how you tested this? I'm assuming that you used the conformance tests draft to guarantee that you are avoiding regressions.

google-cloud-storage/src/main/java/com/google/cloud/storage/Retrying.java
BenWhitehead: Added some javadocs and nullability annotations to this method to help communicate the semantics of this method.

frankyn (approved):
Refactor LGTM, thanks for adding the additional javadoc. 

There are more runWithRetries uses, could you clarify scope? It would be awesome if you could file a tracking issue to point back to as you continue refactoring.

BenWhitehead:
Related to #1024

BenWhitehead:
I created #1024 to track cleaning up the remaining usages

BenWhitehead:
Related to the question of testing @tritone, this is a mainly syntactic refactor because we now have access to lambdas with java 8. Secondarily, because the lambdas are more terse, I also did the extra step of creating Retrying.run to remove the duplicated error handling, retry settings and api clock resolution that was everywhere.

tritone (approved):

Issue #1024, opened 2021-09-15
chore: cleanup usages of `runWithRetries`


BenWhitehead:
There is lots of duplication of parameter resolution, error handling/translations in the code related to `runWithRetries`. With access to lambdas we can also clean up the verbose anonymous callables everywhere.

PR #1029, opened 2021-09-20
chore(retry): introduce RetryAlgorithmManager

BenWhitehead:
### Summary
* RetryAlgorithmManager provides an abstraction point for determining
  which ExceptionHandler should be used for a specific call.
* Implement LegacyRetryAlgorithmManager for which all methods return
  BaseService.EXCEPTION_HANDLER which is hardcoded everywhere
  previously.
* Stub out NewRetryAlgorithmManager which currently extends
  LegacyRetryAlgorithmManager, but allows for overriding individual
  methods in an incremental manner.

### Refactoring
* All usages of BaseService.EXCEPTION_HANDLER in StorageImpl, Blob,
  Bucket etc now instead use RetryAlgorithmManager to resolve a
  handler per method.
* Refactor Blob to use Retrying#run instead of runWithRetries
* Split starting of a resumable upload session out of BlobWriteChannel
  to ResumableMedia. This decouple the ability to instantiate a
  BlobWriteChannel independent of performing an RPC to resolve the
  uploadId. This split is also necessary to allow differing configuration
  of ExceptionHandler for open vs write of a resumable upload session.
* Refactor StorageException#translateAndThrow to allow for an option which
  doesn't throw, new StorageException#coalesce which attempts to coalesce
  to a BaseServiceException but doesn't throw.
* Relax several easymock assertions on StorageOptions method. (Calling get
  on an immutable object shouldn't be so concerned with invocation count)

Related to #1024

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.storage;
 
-import static com.google.cloud.RetryHelper.runWithRetries;
 import static com.google.cloud.storage.Blob.BlobSourceOption.toGetOptions;
 import static com.google.cloud.storage.Blob.BlobSourceOption.toSourceOptions;
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -25,9 +24,8 @@ import static java.util.concurrent.Executors.callable;
 import com.google.api.services.storage.model.StorageObject;
 import com.google.auth.ServiceAccountSigner;
 import com.google.auth.ServiceAccountSigner.SigningException;
+import com.google.cloud.ExceptionHandler;
 import com.google.cloud.ReadChannel;
-import com.google.cloud.RetryHelper;
-import com.google.cloud.Tuple;
 import com.google.cloud.WriteChannel;
 import com.google.cloud.storage.Acl.Entity;
 import com.google.cloud.storage.Storage.BlobTargetOption;
@@ -35,7 +33,6 @@ import com.google.cloud.storage.Storage.BlobWriteOption;
 import com.google.cloud.storage.Storage.CopyRequest;
 import com.google.cloud.storage.Storage.SignUrlOption;
 import com.google.cloud.storage.spi.v1.StorageRpc;
-import com.google.common.base.Function;
 import com.google.common.io.BaseEncoding;
 import com.google.common.io.CountingOutputStream;
 import java.io.IOException;
@@ -50,6 +47,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 
 /**
  * An object in Google Cloud Storage. A {@code Blob} object includes the {@code BlobId} instance,
@@ -78,17 +76,8 @@ public class Blob extends BlobInfo {
   private static final long serialVersionUID = -6806832496717441434L;
 
   private final StorageOptions options;
+  private final RetryAlgorithmManager retryAlgorithmManager;
   private transient Storage storage;
-
-  static final Function<Tuple<Storage, StorageObject>, Blob> BLOB_FROM_PB_FUNCTION =
-      new Function<Tuple<Storage, StorageObject>, Blob>() {
-        @Override
-        public Blob apply(Tuple<Storage, StorageObject> pb) {
-          return Blob.fromPb(pb.x(), pb.y());
-        }
-      };
-
-  private static final int DEFAULT_CHUNK_SIZE = 2 * 1024 * 1024;
 
   /** Class for specifying blob source options when {@code Blob} methods are used. */
   public static class BlobSourceOption extends Option {
@@ -258,26 +247,18 @@ public class Blob extends BlobInfo {
   public void downloadTo(OutputStream outputStream, BlobSourceOption... options) {
     final CountingOutputStream countingOutputStream = new CountingOutputStream(outputStream);
     final StorageRpc storageRpc = this.options.getStorageRpcV1();
+    StorageObject pb = getBlobId().toPb();
     final Map<StorageRpc.Option, ?> requestOptions = StorageImpl.optionMap(getBlobId(), options);
-    try {
-      runWithRetries(
-          callable(
-              new Runnable() {
-                @Override
-                public void run() {
-                  storageRpc.read(
-                      getBlobId().toPb(),
-                      requestOptions,
-                      countingOutputStream.getCount(),
-                      countingOutputStream);
-                }
-              }),
-          this.options.getRetrySettings(),
-          StorageImpl.EXCEPTION_HANDLER,
-          this.options.getClock());
-    } catch (RetryHelper.RetryHelperException e) {
-      StorageException.translateAndThrow(e);
-    }
+    ExceptionHandler exceptionHandler = retryAlgorithmManager.getForObjectsGet(pb, requestOptions);
+    Retrying.run(
+        this.options,
+        exceptionHandler,
+        callable(
+            () -> {
+              storageRpc.read(
+                  pb, requestOptions, countingOutputStream.getCount(), countingOutputStream);
+            }),
+        Function.identity());
   }
 
   /**
@@ -506,6 +487,7 @@ public class Blob extends BlobInfo {
     super(infoBuilder);
     this.storage = checkNotNull(storage);
     this.options = storage.getOptions();
+    this.retryAlgorithmManager = storage.getOptions().getRetryAlgorithmManager();
   }
 
   /**
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
@@ -17,46 +17,40 @@
 package com.google.cloud.storage;
 
 import static com.google.cloud.RetryHelper.runWithRetries;
+import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.Executors.callable;
 
 import com.google.api.services.storage.model.StorageObject;
 import com.google.cloud.BaseWriteChannel;
+import com.google.cloud.ExceptionHandler;
 import com.google.cloud.RestorableState;
 import com.google.cloud.RetryHelper;
 import com.google.cloud.WriteChannel;
-import com.google.cloud.storage.spi.v1.StorageRpc;
 import java.math.BigInteger;
-import java.net.URL;
-import java.util.Map;
-import java.util.concurrent.Callable;
+import java.util.function.Supplier;
+import org.checkerframework.checker.nullness.qual.NonNull;
 
 /** Write channel implementation to upload Google Cloud Storage blobs. */
 class BlobWriteChannel extends BaseWriteChannel<StorageOptions, BlobInfo> {
 
-  BlobWriteChannel(StorageOptions options, BlobInfo blob, Map<StorageRpc.Option, ?> optionsMap) {
-    this(options, blob, open(options, blob, optionsMap));
-  }
-
-  BlobWriteChannel(StorageOptions options, URL signedURL) {
-    this(options, open(signedURL, options));
-  }
-
-  BlobWriteChannel(StorageOptions options, BlobInfo blobInfo, String uploadId) {
-    super(options, blobInfo, uploadId);
-  }
-
-  BlobWriteChannel(StorageOptions options, String uploadId) {
-    super(options, null, uploadId);
-  }
-
-  // Contains metadata of the updated object or null if upload is not completed.
-  private StorageObject storageObject;
-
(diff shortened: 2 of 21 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/RetryAlgorithmManager.java
```diff
@@ -0,0 +1,118 @@
...
+import com.google.api.services.storage.model.Policy;
+import com.google.api.services.storage.model.StorageObject;
+import com.google.cloud.ExceptionHandler;
+import com.google.cloud.storage.spi.v1.StorageRpc;
+import com.google.cloud.storage.spi.v1.StorageRpc.RewriteRequest;
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+interface RetryAlgorithmManager extends Serializable {
```
frankyn, line 32: Why not abstract instead of interface, given additional API operations may be added in the future?

frankyn (commented):
Still reading through the PR, have one leading question on RetryAlgorithmManager definition

google-cloud-storage/src/main/java/com/google/cloud/storage/RetryAlgorithmManager.java
BenWhitehead: Abstract classes aren't mockable without some additional hackery necessary, and a class can only extend one class but can implement multiple interfaces. Theoretically someone might want to provide an implementation which extends their own internal class while also implementing this interface.

We can provide a Delegating implementation which can ensure not needing to re-implement the whole interface to override a single method, while keeping the public usage api against an interface.

frankyn: Naive question: It's a breaking change if a consumer chooses to implement the interface and we add an additional method which would require them to implement the method. We've hit conversation before with Storage interface and the recommendation at that time was to use class's.

BenWhitehead: That's fair, my intent with the interface/delegate approach was to provide the interface always, with the convenience of extending a class. We don't want people to directly extend our internal implementations as that binds those implementations into the public api preventing us from making any modifications we might need to down the road.

Theoretically, we could provide default implementations in the interface now that we have java 8, but I'm not totally sure we want to do that.

Side note, this interface is currently package private and not part of the public api. If we want we can poke ahead with the interface and see how it goes before deciding to expose it publicly.

google-cloud-storage/src/test/java/com/google/cloud/storage/ResumableMediaTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import static org.junit.Assert.assertNotNull;
+
+import com.google.cloud.BaseService;
+import com.google.cloud.ExceptionHandler;
+import java.net.URL;
+import org.junit.Assert;
+import org.junit.Test;
+
+public final class ResumableMediaTest {
```
tritone, line 27: This test seems orthogonal to the other changes?

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
```diff
@@ -17,47 +17,41 @@
...
-import java.net.URL;
-import java.util.Map;
-import java.util.concurrent.Callable;
+import java.util.function.Supplier;
+import org.checkerframework.checker.nullness.qual.NonNull;
 
 /** Write channel implementation to upload Google Cloud Storage blobs. */
 class BlobWriteChannel extends BaseWriteChannel<StorageOptions, BlobInfo> {
 
-  BlobWriteChannel(StorageOptions options, BlobInfo blob, Map<StorageRpc.Option, ?> optionsMap) {
```
tritone, line 36: How are we able to remove these methods?

tritone (commented):
Have a few questions; I'm going to lean on @frankyn and @JesseLovelace to give this a deeper review.

google-cloud-storage/src/test/java/com/google/cloud/storage/ResumableMediaTest.java
BenWhitehead: This is a related move after separating the open of a resumable session from the BlobWriteChannel. Because there are differing retry handling needs for open vs write on a resumable session and because BlobWriteChannel implements Restorable, I separated open out into ResumableMedia helper class and this tests moved here from BlobWriteChannelTest.

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
BenWhitehead: Because there are differing retry handling needs for open vs write on a resumable session and because BlobWriteChannel implements Restorable, I separated open out into ResumableMedia helper class which allowed for the removal of these overloaded constructors.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -258,26 +247,18 @@ public void downloadTo(Path path, BlobSourceOption... options) {
...
-                      countingOutputStream);
-                }
-              }),
-          this.options.getRetrySettings(),
-          StorageImpl.EXCEPTION_HANDLER,
-          this.options.getClock());
-    } catch (RetryHelper.RetryHelperException e) {
-      StorageException.translateAndThrow(e);
-    }
+    ExceptionHandler exceptionHandler = retryAlgorithmManager.getForObjectsGet(pb, requestOptions);
```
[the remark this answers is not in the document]
BenWhitehead: Compared to python we don't have the pattern of the retry algorithm provided as a method argument, this approach allows for per rpc method handling. In our default implementation we'll most likely end up with common `ExceptionHandler`s which can simply be returned when applicable.

Part of this api is also functioning as a bit of a prototype of what we might want the public api to look like. In java-core there is a pattern of being able to configure retry settings per rpc method as there is not necessarily the idempotency buckets available universally.

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
tritone: Gotcha, just wanted to make sure that this isn't something affecting the public surface of the library!

google-cloud-storage/src/test/java/com/google/cloud/storage/ResumableMediaTest.java
tritone: Ahh okay got it, I didn't see where this was moved from.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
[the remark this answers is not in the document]
tritone: Sounds good. So do you think we'll be able to get more conciseness in the public API? Or probably not?

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
BenWhitehead: No worries, that's a totally valid concern. These constructors (and even this whole class) is package private so isn't part of the public api today.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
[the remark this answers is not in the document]
BenWhitehead: Possibly, we'll have to weigh it against how gax-java allows configuration of retry settings as I know there is some ability for per rpc config in their public api.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -504,12 +562,33 @@ public BlobWriteChannel writer(BlobInfo blobInfo, BlobWriteOption... options) {
 
   @Override
   public BlobWriteChannel writer(URL signedURL) {
-    return new BlobWriteChannel(getOptions(), signedURL);
+    ExceptionHandler forResumableUploadSessionCreate =
+        retryAlgorithmManager.getForResumableUploadSessionCreate(
+            Collections
+                .emptyMap()); // TODO: is it possible to know if a signed url is configured to have
+    // a constraint which makes it idempotent?
```
frankyn, line 569: o, this is a really good call out. You can in theory generate many sessions with one signed URL IIRC. As far as making it idempotent, yes, you can use precondition headers for XML API: https://cloud.google.com/storage/docs/xml-api/put-object#request-headers

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -476,7 +531,10 @@ public CopyWriter copy(final CopyRequest copyRequest) {
   public byte[] readAllBytes(BlobId blob, BlobSourceOption... options) {
     final StorageObject storageObject = blob.toPb();
     final Map<StorageRpc.Option, ?> optionsMap = optionMap(blob, options);
-    return run(() -> storageRpc.load(storageObject, optionsMap), Function.identity());
+    ExceptionHandler exceptionHandler =
+        retryAlgorithmManager.getForObjectsGet(storageObject, optionsMap);
+    return run(
```
frankyn, line 536: Question on design, I left a comment on the design document. I'm wondering if we can reduce complexity here a little bit for future us.

frankyn (changes requested):
Overall this LGTM, I have a few nits/questions. I'm really curious about the design before approving.

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
```diff
@@ -2148,16 +2146,6 @@ public void testRuntimeException() {
     }
   }
 
-  @Test
-  public void testWriterWithSignedURL() throws MalformedURLException {
```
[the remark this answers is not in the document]
frankyn: Ignore. It got refactored missed that.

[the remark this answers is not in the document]
BenWhitehead: The actual testing guts were moved to ResumableMediaTest.java and given a more descriptive name here: https://github.com/googleapis/java-storage/pull/1029/files#diff-3d1e0f94d16d08b59c904d3e772743afad8827221342f4d2a13624fcec1ab0ccR34-R45

Signed URL handling is only necessary for the creation of the session but has no bearing on the operation of the BlobWriteChannel itself.

google-cloud-storage/src/test/java/com/google/cloud/storage/BlobTest.java
```diff
@@ -341,12 +345,12 @@ public void testCopyTo() throws Exception {
...
   @Test
   public void testCopyToBlobId() throws Exception {
-    initializeExpectedBlob(2);
+    initializeExpectedBlob();
     BlobInfo target = BlobInfo.newBuilder(BlobId.of("bt", "nt")).build();
     BlobId targetId = BlobId.of("bt", "nt");
     CopyWriter copyWriter = createMock(CopyWriter.class);
     Capture<CopyRequest> capturedCopyRequest = Capture.newInstance();
-    expect(storage.getOptions()).andReturn(mockOptions);
+    expect(storage.getOptions()).andReturn(mockOptions).anyTimes();
```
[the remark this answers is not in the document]
frankyn: Ben in description of PR: 

```
Relax several easymock assertions on StorageOptions method. (Calling get
on an immutable object shouldn't be so concerned with invocation count)
```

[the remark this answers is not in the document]
BenWhitehead: With the introduction of the RetryAlgorithmManager to StorageOptions, the number of times there are interactions with storage to get StorageOptions, or calls against an instance of StorageOptions increased. As far as test maintenence is concerned N calls to a method on an immutable object and produces an immutable result aren't interesting from a number of invocations stand point. So, rather than fighting with the numbers being unexpected, I opted to remove the constraint on the number of times the method is allowed to be called.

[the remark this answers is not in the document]
frankyn: Makes perfect sense, thanks for clarifying the change.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
frankyn: Update: @BenWhitehead and I spoke on Friday through the current design and decided to move forward with the existing design for the following reasons:

1. It's not explicitly exposed as a Public API (yet).
2. It refactors the complexity of Resumable Media operations to map back to gRPC API spec which helps clean up internal maintanence
3. We would like to make progress towards Retry Strategy alignment while acknowledging there's not enough information on gotchas from introducing gRPC Retries yet. We are making decisions with the information we have right now.

frankyn (approved):

PR #1040, opened 2021-09-22
test: retry conformance tests fixes, round 2

BenWhitehead:
* test: fix setup for hmacKey create
* test: fix setup for bucket patch
* test: cleanup bucket.lockRetentionPolicy mapping 
  * Add precondition predicate to mapping with IfMetagenerationMatch and remove mapping with no guard. Metageneration is a required argument.
* test: fix mappings for ReadChannel
  * ReadChannel wrap any exceptions in an IOException, and doesn't extract the cause from the RetryHelperException
* test: fix mappings hmacKey.delete
  * As part of setup for delete we need to set the key to inactive
* test: individualMapping(int) -> specificMappings(int...)

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
@@ -97,10 +97,10 @@ final class CtxFunctions {
           Blob resolvedBlob = ctx.getStorage().create(blobInfo);
           return ctx.map(s -> s.with(resolvedBlob));
         };
-    private static final CtxFunction serviceAccount =
+    static final CtxFunction serviceAccount =
         (ctx, c) ->
             ctx.map(s -> s.with(ServiceAccount.of(c.getServiceAccountSigner().getAccount())));
-    private static final CtxFunction hmacKey =
+    static final CtxFunction hmacKey =
         (ctx, c) -> ctx.map(s -> s.with(ctx.getStorage().createHmacKey(s.getServiceAccount())));
 
     private static final CtxFunction processResources =
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
@@ -27,6 +27,7 @@ import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.HmacKey;
 import com.google.cloud.storage.ServiceAccount;
 import com.google.cloud.storage.conformance.retry.Functions.CtxFunction;
 import com.google.common.base.Joiner;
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
```diff
@@ -97,11 +98,16 @@
...
+    static final CtxFunction serviceAccount =
         (ctx, c) ->
             ctx.map(s -> s.with(ServiceAccount.of(c.getServiceAccountSigner().getAccount())));
-    private static final CtxFunction hmacKey =
-        (ctx, c) -> ctx.map(s -> s.with(ctx.getStorage().createHmacKey(s.getServiceAccount())));
+    static final CtxFunction hmacKey =
+        (ctx, c) ->
+            ctx.map(
+                s -> {
+                  HmacKey hmacKey1 = ctx.getStorage().createHmacKey(s.getServiceAccount());
```
tritone, line 108: Not a big deal either way but generally the service account returned by `getServiceAccount` is the default GCS service account and not one you would use to create HMAC keys-- that said it doesn't matter either way to the emulator and you could even use a dummy email and it would be fine.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
```diff
@@ -1130,11 +1140,21 @@ private static void get(ArrayList<RpcMethodMapping> a) {
...
-                                  WritableByteChannel write =
-                                      Channels.newChannel(NullOutputStream.INSTANCE);
-                                  ByteStreams.copy(reader, write);
+                                  try {
+                                    ReadChannel reader =
+                                        ctx.getStorage().reader(ctx.getState().getBlobId());
+                                    WritableByteChannel write =
+                                        Channels.newChannel(NullOutputStream.INSTANCE);
+                                    ByteStreams.copy(reader, write);
+                                  } catch (IOException e) {
```
tritone, line 1149: Why does this end up being necessary?

tritone (approved):
Generally looks good!

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
BenWhitehead: In this case, the `getServiceAccount()` is against the test state, so it's a service account that was built as part of setup based on a canned email address.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
BenWhitehead: BlobReadChannel wraps any retry helper exception in an IOException
https://github.com/googleapis/java-storage/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobReadChannel.java#L140-L142

We can't change the exception hierarchy without the possibility of breaking customers error handling code, so I opted to unwrap here in the test to get to the actual StorageException/Rpc Exception.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
tritone: Ah, got it!

PR #1051, opened 2021-09-28
fix: update PAP to use inherited instead of unspecified

BenWhitehead:


frankyn (approved):
LGTM

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -113,13 +113,19 @@ public BucketInfo apply(com.google.api.services.storage.model.Bucket pb) {
...
     ENFORCED("enforced"),
-    /** Default value for Public Access Prevention */
-    UNSPECIFIED("unspecified"),
+    /**
+     * Default value for Public Access Prevention
+     *
+     * @deprecated use {@link #INHERITED}
+     */
+    @Deprecated
+    UNSPECIFIED("inherited"),
```
frankyn, line 122: @shaffeeullah will the service still support unspecified to continue supporting clients that are released?

PR #1098, opened 2021-10-08
chore(retry): cleanup RpcMethodMappings and associated setup

BenWhitehead:
* Cleanup left over todo's that have been addressed
* Ensure all setup operations are taking place during setup and not in the test
* Update CtxFunctions to narrow scope of package public functions
* Update object setup to also set BlobInfo and BlobId

*note* because several setup steps from from test to setup the formatting changed indentation and made this PR a bit noisy. The actual substance of changes is in the first commit, and probably easier to filter down to that for review.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
@@ -61,7 +61,8 @@ final class CtxFunctions {
     /**
      * Populate a bucket info for the state present in the ctx.
      *
-     * this is primarily useful in the case when you want to insert a bucket during the test
+     * <p>this is primarily useful in the case when you want to insert a bucket during the test
+     *
      * @see State#getBucketInfo()
      */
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
```diff
@@ -123,8 +107,12 @@
         (ctx, c) -> {
           BlobInfo blobInfo =
               BlobInfo.newBuilder(ctx.getState().getBucket().getName(), c.getObjectName()).build();
-          Blob resolvedBlob = ctx.getStorage().create(blobInfo);
-          return ctx.map(s -> s.with(resolvedBlob));
+          Blob resolvedBlob = ctx.getStorage().create(blobInfo, c.getHelloWorldUtf8Bytes());
```
frankyn, line 110: Not necessary right now, but could you add comments as to why each object is necessary? Right now I'm not sure why I would use one over the other.

frankyn (approved):

PR #1105, opened 2021-10-13
feat: Add RPO metadata settings

JesseLovelace:
Adds RPO metadata settings for buckets.  

Holding off on adding documentation until this general implementation is approved 

See https://github.com/googleapis/google-cloud-java/pull/7700


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -54,34 +54,16 @@
 import com.google.cloud.kms.v1.KeyManagementServiceGrpc.KeyManagementServiceBlockingStub;
 import com.google.cloud.kms.v1.KeyRingName;
 import com.google.cloud.kms.v1.LocationName;
-import com.google.cloud.storage.Acl;
+import com.google.cloud.storage.*;
```
frankyn, line 57: Revert this change

google-cloud-storage/src/main/java/com/google/cloud/storage/Rpo.java
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.storage;
+
+import com.google.api.core.ApiFunction;
+import com.google.cloud.StringEnumType;
+import com.google.cloud.StringEnumValue;
+
+public class Rpo extends StringEnumValue {
```
frankyn, line 22: It hurts me how unreadable `Rpo` is. @ddelgrosso1 for other languages, ya'll are keeping Rpo naming, is that accurate?

google-cloud-storage/clirr-ignored-differences.xml
```diff
@@ -21,6 +21,11 @@
         <method>com.google.cloud.storage.BucketInfo$Builder deleteLifecycleRules()</method>
         <differenceType>7013</differenceType>
     </difference>
+    <difference>
```
frankyn, line 24: @BenWhitehead we hit this each time we make a change to a public interface, we've been told it's to protect us from breaking changes, however we are committing breaking changes. It has gone on for years. What are your thoughts on this?

frankyn (changes requested):

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
frankyn: https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports

google-cloud-storage/clirr-ignored-differences.xml
BenWhitehead: Clirr is in place to primarily help us know if we are committing a change that breaks binary compatibility. The place it's most beneficial is on our generated code and our "main" apis. In this case, it's less beneficial and unfortunately BucketInfo.Builder is abstract and not annotated with [`@InternalExtensionOnly`](https://github.com/googleapis/api-common-java/blob/v2.0.5/src/main/java/com/google/api/core/InternalExtensionOnly.java#L69) which would give us leeway to add methods with no guarantee of binary compatibility for out of project implementations. However, adding this annotation is itself considered a breaking change and would need to be performed at a major version.

Our policy for model classes in general has been that we consider them to generally be safe for adding new methods, unfortunately clirr isn't capable of being made aware that this is a model class and instead requires an entry defined. Unfortunately, at this point, I think we're sort of stuck having to deal with this file until either we reach a new major version when we can annotate this builder and define a general rule or we decide to refactor things and make this class not be abstract and instead simply non-final.

PR #1132, opened 2021-11-01
feat: update all automatic retry behavior to be idempotency aware

BenWhitehead:
1. Add new configuration option `StorageOptions.Builder#setStorageRetryStrategy`
2. Add new interface `StorageRetryStrategy`
3. Update all calls to be made idempotency aware
4. Change default behavior of automatic retries to only retry those operations which are deemed to be idempotent
5. Add @deprecated method `StorageRetryStrategy.getLegacyStorageRetryStrategy` providing access to a strategy which matches the previous sometimes unsafe behavior


frankyn (approved):
:ship: it!

tritone (approved):
Can I just say how delightful it is that the PR to actually change this behavior ended up being <10 lines?

BenWhitehead:
Hehe, yeah the 10 lines felt great. Then the tests not setting IfGenerationMatch reared their head :(

PR #1160, opened 2021-12-01
fix: Remove all client side validation for OLM, allow nonspecific lif…

JesseLovelace:
Removes client side validation by refactoring `LifecycleAction` into a non-abstract class and allowing for nonspecific lifecycle actions.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -548,8 +548,12 @@ public class BucketInfo implements Serializable {
                   StorageClass.valueOf(action.getStorageClass()));
           break;
         default:
-          throw new UnsupportedOperationException(
-              "The specified lifecycle action " + action.getType() + " is not currently supported");
+          log.warning(
+              "Creating a lifecycle rule with no supported actions: "
+                  + rule
+                  + ". Attempting to update with this rule may cause errors. Please "
+                  + "update to the latest version of google-cloud-storage.");
+          lifecycleAction = LifecycleAction.newLifecycleAction(action.getType());
       }
 
       Rule.Condition condition = rule.getCondition();
@@ -799,13 +803,25 @@ public class BucketInfo implements Serializable {
     }
 
     /**
-     * Base class for the Action to take when a Lifecycle Condition is met. Specific Actions are
+     * Base class for the Action to take when a Lifecycle Condition is met. Supported Actions are
      * expressed as subclasses of this class, accessed by static factory methods.
      */
-    public abstract static class LifecycleAction implements Serializable {
+    public static class LifecycleAction implements Serializable {
       private static final long serialVersionUID = 5801228724709173284L;
 
-      public abstract String getActionType();
+      private final String actionType;
+
+      public String getActionType() {
+        return actionType;
+      }
+
+      private LifecycleAction(String actionType) {
+        this.actionType = actionType;
+      }
+
+      public static LifecycleAction newLifecycleAction(String actionType) {
+        return new LifecycleAction(actionType);
+      }
 
       @Override
       public String toString() {
@@ -836,11 +852,8 @@ public class BucketInfo implements Serializable {
       public static final String TYPE = "Delete";
       private static final long serialVersionUID = -2050986302222644873L;
 
-      private DeleteLifecycleAction() {}
-
-      @Override
-      public String getActionType() {
-        return TYPE;
+      private DeleteLifecycleAction() {
+        super(TYPE);
       }
     }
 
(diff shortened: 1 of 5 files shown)


frankyn (changes requested):
Thanks Jesse

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -548,8 +548,12 @@ static LifecycleRule fromPb(Rule rule) {
                   StorageClass.valueOf(action.getStorageClass()));
           break;
         default:
-          throw new UnsupportedOperationException(
-              "The specified lifecycle action " + action.getType() + " is not currently supported");
+          log.warning(
```
danielduhh, line 551: Can we add tests to ensure unknown types no longer throw exceptions?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -799,13 +803,25 @@ public LifecycleCondition build() {
     }
 
     /**
-     * Base class for the Action to take when a Lifecycle Condition is met. Specific Actions are
+     * Base class for the Action to take when a Lifecycle Condition is met. Supported Actions are
      * expressed as subclasses of this class, accessed by static factory methods.
      */
-    public abstract static class LifecycleAction implements Serializable {
+    public static class LifecycleAction implements Serializable {
```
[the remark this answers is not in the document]
JesseLovelace: Abstract classes can't be instantiated so we do if we want to allow for unsupported action types to be created in this way.

The alternative would be to make an implementation of LifecycleAction for that purpose, like "CustomLifecycleAction" or something, but I'm worried that doing that would be confusing, or imply that this is the correct way to use action types that aren't in the library (when in actuality users should upgrade to the latest version of the library)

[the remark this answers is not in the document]
frankyn: Ah, totally missed that. Thanks Jesse, I would confirm what happens when a say setStorageClass doesn't have storageClass in the rule (the case an action type is added that requires additional parameters) and see how GCS handles the request on a PATCH request. 

I suspect it will fail; so we may want to clarify if we want the library *not* to create an instance for that unknown action type.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1007,7 +1017,11 @@ void populateCondition(Rule.Condition condition) {
 
     @Override
     void populateCondition(Rule.Condition condition) {
-      throw new UnsupportedOperationException();
+      log.warning(
```
BenWhitehead, line 1020: This change doesn't appear to be exercised by the updated test. (Ran the test with code coverage and it didn't hit this line.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -799,13 +804,25 @@ public LifecycleCondition build() {
...
       private static final long serialVersionUID = 5801228724709173284L;
 
-      public abstract String getActionType();
+      private final String actionType;
+
+      public String getActionType() {
+        return actionType;
+      }
+
+      private LifecycleAction(String actionType) {
```
BenWhitehead, line 819: nit: move constructor before public methods, after private fields.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -799,13 +804,25 @@ public LifecycleCondition build() {
...
+        this.actionType = actionType;
+      }
+
+      public static LifecycleAction newLifecycleAction(String actionType) {
```
BenWhitehead, line 823: nit: move static method down with other static methods.

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
JesseLovelace: this change is for the old, deprecated version of lifecycle rules, but i can add a test for that if you think it's necessary

BenWhitehead (approved):

frankyn (approved):

frankyn:
@JesseLovelace are you still addressing the clirr issue?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -799,13 +804,21 @@ public LifecycleCondition build() {
...
      * expressed as subclasses of this class, accessed by static factory methods.
      */
-    public abstract static class LifecycleAction implements Serializable {
+    public static class LifecycleAction implements Serializable {
       private static final long serialVersionUID = 5801228724709173284L;
 
-      public abstract String getActionType();
+      private final String actionType;
+
+      private LifecycleAction(String actionType) {
```
BenWhitehead, line 815: Let's make this constructor `public` to allow someone (as remote a chance as it is) that is already extending themselves allow them to continue to extend, but now they would need to provide the `actionType`.

danielduhh (approved):

PR #1177, opened 2021-12-14
fix: do not cause a failure when encountering no bindings

### LINKED ISSUE #1175: Verify if IAM policy validation is needed when received from API
[IllegalArgumentException in PostPolicyV4.java](https://github.com/googleapis/java-storage/blob/main/google-cloud-storage/src/main/java/com/google/cloud/storage/PostPolicyV4.java#L119) when calling validateFields; it checks to make sure provided input field keys are part of a known list. This is restrictive on input and would require an update if we introduce more fields in the future (which is possible).

Follow-up:
- [ ] Does the server return a useful error?
- [ ] Is the error state opaque such that the server can't return a better error?
- [ ] If the server can be improved how can we advocate that?
- [ ] If the server error can only be opaque, do we want to introduce a strict/relaxed validation option? strict: throw exception relaxed: log a warning for customer

frankyn:

Fixes #1175  ☕️


BenWhitehead (approved):
The code changes look good to me. For the javadoc updates, we might want to get someone from the writing team to take a look and see if the voice we're using make sense.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3540,6 +3540,10 @@ HmacKeyMetadata updateHmacKeyState(
   /**
    * Gets the IAM policy for the provided bucket.
    *
+   * <p>It's possible for bindings to be empty and instead have permissions inherited through
+   * Project or Organization IAM Policies. To prevent corrupting policies when performing a {@code
```
cbonnie, line 3544: "To prevent corrupting policies when you update an IAM policy with {@code Storage.setIamPolicy, the ETAG value..."

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -3556,6 +3560,9 @@ HmacKeyMetadata updateHmacKeyState(
   /**
    * Updates the IAM policy on the specified bucket.
    *
+   * <p>To prevent corrupting policies when performing a {@code Storage.setIamPolicy}, the ETAG
```
cbonnie, line 3563: Same comment as above

cbonnie (approved):

PR #1180, opened 2021-12-16
feat: allow limiting ReadChannel

BenWhitehead:
The ReadChannel returned from storage.reader and blob.reader now also implemented LimitableReadChannel which allows limiting the max number of bytes the channel will make available for read.

This can be used in conjunction with seek to allow for range gets of objects independent of any buffer or chunk sizes.

See the integration tests for an example of how this could be used.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobReadChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobReadChannel.java
@@ -123,8 +123,8 @@ class BlobReadChannel implements ReadChannel {
       if (endOfStream) {
         return -1;
       }
-      final int toRead =
-          Math.min(Math.toIntExact(limit - position), Math.max(byteBuffer.remaining(), chunkSize));
+      final int toRead = Math.toIntExact(
+          Math.min(limit - position, Math.max(byteBuffer.remaining(), chunkSize)));
       try {
         ResultRetryAlgorithm<?> algorithm =
             retryAlgorithmManager.getForObjectsGet(storageObject, requestOptions);
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobReadChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobReadChannel.java
@@ -18,6 +18,7 @@ package com.google.cloud.storage;
 
 import static com.google.cloud.RetryHelper.runWithRetries;
 
+import com.google.api.client.util.Preconditions;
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobReadChannel.java
```diff
@@ -158,6 +164,13 @@ public int read(ByteBuffer byteBuffer) throws IOException {
     return toWrite;
   }
 
+  @Override
+  public LimitableReadChannel limit(long limit) {
```
frankyn, line 168: I was thinking you could overload read() but that interface comes from ReadableByteChannel.

What about introducing a new option into requestOptions to define range?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBlobReadChannelTest.java
```diff
@@ -0,0 +1,108 @@
...
+    ByteBuffer expectedSubContent = content.duplicate();
+    expectedSubContent.position(rangeBegin);
+    expectedSubContent.limit(rangeEnd);
+    try (WriteChannel writer = s.writer(src)) {
+      writer.write(content);
+    }
+
+    ByteBuffer actual = ByteBuffer.allocate(rangeEnd - rangeBegin);
+
+    try (ReadChannel reader = s.reader(src.getBlobId())) {
```
BenWhitehead, line 95: Investigate if it's possible (binary & source compatibility) to directly return, instead of down-casting.

frankyn: The alternative is to expose an alias that returns the new LimitableReadChannel class type, but based on discussion we didn't want to increase the surface of the API in this library so easily; which is fair.

BenWhitehead: limit feature added to ReadChannel, no need for changes here.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBlobReadChannelTest.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2021 Google LLC
```
frankyn, line 2: 2022.

frankyn (commented):
One nit, otherwise LGTM, amazing work Ben!!

frankyn (approved):

PR #1236, opened 2022-02-10
feat: Adding Samples for Printing Bucket ACLs and Printing Bucket ACL for a specific user.

sydney-munro:
Internal issue b/215600897


samples/snippets/src/test/java/com/example/storage/bucket/PrintBucketAclFilterByUserTest.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import com.google.cloud.storage.testing.RemoteStorageHelper;
+import com.google.cloud.testing.junit4.StdOutCaptureRule;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class PrintBucketAclFilterByUserTest {
```
sydney-munro, line 19: So I have some thoughts around testing in this package. Currently there are individual tests per sample. Although given there are only 2 other samples we are in a position to change this.

1) In https://github.com/googleapis/google-cloud-java tests for multiple samples are in a single file. We could do this here as well.

2) We could create a TestBase which we extend in each of our tests to eliminate some repeated setup, teardown etc code and keep doing separate tests for each sample.

3) Continue with individual tests without a common base since its not too much repeat.

I am slightly partial to creating a test base but am not sure if its worth it since we are only repeating a few lines. Let me know any thoughts on this.

BenWhitehead: I'm generally against mega test classes, too much gets intertwined in the setup and tear down over time.

I think option 2 seems good in this scenario since all the test follow a similar pattern ("Do something with a bucket, check std out for some expected tokens")

samples/snippets/src/test/java/com/example/storage/bucket/PrintBucketAclFilterByUserTest.java
```diff
@@ -0,0 +1,64 @@
...
+  @Before
+  public void setUp() {
+    bucketName = RemoteStorageHelper.generateBucketName();
+    storage = StorageOptions.getDefaultInstance().getService();
+    storage.create(BucketInfo.of(bucketName));
+  }
+
+  @After
+  public void tearDown() {
+    storage.delete(bucketName);
```
BenWhitehead, line 53: Here, we can actually leverage `RemoteStorageHelper.forceDelete(storage, bucketName)` do recursively cleanup the bucket for us, including any blobs which might be left over.

samples/snippets/src/test/java/com/example/storage/bucket/PrintBucketAclFilterByUserTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class PrintBucketAclFilterByUserTest {
+
+  @Rule public StdOutCaptureRule stdOut = new StdOutCaptureRule();
+  private static final String USER_EMAIL =
+      "google-cloud-java-tests@" + "java-docs-samples-tests.iam.gserviceaccount.com";
```
BenWhitehead, line 39: We probably want to parameterize this as an environment variable, similar to what is done for the integration tests.

See https://github.com/googleapis/java-storage/blob/9266cefe2b522493cbf2e940ff59baf8c8a0c426/.kokoro/presubmit/integration.cfg#L30-L33 for how they get passed in. And https://github.com/googleapis/java-storage/blob/9266cefe2b522493cbf2e940ff59baf8c8a0c426/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java#L2433-L2434 for how it's loaded/resolved.

BenWhitehead (changes requested):

BenWhitehead:
Owlbot is a bit aggressive and is going to clobber any changes you make to the templates. In order to allow us to change the template, we'll need to add an ignore similar to this one https://github.com/googleapis/java-storage/blob/923ea5b3b86712a10f5185a08410f3503dc144c1/owlbot.py#L33

sydney-munro:
The yoshi team has a job config that specifies which environment variables we are allowed to declare, USER_EMAIL was not one of them. I have changed it to the preferred naming which is IT_SERVICE_ACCOUNT_EMAIL which I think is not as clear but fine.

.kokoro/presubmit/samples.cfg
```diff
@@ -30,4 +30,9 @@ env_vars: {
 env_vars: {
   key: "SECRET_MANAGER_KEYS"
   value: "java-docs-samples-service-account"
-}
\ No newline at end of file
+}
+
+env_vars: {
+  key: "IT_SERVICE_ACCOUNT_EMAIL"
+  value: "google-cloud-java-tests@java-docs-samples-tests.iam.gserviceaccount.com"
```
BenWhitehead, line 37:
```suggestion
  value: "java-docs-samples-testing@java-docs-samples-testing.iam.gserviceaccount.com"
```

BenWhitehead (approved):

PR #1258, opened 2022-02-16
docs: Copy all storage samples from java-docs-samples

JesseLovelace:
Internal bug b/190102205.

Makes a copy of all (published) samples from java-docs-samples. Leaves behind everything that isn't published. 

After this is merged, all docs pages will be updated to refer to these instead of the samples in java-docs-samples. Then, the java-docs-samples samples will be deleted. 

This will allow anyone developing new library features to include samples in the same PR


BenWhitehead:
I changed the PR type to docs so it doesn't cause a minor version bump.

BenWhitehead:
In the ultimate commit comment I think it'd be good to include a link to the SHA that the samples were imported from. Just in case we ever need to try and backtrace something.

.kokoro/presubmit/integration.cfg
```diff
@@ -27,11 +27,6 @@ env_vars: {
   value: "secret_manager/java-it-service-account"
 }
 
-env_vars: {
```
frankyn, line 30: was this intentional?

frankyn (approved):
Thanks for moving over the samples!

.kokoro/presubmit/integration.cfg
JesseLovelace: owlbot did this, i didn't touch the this file at all. it's in one of the "updates from owlbot" commits

JesseLovelace: i think it's because of a mistake i made with the owlbot.py file at one point, i can fix it

PR #1288, opened 2022-03-04
docs: Adding Samples for printing all Acls for a file and for a specific user

sydney-munro:
Fixes b/222521311 ☕️
Fixes b/222521357 ☕️


samples/snippets/src/main/java/com/example/storage/object/PrintFileAcl.java
```diff
@@ -0,0 +1,56 @@
...
+
+      // This will give you the role.
+      // See https://cloud.google.com/storage/docs/access-control/lists#permissions
+      String role = acl.getRole().name();
+
+      // This will give you the Entity type (i.e. User, Group, Project etc.)
+      // See https://cloud.google.com/storage/docs/access-control/lists#scopes
+      String entityType = acl.getEntity().getType().name();
+
+      System.out.printf("%s: %s \n", role, entityType);
```
BenWhitehead, line 52: `%n` can be used to output the system dependent line endings similar to what `println()` does.
```suggestion
      System.out.printf("%s: %s %n", role, entityType);
```

samples/snippets/src/main/java/com/example/storage/object/PrintFileAclForUser.java
```diff
@@ -0,0 +1,49 @@
...
+
+    // The email of the user whose acl is being retrieved.
+    // String userEmail = "someuser@domain.com"
+
+    Storage storage = StorageOptions.newBuilder().build().getService();
+    Blob blob = storage.get(BlobId.of(bucketName, blobName));
+    Acl blobAcl = blob.getAcl(new User(userEmail));
+    String userRole = blobAcl.getRole().name();
+
+    System.out.println("User " + userEmail + " has role " + userRole);
```
BenWhitehead, line 46: I think we want to check that `blobAcl` isn't null before we use it.

If the specified user doesn't have an `Acl`, the result of `getAcl()`[1] will be `null`.

[1] [Java docs](https://cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.Blob#com_google_cloud_storage_Blob_getAcl_com_google_cloud_storage_Acl_Entity_) for `getAcl()`

BenWhitehead (changes requested):

BenWhitehead (approved):

PR #1305, opened 2022-03-23
fix(java): add configurations for Storage tests

mpeddada1:
This PR adds a few reflection configurations for native image support in Storage.

`prox-config.json`:
```
JUnit Vintage:ITBlobWriteChannelTest:testJsonEOF_0B
    MethodSource [className = 'com.google.cloud.storage.it.ITBlobWriteChannelTest', methodName = 'testJsonEOF_0B', methodParameterTypes = '']
    => com.oracle.svm.core.jdk.UnsupportedFeatureError: Proxy class defined by interfaces [interface com.google.cloud.storage.spi.v1.StorageRpc] not found. Generating proxy classes at runtime is not supported. Proxy classes need to be defined at image build time by specifying the list of interfaces that they implement. To define proxy classes use -H:DynamicProxyConfigurationFiles=<comma-separated-config-files> and -H:DynamicProxyConfigurationResources=<comma-separated-config-resources> options.
       com.oracle.svm.core.util.VMError.unsupportedFeature(VMError.java:87)
       com.oracle.svm.reflect.proxy.DynamicProxySupport.getProxyClass(DynamicProxySupport.java:146)
       java.lang.reflect.Proxy.getProxyConstructor(Proxy.java:66)
       java.lang.reflect.Proxy.newProxyInstance(Proxy.java:1006)
       com.google.common.reflect.Reflection.newProxy(Reflection.java:86)
       com.google.cloud.storage.it.ITBlobWriteChannelTest.lambda$doJsonUnexpectedEOFTest$1(ITBlobWriteChannelTest.java:140)
       com.google.cloud.ServiceOptions.getRpc(ServiceOptions.java:560)
       com.google.cloud.storage.StorageOptions.getStorageRpcV1(StorageOptions.java:149)
       com.google.cloud.storage.StorageImpl.<init>(StorageImpl.java:115)
       com.google.cloud.storage.StorageOptions$DefaultStorageFactory.create(StorageOptions.java:50)
```

The `reflect-config.json` file registers some Java classes for reflection:
```
 JUnit Vintage:ITStorageTest:testUpdateBlobReplaceMetadata
    MethodSource [className = 'com.google.cloud.storage.it.ITStorageTest', methodName = 'testUpdateBlobReplaceMetadata', methodParameterTypes = '']
    => java.lang.IllegalArgumentException: unable to create new instance of class com.google.cloud.storage.BlobInfo$ImmutableEmptyMap because it has no accessible default constructor
       com.google.api.client.util.Types.handleExceptionForNewInstance(Types.java:162)
       com.google.api.client.util.Types.newInstance(Types.java:117)
       com.google.api.client.util.Data.createNullInstance(Data.java:166)
       com.google.api.client.util.Data.nullOf(Data.java:134)
       com.google.cloud.storage.BlobInfo$BuilderImpl.setMetadata(BlobInfo.java:585)
       [...]
     Caused by: java.lang.InstantiationException: Type `com.google.cloud.storage.BlobInfo.ImmutableEmptyMap` can not be instantiated reflectively as it does not have a no-parameter constructor or the no-parameter constructor has not been added explicitly to the native image.
       java.lang.Class.newInstance(DynamicHub.java:909)
       com.google.api.client.util.Types.newInstance(Types.java:113)
       [...]
```
The `resource-config.json` file makes certain test resource accessible to the native image builder. Specifically, `testNamesWhichCanFail.txt`. It is currently located under src/test as it applies to only tests.

The `native-image.properties` file is located under src/test as it only applies to the tests :

This file handles class initialization during native image generation. Classes can be initialized at either run-time or image build time and class initialization can result in recursive initialization of other classes. The tests in the Storage repo use `Parameterized` which is [explicitly initialized at image build time by the configurations provided by GraalVM](https://github.com/oracle/graal/blob/e9166a984e4c73087713d79fdf0fa7d5e67f16fc/substratevm/src/com.oracle.svm.junit/src/com/oracle/svm/junit/JUnitFeature.java#L76). This leads to a bunch of other unexpected classes being initialized at build time as well and leads to the following error:

```
Error: Classes that should be initialized at run time got initialized during image building:
 com.google.cloud.conformance.storage.v1.RetryTests was unintentionally initialized at build time. To see why com.google.cloud.conformance.storage.v1.RetryTests got initialized use --trace-class-initialization=com.google.cloud.conformance.storage.v1.RetryTests
com.google.cloud.conformance.storage.v1.InstructionList was unintentionally initialized at build time. To see why com.google.cloud.conformance.storage.v1.InstructionList got initialized use --trace-class-initialization=com.google.cloud.conformance.storage.v1.InstructionList
```
And initializing those same classes at run-time results in image generation failing with inconclusive results:
```
Error: Classes that should be initialized at run time got initialized during image building:
 com.google.cloud.storage.conformance.retry.ITRetryConformanceTest$RetryTestCaseResolver the class was requested to be initialized at run time
```
Initializing these classes explicitly at build-time results in a successful build.


google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/resource-config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "resources":{
+    "includes":[
+      {"pattern": ".*.txt"}
```
BenWhitehead, line 4: Is it bad if we include everything under `src/test/resources/`? That would match more the expectation of the full classpath being included.

google-cloud-storage/src/main/resources/META-INF/native-image/com/google/cloud/storage/reflect-config.json
```diff
@@ -0,0 +1,27 @@
...
+  {
+    "name":"org.apache.commons.logging.impl.Jdk14Logger",
+    "methods":[{"name":"<init>","parameterTypes":["java.lang.String"] }]}
+,
+  {
+    "name":"org.apache.commons.logging.impl.LogFactoryImpl",
+    "allDeclaredFields":true,
+    "allDeclaredMethods":true,
+    "methods":[{"name":"<init>","parameterTypes":[] }]}
+,
```
BenWhitehead, line 17: We're not supposed to have any dependency on apache logging. Do you know how I might find the reference path for these classes to try and eliminate them?

google-cloud-storage/src/main/resources/META-INF/native-image/com/google/cloud/storage/proxy-config.json
```diff
@@ -0,0 +1,5 @@
+[
+  {
+    "interfaces":["com.google.cloud.storage.spi.v1.StorageRpc"]
+  }
+]
```
BenWhitehead, line 5: Is it possible to move this file to `src/test/resources/...` since it's only needed for tests?

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/native-image.properties
```diff
@@ -0,0 +1,26 @@
...
+  com.google.common.collect,\
+  com.google.gson.internal.reflect,\
+  com.google.gson.internal.bind,\
+  com.google.gson.internal,\
+  com.google.gson.internal.sql.SqlTypesSupport,\
+  com.google.gson.FieldNamingPolicy$3,\
+  com.google.gson.LongSerializationPolicy$2
+
+
+
```
BenWhitehead, line 26: Does this mean that `com.google.cloud.storage.conformance.retry.ITRetryConformanceTest#testCases` is being invoked during the building of the native image?

We're using `@RunWith(ParallelParameterized.class)` to resolve test cases relative to a classpath resource, and run them in parallel using a threadpool.

google-cloud-storage/src/main/resources/META-INF/native-image/com/google/cloud/storage/reflect-config.json
```diff
@@ -0,0 +1,27 @@
...
+{
+  "name":"com.google.cloud.storage.BlobInfo$ImmutableEmptyMap",
+  "methods":[{"name":"<init>","parameterTypes":[] }]}
+
+,
+{
+  "name":"com.google.cloud.storage.conformance.retry.TestBench$RetryTestResource",
+  "allDeclaredFields":true,
+  "methods":[{"name":"<init>","parameterTypes":[] }]}
```
BenWhitehead, line 26: This class is test only. I don't think we'd want it in a manifest that ends up in our released jar. Is there anyway to move this to a sibling file under `src/test/resources/...`?

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/resource-config.json
mpeddada1: It might be okay to move this under src/test/resources. Are we referencing any `.txt` files in the regular application?

google-cloud-storage/src/main/resources/META-INF/native-image/com/google/cloud/storage/reflect-config.json
mpeddada1: I looked into this a bit by removing the `LogFactoryImpl` configuration and running `mvn clean install -DskipTests` and `mvn test -Pnative`. This results in the following stacktrace:
```
JUnit Vintage:ITStorageTest:testWriteChannelWithConnectionPool
    MethodSource [className = 'com.google.cloud.storage.it.ITStorageTest', methodName = 'testWriteChannelWithConnectionPool', methodParameterTypes = '']
    => java.lang.ExceptionInInitializerError
       org.apache.http.conn.ssl.SSLConnectionSocketFactory.<clinit>(SSLConnectionSocketFactory.java:151)
       org.apache.http.impl.conn.PoolingHttpClientConnectionManager.getDefaultRegistry(PoolingHttpClientConnectionManager.java:116)
       org.apache.http.impl.conn.PoolingHttpClientConnectionManager.<init>(PoolingHttpClientConnectionManager.java:123)
       com.google.cloud.storage.it.ITStorageTest$CustomHttpTransportFactory.create(ITStorageTest.java:282)
       com.google.cloud.storage.spi.v1.HttpStorageRpc.<init>(HttpStorageRpc.java:110)
       [...]
     Caused by: org.apache.commons.logging.LogConfigurationException: java.lang.ClassNotFoundException: org.apache.commons.logging.impl.LogFactoryImpl (Caused by java.lang.ClassNotFoundException: org.apache.commons.logging.impl.LogFactoryImpl)
       org.apache.commons.logging.LogFactory.createFactory(LogFactory.java:1158)
       org.apache.commons.logging.LogFactory$2.run(LogFactory.java:960)
       java.security.AccessController.doPrivileged(AccessController.java:87)
       org.apache.commons.logging.LogFactory.newFactory(LogFactory.java:957)
       org.apache.commons.logging.LogFactory.getFactory(LogFactory.java:624)
       [...]
     Caused by: java.lang.ClassNotFoundException: org.apache.commons.logging.impl.LogFactoryImpl
       jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:53)
       jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
       java.lang.ClassLoader.loadClass(ClassLoader.java:139)
       org.apache.commons.logging.LogFactory.createFactory(LogFactory.java:1020)
       [...]
```

It looks the usage of LogFactory is coming from `PoolingHttpClientConnectionManager` which is referenced in `ITStorageTest`.

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/native-image.properties
mpeddada1: While standard [Java initializes classes the first time they are accessed at run-time](https://docs.oracle.com/en/graalvm/enterprise/20/docs/reference-manual/native-image/ClassInitialization/), Native image compilation tries to do some optimizations were it initializes some classes at image build time to decrease the start-up time of an application. 

In our case, the `JunitFeature` which is brought in by the native-maven-plugin (plugin we use to run tests with native image compilation) explicitly initializes `Parameterized` at image build time. This causes `ParallelParameterized` and subsequently `com.google.cloud.storage.conformance.retry.ITRetryConformanceTest` and other classes `ITRetryConformanceTest` references to also be initialized at build time. While the class itself is initialized at build time, it still runs at run-time (after the native image has been built).

This [article](https://medium.com/graalvm/updates-on-class-initialization-in-graalvm-native-image-generation-c61faca461f7) has been helping me understand class initialization a bit in graal.

google-cloud-storage/src/main/resources/META-INF/native-image/com/google/cloud/storage/reflect-config.json
mpeddada1: Good catch, let's move this to src/test/resources to prevent it from ending up in the released jar.

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/resource-config.json
BenWhitehead: No, we currently don't have anything under `src/main/resources/` that would need to ship in the jar or be taken into consideration for native-image.

google-cloud-storage/src/main/resources/META-INF/native-image/com/google/cloud/storage/reflect-config.json
BenWhitehead: Ah okay, so that's testing for an optional feature. Thanks for trying it out

google-cloud-storage/src/main/resources/META-INF/native-image/com/google/cloud/storage/proxy-config.json
BenWhitehead: Great, thank you!

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/resource-config.json
mpeddada1: Nice, keeping this in `src/test/resources` in that case.

mpeddada1:
@BenWhitehead any idea why some checks are stalled at the moment?

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/native-image.properties
```diff
@@ -0,0 +1,26 @@
+Args = \
```
suztomo, line 1: The file that tells build-time initialization of classes of the library shouldn't be in test. Because the file is not shipped as part of the library.

As discussed today, this file being under the test directory means the users would have to write something similar to use google-cloud-storage with native-image.

mpeddada1: Hm, the only reason why these configurations are initialized at build-time is because  `com.google.cloud.storage.conformance.retry.ITRetryConformanceTest` which references them is initialized at build-time (following[ this logic](https://github.com/googleapis/java-storage/pull/1305#discussion_r834555617)).

suztomo: That's my misunderstanding then. Sorry I missed the discussion. Would you add that comment in the file about the configuration being for the test classes?

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/reflect-config.json
```diff
@@ -0,0 +1,27 @@
...
+    "methods":[{"name":"<init>","parameterTypes":["java.lang.String"] }]}
+,
+  {
+    "name":"org.apache.commons.logging.impl.LogFactoryImpl",
+    "allDeclaredFields":true,
+    "allDeclaredMethods":true,
+    "methods":[{"name":"<init>","parameterTypes":[] }]}
+,
+{
+  "name":"com.google.cloud.storage.BlobInfo$ImmutableEmptyMap",
```
suztomo, line 19: Would you add comment explaining why this configuration is not in main?

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/proxy-config.json
```diff
@@ -0,0 +1,5 @@
+[
+  {
+    "interfaces":["com.google.cloud.storage.spi.v1.StorageRpc"]
```
suztomo, line 3: Would you add comment explaining why this configuration is not in main?

suztomo:
Now the checks are running (Kokoro sample failed)

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/reflect-config.json
mpeddada1: Sorry, my bad. I unintentionally placed this class config in the test directory. This configuration is used outside of the tests so needs to in the main/resources directory.

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/proxy-config.json
mpeddada1: Thanks for point this out! I realized that this class also needs to be in the main config directory since it is referenced outside of tests. For example:
https://github.com/googleapis/java-storage/blob/3716105b5ea83378e5cacb4c4cd480b6dfe0fece/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L90
Apologies for the confusion @BenWhitehead!

BenWhitehead (approved):
Verified config by running:
```bash
export JAVA_HOME=~/opt/java/11-native
mvn -Pnative clean verify
```

BenWhitehead:
Feel free to merge when you're ready. The test failure in Samples is being dealt with separately and isn't related to any change here.

suztomo (approved):

google-cloud-storage/src/test/resources/META-INF/native-image/com/google/cloud/storage/native-image.properties
suztomo: Now the `--initialize-at-build-time` doesn't have google-cloud-storage's main package. Nice.

mpeddada1:
Thank you!

PR #1309, opened 2022-03-25
fix: add gccl-invocation-id interceptor

frankyn:
Joint effort with @BenWhitehead who is the author of the original design, thank you Ben!
- [x] Add option to disable the invocation id
- [x] Make sure this header isn't passed in for signed URLs


google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -38,8 +38,8 @@
   private static final String GCS_SCOPE = "https://www.googleapis.com/auth/devstorage.full_control";
   private static final Set<String> SCOPES = ImmutableSet.of(GCS_SCOPE);
   private static final String DEFAULT_HOST = "https://storage.googleapis.com";
-
   private final RetryAlgorithmManager retryAlgorithmManager;
+  private final Boolean invocationIDEnabled;
```
BenWhitehead, line 42: Since we're always going to have a value, we can use `boolean` throughout for this flag. `Boolean` is really only useful if you need to deal with `true|false|null`.

BenWhitehead, line 42: nit: Thoughts on the name `includeGcclInvocationId`?

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -93,6 +94,18 @@ public Builder setStorageRetryStrategy(StorageRetryStrategy storageRetryStrategy
...
 
+    /**
+     * Override default enablement of invocation id added to x-goog-api-client header.
+     *
+     * @param storageInvocationIDEnabled a non-null Boolean to change enablement of invocation id
+     * @return the builder
+     */
+    public Builder setStorageInvocationIDEnabled(Boolean storageInvocationIDEnabled) {
+      this.storageInvocationIDEnabled =
+          requireNonNull(storageInvocationIDEnabled, "storageInvocationIDEnabled must be non null");
```
BenWhitehead, line 105: Once we change from `Boolean` -> `boolean` we won't need this nonNull check anymore.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -105,6 +118,9 @@ private StorageOptions(Builder builder, StorageDefaults serviceDefaults) {
         new RetryAlgorithmManager(
             MoreObjects.firstNonNull(
                 builder.storageRetryStrategy, serviceDefaults.getStorageRetryStrategy()));
+    this.invocationIDEnabled =
+        MoreObjects.firstNonNull(
+            builder.storageInvocationIDEnabled, serviceDefaults.getStorageInvocationIDEnabled());
```
BenWhitehead, line 123: We won't need this after moving to `boolean`. We'll be able to update the builder to read `serviceDefaults.getStorageInvocationIDEnabled()` when constructed, and not worry about it after that point.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -127,6 +143,10 @@ public TransportOptions getDefaultTransportOptions() {
     public StorageRetryStrategy getStorageRetryStrategy() {
       return StorageRetryStrategy.getDefaultStorageRetryStrategy();
     }
+
+    public Boolean getStorageInvocationIDEnabled() {
+      return Boolean.TRUE;
```
BenWhitehead, line 148:
```suggestion
      return true;
```

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -153,6 +173,11 @@ RetryAlgorithmManager getRetryAlgorithmManager() {
     return retryAlgorithmManager;
   }
 
+  /** Returns if Invocation ID is enabled and transmitted through x-goog-api-client header. */
+  public boolean getInvocationIDEnabled() {
+    return invocationIDEnabled.booleanValue();
```
BenWhitehead, line 178: drop `booleanValue()` after changing to `boolean`.

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
```diff
@@ -424,6 +425,13 @@ private void initializeServiceDependentObjects() {
     expectedBucket3 = new Bucket(storage, new BucketInfo.BuilderImpl(BUCKET_INFO3));
   }
 
+  @Test
+  public void testInvocationIDDisabledBucket() {
```
BenWhitehead, line 429: What does `Bucket` mean in this test name?

BenWhitehead (changes requested):

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplTest.java
frankyn: typo!

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -153,6 +171,11 @@ RetryAlgorithmManager getRetryAlgorithmManager() {
     return retryAlgorithmManager;
   }
 
+  /** Returns if Invocation ID is enabled and transmitted through x-goog-api-client header. */
+  public boolean getIncludeInvocationId() {
```
BenWhitehead, line 175: Ditto
```suggestion
  public boolean isIncludeInvocationId() {
```

BenWhitehead (changes requested):
We also need to add the following to `private Builder(StorageOptions options)` so we ensure we carry through any value set previously.

```
      this.storageIncludeInvocationId = options.includeInvocationId;
```

BenWhitehead (approved):

PR #1324, opened 2022-03-29
fix: update boundary checking of BlobReadChannel when limit() is used

BenWhitehead:
Add several new integration tests to exercise various boundary handling.


tritone (approved):
Nice! curious about how you ran into this condition...

PR #1348, opened 2022-04-08
fix(java): add service account email to Native Image testing kokoro job

mpeddada1:
This PR reduces the discrepancy between the standard Java and Native Image kokoro job. It should also fix the following error:
```
java.lang.AssertionError: Unable to determine service account email
	at com.google.cloud.storage.it.ITStorageTest.testHmacKey(ITStorageTest.java:2472)
```

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -91,6 +91,20 @@ branchProtectionRules:
       - units (11)
       - 'Kokoro - Test: Integration'
       - cla/google
+  - pattern: feat/grpc-storage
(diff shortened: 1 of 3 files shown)


BenWhitehead:
@mpeddada1 You'll need to add an exclude to owlbot.py so it doesn't clobber your update. We had to do the same thing for integration and samples https://github.com/googleapis/java-storage/blob/a4f95038e56ac36badf68edd072705621fec1dbe/owlbot.py#L34

mpeddada1:
Thank you @BenWhitehead!

PR #1354, opened 2022-04-15
feat: Add downloadTo to the Storage interface

### LINKED ISSUE #1158: Provide an efficient way to download the latest version of an object
`Blob#downloadTo` is the library's efficient method for downloading blob data to an arbitrary sink. It is efficient because, unlike `Storage#reader`, it downloads the blob in one request as opposed to many, and because it performs much less copying. This results in significantly higher throughput and significantly lower memory pressure.

The sequence of `Storage#get` followed by `Blob#downloadTo`, which is required to use this API, is prone to races. This is due to the way that the blob's generation, as it existed at `Storage#get`, is unconditionally encoded into the HTTP request generated by `Blob#downloadTo`. If the bucket does not use versioning, the generation may change between `Storage#get` and `Blob#downloadTo`, leading to a 404.

With `Storage#reader` we can avoid this problem by not specifying the generation, for example using `reader(String, String)`, which will read the latest generation. No such workaround is possible with `Blob#downloadTo`, meaning that a usage of `Blob#downloadTo` has to have a fallback that catches the 404 and retries using `Storage#reader` with the generation unspecified. This is inefficient and inelegant, and should not be required.

sydney-munro:

Fixes #1158  ☕️


google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2672,6 +2674,22 @@ Blob createFrom(
    */
   ReadChannel reader(BlobId blob, BlobSourceOption... options);
 
+  /**
```
sydney-munro, line 2677: I wasn't sure if the code sample was required for each of these javadoc comments? I can add it if wanted.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -245,20 +241,7 @@ public void downloadTo(Path path, BlobSourceOption... options) {
...
-    Retrying.run(
-        this.options,
-        algorithm,
-        callable(
-            () -> {
-              storageRpc.read(
-                  pb, requestOptions, countingOutputStream.getCount(), countingOutputStream);
-            }),
-        Function.identity());
+    storage.downloadTo(getBlobId(), outputStream, options);
```
BenWhitehead, line 244: Looks great! Can we also do the same thing for `#downloadTo(Path, BlobSourceOptions...)`?

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
BenWhitehead: Yeah, let's add the small sample here similar to the other methods.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -543,6 +547,32 @@ public ReadChannel reader(BlobId blob, BlobSourceOption... options) {
     return new BlobReadChannel(getOptions(), blob, optionsMap);
   }
 
+  @Override
+  public void downloadTo(BlobId blob, Path path, Blob.BlobSourceOption... options) {
```
BenWhitehead, line 551: Oh, you did get the method added here. Can you update `Blob#downloadTo(Path, BlobSourceOption...)` to call this?

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2672,6 +2674,24 @@ Blob createFrom(
...
+   * Downloads the given blob to the given path using specified blob read options.
+   *
+   * @param blob
+   * @param path
+   * @param options
+   */
+  void downloadTo(BlobId blob, Path path, Blob.BlobSourceOption... options);
```
BenWhitehead, line 2684:
```suggestion
  void downloadTo(BlobId blob, Path path, BlobSourceOption... options);
```

The storage interface already has a BlobSourceOption type used by the other methods (`reader`, `readAllBytes`) we'll want to use the same thing here.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2672,6 +2674,24 @@ Blob createFrom(
...
+
+  /**
+   * Downloads the given blob to the given output stream using specified blob read options.
+   *
+   * @param blob
+   * @param outputStream
+   * @param options
+   */
+  void downloadTo(BlobId blob, OutputStream outputStream, Blob.BlobSourceOption... options);
```
BenWhitehead, line 2693:
```suggestion
  void downloadTo(BlobId blob, OutputStream outputStream, BlobSourceOption... options);
```

Same as above

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
BenWhitehead, line 244: After changing the signature on storage, we'll need to convert the options before calling. `com.google.cloud.storage.Blob.BlobSourceOption#toSourceOptions(com.google.cloud.storage.BlobInfo, com.google.cloud.storage.Blob.BlobSourceOption...)` can be used to perform the conversion.

BenWhitehead (changes requested):

BenWhitehead (approved):

PR #1385, opened 2022-05-12
feat: bucket metadata

frankyn:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
@@ -20,7 +20,7 @@ import static com.google.cloud.storage.Utils.ifNonNull;
 import static com.google.cloud.storage.Utils.todo;
 
 import com.google.cloud.storage.Conversions.Codec;
-import com.google.protobuf.Timestamp;
+import com.google.protobuf.util.Timestamps;
 import com.google.storage.v2.Bucket;
 import com.google.storage.v2.Bucket.Billing;
 import com.google.storage.v2.Object;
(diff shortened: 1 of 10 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -106,4 +111,40 @@ Codec<BlobId, Object> blobId() {
...
+    ifNonNull(retentionPolicy, Bucket.RetentionPolicy::getRetentionPeriod, to::setRetentionPeriod);
+    if (from.hasRetentionPolicy() && retentionPolicy.hasEffectiveTime()) {
+      to.setRetentionEffectiveTime(retentionPolicy.getEffectiveTime().getSeconds());
+    }
+    ifNonNull(from.getBucketId(), to::setGeneratedId);
+    ifNonNull(from.getLocation(), to::setLocation);
+    ifNonNull(from.getLocationType(), to::setLocationType);
+    ifNonNull(from.getMetageneration(), to::setMetageneration);
+    ifNonNull(from.getBilling(), Billing::getRequesterPays, to::setRequesterPays);
+    ifNonNull(from.getCreateTime(), Timestamp::getSeconds, to::setCreateTime);
```
sydney-munro, line 128: The time values in these conversions is a bit confusing.. I believe that for BucketInfo the values should be in milliseconds but there is no option in Timestamp to set milliseconds. see https://cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.BucketInfo#com_google_cloud_storage_BucketInfo_getUpdateTime__
however update time is the only field that specifically calls out milliseconds from unix epoch.

frankyn: Good catch, we may not catch more of these until we get into integration testing.

We can use [toMillis()](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/util/Timestamps.html#toMillis-com.google.protobuf.Timestamp-), thanks!

BenWhitehead (approved):

PR #1394, opened 2022-05-16
feat: refactor model classes to prefer to java.time types

BenWhitehead:
### What's been added

Fields which map to apiary DateTime RFC 3339 fields will now also have
a preferred java.time.OffsetDateTime getter & setter.

Fields which map to apiary long to represent a duration in epoch millis
will no also have a preferred java.time.Duration getter & setter.

All new methods currently have the name of the corresponding java.time
type appended to the property name.

#### Refactor Long -> java.time.OffsetDateTime
* BlobInfo#createTime
* BlobInfo#customTime
* BlobInfo#deleteTime
* BlobInfo#retentionExpirationTime
* BlobInfo#timeStorageClassUpdated
* BlobInfo#updateTime
* Blob#createTime
* Blob#customTime
* Blob#deleteTime
* Blob#retentionExpirationTime
* Blob#timeStorageClassUpdated
* Blob#updateTime
* BucketInfo.CreatedBeforeDeleteRule
* BucketInfo.DeleteRule
* BucketInfo.IsLiveDeleteRule
* BucketInfo.NumNewerVersionsDeleteRule
* BucketInfo.RawDeleteRule
* BucketInfo#createTime
* BucketInfo#updateTime
* Bucket#createTime
* Bucket#updateTime
* Bucket#retentionEffectiveTime
* HmacKey.HmacKeyMetadata#createTime
* HmacKey.HmacKeyMetadata#updateTime

#### Refactor Long -> java.time.Duration
* BucketInfo#retentionPeriod
* Bucket#retentionPeriod

### Deprecation of existing methods
All existing "Long" based methods have been marked as `@Deprecated`
and had their javadocs updated to point to the new preferred
corresponding java.time method.

When the existing "Long" based methods are removed, the corresponding
java.time method will replace it at which point the java.time suffix
methods will be deprecated for later cleanup.

All existing "Long" based methods take care of performing the necessary
conversion to the respective java.time type and will continue to work
until their removal.

##### Why append the java.time type as a suffix?
1. Getters can't be overloaded, so we already need something to disambiguate
   which type should be returned.
2. For setters: we theoretically could overload since we have different 
   method arguments, however, if a customer is already passing `null` as the
   method argument we've now created an ambiguity as to which method to
   invoke. By suffixing the setters as well, we avoid creating this possible
   ambiguity.

Part 2 of https://github.com/googleapis/java-storage/pull/1388

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -17,6 +17,8 @@
 package com.google.cloud.storage;
 
 import static com.google.cloud.storage.BackwardCompatibilityUtils.millisOffsetDateTimeCodec;
+import static com.google.cloud.storage.BackwardCompatibilityUtils.millisUtcCodec;
+import static com.google.cloud.storage.BackwardCompatibilityUtils.nullableDurationMillisCodec;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkNotNull;
 
@@ -34,6 +36,7 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
+import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Map;
@@ -81,7 +84,7 @@ public class BucketInfo implements Serializable {
   private final Boolean defaultEventBasedHold;
   private final OffsetDateTime retentionEffectiveTime;
   private final Boolean retentionPolicyIsLocked;
-  private final Long retentionPeriod; // TODO: duration
+  private final Duration retentionPeriod;
   private final IamConfiguration iamConfiguration;
   private final String locationType;
   private final Logging logging;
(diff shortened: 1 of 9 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1414,14 +1416,16 @@ public Builder setDefaultEventBasedHold(Boolean defaultEventBasedHold) {
       return this;
     }
 
+    /** @deprecated Use {@link #setRetentionEffectiveTimeOffsetDateTime(OffsetDateTime)} */
     @Override
     @Deprecated
     Builder setRetentionEffectiveTime(Long retentionEffectiveTime) {
-      return setRetentionEffectiveTime(millisOffsetDateTimeCodec.encode(retentionEffectiveTime));
+      return setRetentionEffectiveTimeOffsetDateTime(
```
sydney-munro, line 1423: nit: inconsistencies between style.
before you do 
```
Builder setUpdateTime(Long updateTime) {
      this.updateTime = millisOffsetDateTimeCodec.encode(updateTime);
      return this;
    }
```
vs calling the offset method in the deprecated old method.

sydney-munro (approved):
LGTM just a small nit.

PR #1395, opened 2022-05-16
test: introduce BaseConvertablePropertyTest

BenWhitehead:
All Convertable types share a common set of properties that they need to provide. This new class will ensure in a central location they do.

First property: a round trip through the codec should provide a value equal to the initial input


sydney-munro (approved):
LGTM

PR #1401, opened 2022-05-19
test: update ITStorageTest#testHmacKey to be courteous

BenWhitehead:
Previously testHmacKey would clobber any existing hmac keys for the service account when it started running, and multiple points during the test. This change updates the behavior to only try and cleanup an hmac key if it's been more than 24hrs since last updated.

Additionally, a project is only allowed 5 hmac keys total, creating 4 keys in the test to check pagination creates contention for other simultaneous runs that need to use hmac keys. I've removed this, the logic for pagination is the same as for other things like blobs and buckets we can depend on those to ensure dynamic page sizing works.


JesseLovelace (approved):
Thanks for this ben!

PR #1403, opened 2022-05-23
WIP Blob/Object get

sydney-munro:


google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
```diff
@@ -137,4 +151,48 @@ public String get() {
...
+
+    public Arbitrary<ObjectChecksums> objectChecksumsArbitrary() {
+      return Combinators.combine(
+              Arbitraries.integers().greaterOrEqual(1),
+              Arbitraries.strings()
+                  .map(
+                      s ->
+                          BaseEncoding.base64()
+                              .encode(Hashing.md5().hashBytes(s.getBytes()).asBytes())))
+          .as(
```
BenWhitehead, line 177: This can be done in a later PR if you feel like it.

This will need to changes from a combine of two arbitraries, to a single arbitrary which yields both crc32c and md5 values.

We want to ensure the value for those fields is generated base on the same bytes in the general case. Later we can investigate if we want to add an edge case where the md5 does not correspond to crc32c of the same input data.

google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
```diff
@@ -137,4 +151,48 @@ public String get() {
...
+              Arbitraries.strings()
+                  .map(s -> Hashing.sha256().hashString(s, StandardCharsets.UTF_8).asBytes()))
+          .as(
+              (algorithm, key) ->
+                  CustomerEncryption.newBuilder()
+                      .setEncryptionAlgorithm(algorithm)
+                      .setKeySha256Bytes(ByteString.copyFrom(key))
+                      .build());
+    }
+  }
```
BenWhitehead, line 197: Nit: move the ByteString copy up in the arbitrary generation
```suggestion
    public Arbitrary<CustomerEncryption> customerEncryptionArbitrary() {
      return Combinators.combine(
              Arbitraries.strings().ofMinLength(1).ofMaxLength(1024),
              Arbitraries.strings()
                  .map(s -> Hashing.sha256().hashString(s, StandardCharsets.UTF_8).asBytes())
                  .map(ByteString::copyFrom))
          .as(
              (algorithm, key) ->
                  CustomerEncryption.newBuilder()
                      .setEncryptionAlgorithm(algorithm)
                      .setKeySha256Bytes(key)
                      .build());
    }

```

BenWhitehead (changes requested):
Overall looks pretty good. A few changes about base64 handling need to be addressed, but other than that looks pretty good.

google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
sydney-munro: ah I see makes sense. I will look into it.

BenWhitehead (approved):

PR #1415, opened 2022-05-25
chore: refactor GrpcConversions to use OffsetDateTime instead of deprecated methods

BenWhitehead:
* Add GrpcConversions#timestampCodec convert between java.time.OffsetDateTime and com.google.protobuf.Timestamp
* Add property test for timestampCodec
* Replace all usage of "Long" methods in GrpcConversions with their appropriate java.time alternatives


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageTest.java
```diff
@@ -167,6 +167,7 @@
 import org.threeten.bp.ZoneOffset;
 import org.threeten.bp.format.DateTimeFormatter;
 
+@org.junit.Ignore
```
sydney-munro, line 170: Why are we ignoring this?

BenWhitehead: Woops, accidental. I meant to unstage this when I commited. I'll pull it out.

google-cloud-storage/src/test/java/com/google/cloud/storage/TimestampCodecPropertyTest.java
```diff
@@ -0,0 +1,55 @@
...
+    OffsetDateTime decode = codec.decode(ts);
+    Timestamp encode = codec.encode(decode);
+
+    assertThat(encode).isEqualTo(ts);
+  }
+
+  private static final class Supp implements ArbitrarySupplier<Timestamp> {
+    @Override
+    public Arbitrary<Timestamp> get() {
+      return StorageArbitraries.timestamp();
```
frankyn, line 52: Why did you need this helper static method?

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -63,6 +66,19 @@ final class GrpcConversions {
   private final Codec<Integer, String> crc32cCodec =
       Codec.of(this::crc32cEncode, this::crc32cDecode);
 
+  @VisibleForTesting
+  final Codec<OffsetDateTime, Timestamp> timestampCodec =
```
frankyn, line 70: Why isn't this in our codecs source instead?

frankyn (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
BenWhitehead: Following the pattern from ApiaryConversions for that conversion to DataTime[1].

Give this conversion is specific to the use of protos it probably makes sense to keep it near the proto conversions. Or, is there another location you'd like to see this moved to?

[1] https://github.com/googleapis/java-storage/blob/359504b1b8ca4f15b13106cfbd0440cc22598511/google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryConversions.java#L131-L152

google-cloud-storage/src/test/java/com/google/cloud/storage/TimestampCodecPropertyTest.java
BenWhitehead: Rather than the `Arbitrary<Timestamp>` from `StorageArbitraries` being registered globally, this method provides the resolution handle to jqwik so it knows how to generate arbitraries for the test method.

Above on line 41 `@ForAll(supplier = Supp.class)` points to this class as the means to resolve the Arbitrary<Timestamp>.

BenWhitehead: If we'd rather it be registered globally, I can do that too.

frankyn: oh, this is just to test the timestampCodec specifically; wouldn't you need a timestamp supplier for other tests such as bucket/ object metadata? I think that's what is confusing me. This is more of a question that can be answered after PR approval. Thanks Ben

frankyn (approved):

google-cloud-storage/src/test/java/com/google/cloud/storage/TimestampCodecPropertyTest.java
BenWhitehead: Other things would also need the timestamp supplier, which so far has been accessed and composed directly via `StorageArbitraries#timestamp()`.

This supplier class is effectively wiring so that jqwik finds that method for the test.

Issue #294, opened 2020-04-30
Make Storage Interface Extend Autocloseable 


mikail-khan:
The Firestore interface extends the Autocloseable interface and thereby offers a close method which can be called to guarantee resource clean up. The Storage interface does not extend Autocloseable.

I would like the Storage interface to extend Autocloseable so that a close method is provided. This will allow me to write conciser code in my Scala library I am building which wraps `java-storage`.

For now I can use the traditional `try-with-resources` idiom. But it would be great if the GCP Java APIs are consistent where they can be and is reasonable to do so.

athakor:
@frankyn I am planing to introduced the `close `method of `AutoCloseable` like below way but there is one problem with `Transportation `layer. 
As we know `StorageOption.setTransportOptions(TransportOptions transportOptions)` method accept the `TransportOptions` and that is implemented in [HttpTransportOptions](https://github.com/googleapis/java-core/blob/master/google-cloud-core-http/src/main/java/com/google/cloud/http/HttpTransportOptions.java), these `HttpTransport `layer use the following factories to create the connection.

**CustomHttpTransportFactory**
     
Below block of code works because these factory used the `ApacheHttpTransport` layer and that layer have `shutdown `method and it's allow us to shutdown the connection manager and releases the allocated resources.you can see here https://github.com/googleapis/google-http-java-client/blob/1624d55f9864ff1253321e9bfe715e2f370cc27e/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpTransport.java#L188

**DefaultHttpTransportFactory**

Below block of code doesn't works here, because here by default use `NetHttpTransport `layer and that layer doesn't have `shutdown `or close method to shutdown the connection manager or to release allocated resources.you can check here https://github.com/googleapis/google-http-java-client/blob/master/google-http client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java

```
    @Override
    public void close() throws Exception {
        if (closed) {
            return;
        }
        closed = true;
        transport.shutdown();
    }
```

frankyn:
I would hold on this for now. Let's revisit at a later time, there are definitely unknowns that must considered but this isn't prioritized right now.

PR #1422, opened 2022-05-27
feat: make Storage extend AutoClosable

BenWhitehead:
* Add StorageFixture to manage the test scoped lifecycle of a storage instance
* Update ITStorageTest to use new fixture
* Update ITBlobReadChannelTest to use new fixture


rdesgroppes:
For the record: #294.

PR #1431, opened 2022-06-02
feat: BucketInfo conversion for gRPC

frankyn:
Introducing more support for BucketInfo conversions

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryConversions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryConversions.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.storage;
 
+import static com.google.cloud.storage.Utils.dateTimeCodec;
 import static com.google.cloud.storage.Utils.durationMillisCodec;
 import static com.google.cloud.storage.Utils.ifNonNull;
 import static com.google.cloud.storage.Utils.lift;
@@ -192,10 +193,10 @@ final class ApiaryConversions {
   private StorageObject blobInfoEncode(BlobInfo from) {
     StorageObject to = blobIdEncode(from.getBlobId());
     ifNonNull(from.getAcl(), toImmutableListOf(objectAcl()::encode), to::setAcl);
-    ifNonNull(from.getDeleteTimeOffsetDateTime(), Utils.dateTimeCodec::encode, to::setTimeDeleted);
-    ifNonNull(from.getUpdateTimeOffsetDateTime(), Utils.dateTimeCodec::encode, to::setUpdated);
-    ifNonNull(from.getCreateTimeOffsetDateTime(), Utils.dateTimeCodec::encode, to::setTimeCreated);
-    ifNonNull(from.getCustomTimeOffsetDateTime(), Utils.dateTimeCodec::encode, to::setCustomTime);
+    ifNonNull(from.getDeleteTimeOffsetDateTime(), dateTimeCodec::encode, to::setTimeDeleted);
+    ifNonNull(from.getUpdateTimeOffsetDateTime(), dateTimeCodec::encode, to::setUpdated);
+    ifNonNull(from.getCreateTimeOffsetDateTime(), dateTimeCodec::encode, to::setTimeCreated);
+    ifNonNull(from.getCustomTimeOffsetDateTime(), dateTimeCodec::encode, to::setCustomTime);
     ifNonNull(from.getSize(), BigInteger::valueOf, to::setSize);
     ifNonNull(
         from.getOwner(),
@@ -204,13 +205,13 @@ final class ApiaryConversions {
     ifNonNull(from.getStorageClass(), StorageClass::toString, to::setStorageClass);
     ifNonNull(
         from.getTimeStorageClassUpdatedOffsetDateTime(),
-        Utils.dateTimeCodec::encode,
+        dateTimeCodec::encode,
         to::setTimeStorageClassUpdated);
     ifNonNull(
         from.getCustomerEncryption(), this::customerEncryptionEncode, to::setCustomerEncryption);
     ifNonNull(
         from.getRetentionExpirationTimeOffsetDateTime(),
-        Utils.dateTimeCodec::encode,
+        dateTimeCodec::encode,
         to::setRetentionExpirationTime);
     to.setKmsKeyName(from.getKmsKeyName());
     to.setEventBasedHold(from.getEventBasedHold());
@@ -256,10 +257,10 @@ final class ApiaryConversions {
     ifNonNull(from.getId(), to::setGeneratedId);
     ifNonNull(from.getSelfLink(), to::setSelfLink);
     ifNonNull(from.getMetadata(), to::setMetadata);
-    ifNonNull(from.getTimeDeleted(), Utils.dateTimeCodec::decode, to::setDeleteTimeOffsetDateTime);
-    ifNonNull(from.getUpdated(), Utils.dateTimeCodec::decode, to::setUpdateTimeOffsetDateTime);
-    ifNonNull(from.getTimeCreated(), Utils.dateTimeCodec::decode, to::setCreateTimeOffsetDateTime);
-    ifNonNull(from.getCustomTime(), Utils.dateTimeCodec::decode, to::setCustomTimeOffsetDateTime);
+    ifNonNull(from.getTimeDeleted(), dateTimeCodec::decode, to::setDeleteTimeOffsetDateTime);
+    ifNonNull(from.getUpdated(), dateTimeCodec::decode, to::setUpdateTimeOffsetDateTime);
+    ifNonNull(from.getTimeCreated(), dateTimeCodec::decode, to::setCreateTimeOffsetDateTime);
+    ifNonNull(from.getCustomTime(), dateTimeCodec::decode, to::setCustomTimeOffsetDateTime);
     ifNonNull(from.getSize(), BigInteger::longValue, to::setSize);
     ifNonNull(from.getOwner(), lift(Owner::getEntity).andThen(this::entityDecode), to::setOwner);
     ifNonNull(from.getAcl(), toImmutableListOf(objectAcl()::decode), to::setAcl);
@@ -271,14 +272,14 @@ final class ApiaryConversions {
     ifNonNull(from.getStorageClass(), StorageClass::valueOf, to::setStorageClass);
     ifNonNull(
         from.getTimeStorageClassUpdated(),
-        Utils.dateTimeCodec::decode,
+        dateTimeCodec::decode,
         to::setTimeStorageClassUpdatedOffsetDateTime);
     ifNonNull(from.getKmsKeyName(), to::setKmsKeyName);
     ifNonNull(from.getEventBasedHold(), to::setEventBasedHold);
     ifNonNull(from.getTemporaryHold(), to::setTemporaryHold);
     ifNonNull(
         from.getRetentionExpirationTime(),
-        Utils.dateTimeCodec::decode,
+        dateTimeCodec::decode,
         to::setRetentionExpirationTimeOffsetDateTime);
     return to.build();
   }
@@ -309,7 +310,7 @@ final class ApiaryConversions {
     Bucket to = new Bucket();
     ifNonNull(from.getAcl(), toImmutableListOf(bucketAcl()::encode), to::setAcl);
     ifNonNull(from.getCors(), toImmutableListOf(cors()::encode), to::setCors);
-    ifNonNull(from.getCreateTimeOffsetDateTime(), Utils.dateTimeCodec::encode, to::setTimeCreated);
+    ifNonNull(from.getCreateTimeOffsetDateTime(), dateTimeCodec::encode, to::setTimeCreated);
     ifNonNull(
         from.getDefaultAcl(), toImmutableListOf(objectAcl()::encode), to::setDefaultObjectAcl);
     ifNonNull(from.getLocation(), to::setLocation);
@@ -321,7 +322,7 @@ final class ApiaryConversions {
         to::setOwner);
     ifNonNull(from.getRpo(), Rpo::toString, to::setRpo);
     ifNonNull(from.getStorageClass(), StorageClass::toString, to::setStorageClass);
-    ifNonNull(from.getUpdateTimeOffsetDateTime(), Utils.dateTimeCodec::encode, to::setUpdated);
+    ifNonNull(from.getUpdateTimeOffsetDateTime(), dateTimeCodec::encode, to::setUpdated);
     ifNonNull(from.versioningEnabled(), b -> new Versioning().setEnabled(b), to::setVersioning);
     to.setEtag(from.getEtag());
     to.setId(from.getGeneratedId());
@@ -381,7 +382,7 @@ final class ApiaryConversions {
       retentionPolicy.setRetentionPeriod(durationMillisCodec.encode(retentionPeriod));
       ifNonNull(
           from.getRetentionEffectiveTimeOffsetDateTime(),
-          Utils.dateTimeCodec::encode,
+          dateTimeCodec::encode,
           retentionPolicy::setEffectiveTime);
       ifNonNull(from.retentionPolicyIsLocked(), retentionPolicy::setIsLocked);
       to.setRetentionPolicy(retentionPolicy);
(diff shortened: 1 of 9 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryConversions.java
```diff
@@ -221,10 +192,10 @@ Codec<NotificationInfo, com.google.api.services.storage.model.Notification> noti
   private StorageObject blobInfoEncode(BlobInfo from) {
     StorageObject to = blobIdEncode(from.getBlobId());
     ifNonNull(from.getAcl(), toImmutableListOf(objectAcl()::encode), to::setAcl);
-    ifNonNull(from.getDeleteTimeOffsetDateTime(), dateTimeCodec::encode, to::setTimeDeleted);
-    ifNonNull(from.getUpdateTimeOffsetDateTime(), dateTimeCodec::encode, to::setUpdated);
-    ifNonNull(from.getCreateTimeOffsetDateTime(), dateTimeCodec::encode, to::setTimeCreated);
-    ifNonNull(from.getCustomTimeOffsetDateTime(), dateTimeCodec::encode, to::setCustomTime);
+    ifNonNull(from.getDeleteTimeOffsetDateTime(), Utils.dateTimeCodec::encode, to::setTimeDeleted);
```
BenWhitehead, line 195: If you do a static import of `dateTimeCodec`, these lines won't be modified.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -549,13 +556,37 @@ public Integer getDaysSinceNoncurrentTime() {
...
+       * @deprecated {@link #getNoncurrentTimeBeforeOffsetDateTime()}
        */
+      @Deprecated
       public DateTime getNoncurrentTimeBefore() {
+        if (noncurrentTimeBefore == null) return null;
+        return new DateTime(millisOffsetDateTimeCodec.decode(noncurrentTimeBefore));
+      }
+
+      /**
+       * Returns the date in RFC 3339 format with only the date part (for instance, "2013-01-15").
```
BenWhitehead, line 569: `OffesetDateTime` isn't exactly an RFC 3339 date.

I think we'll want to change the comment here to be a bit more explicit. Something like:

> Returns the date and offset from UTC for this condition.
> If a time other than `00:00:00.000` is present in the value, GCS will truncate to `00:00:00.000`.

google-cloud-storage/src/test/java/com/google/cloud/storage/BucketPropertyTest.java
```diff
@@ -16,33 +16,12 @@
...
 
-import static com.google.common.truth.Truth.assertThat;
-
-import com.google.protobuf.util.Timestamps;
 import com.google.storage.v2.Bucket;
-import net.jqwik.api.ForAll;
-import net.jqwik.api.Property;
 
-final class BucketPropertyTest {
+final class BucketInfoPropertyTest extends BaseConvertablePropertyTest<BucketInfo, Bucket> {
```
BenWhitehead, line 21: 👍🏻

google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
```diff
@@ -49,6 +53,20 @@ public static Arbitrary<Timestamp> timestamp() {
                 Timestamp.newBuilder().setSeconds(odt.toEpochSecond()).setNanos(nanos).build());
   }
 
+  public static Arbitrary<Date> date() {
+    return DateTimes.offsetDateTimes()
+        .offsetBetween(ZoneOffset.UTC, ZoneOffset.UTC)
+        .map(
+            odt -> {
+              OffsetDateTime utc = odt.withOffsetSameInstant(ZoneOffset.UTC);
```
BenWhitehead, line 61: I don't think you need this transform here, the input has previously been constrained to UTC.

google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
```diff
@@ -122,6 +140,65 @@ public static final class Buckets {
...
+                              .addAllMatchesStorageClass(s)
+                              .build())
+                      .build());
+    }
+
+    public Arbitrary<Bucket.Lifecycle> lifecycle() {
+      return rule()
+          .list()
+          .ofMinSize(0)
+          .ofMaxSize(100)
```
BenWhitehead, line 198: Is there an actual limit on the gcs side for how many rules can be present? If so it might make sense to use a similar value here.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -549,13 +556,37 @@ public Integer getDaysSinceNoncurrentTime() {
...
-      /* Returns the date in RFC 3339 format with only the date part (for instance, "2013-01-15").*/
+      /**
+       * Returns the date in RFC 3339 format with only the date part (for instance, "2013-01-15").
+       *
+       * @deprecated Use {@link #getCustomTimeBeforeOffsetDateTime()}
+       */
+      @Deprecated
       public DateTime getCustomTimeBefore() {
+        if (customTimeBefore == null) return null;
+        return new DateTime(millisOffsetDateTimeCodec.decode(customTimeBefore));
```
BenWhitehead, line 583: I think these two lines can be replaced with:
```suggestion
        return Utils.dateTimeCodec.nullable().encode(customTimeBefore);
```

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -549,13 +556,37 @@ public Integer getDaysSinceNoncurrentTime() {
 
       /**
        * Returns the date in RFC 3339 format with only the date part (for instance, "2013-01-15").
+       *
+       * @deprecated {@link #getNoncurrentTimeBeforeOffsetDateTime()}
        */
+      @Deprecated
       public DateTime getNoncurrentTimeBefore() {
+        if (noncurrentTimeBefore == null) return null;
+        return new DateTime(millisOffsetDateTimeCodec.decode(noncurrentTimeBefore));
```
BenWhitehead, line 565: Same as `#getCustomTimeBefore`

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -526,7 +526,14 @@ public Integer getAge() {
         return age;
       }
 
+      /** @deprecated Use {@link #getCreatedBeforeOffsetDateTime()} */
+      @Deprecated
       public DateTime getCreatedBefore() {
+        if (createdBefore == null) return null;
+        return new DateTime(millisOffsetDateTimeCodec.decode(createdBefore));
```
BenWhitehead, line 533: Same as `#getCustomTimeBefore`

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -594,9 +625,24 @@ public Builder setAge(Integer age) {
...
-         * UTC. *
+         * UTC.
+         *
+         * @deprecated Use {@link #setCreateBeforeOffsetDateTime(OffsetDateTime)}
          */
+        @Deprecated
         public Builder setCreatedBefore(DateTime createdBefore) {
+          if (createdBefore == null) return this;
+          return setCreateBeforeOffsetDateTime(
+              millisOffsetDateTimeCodec.encode(createdBefore.getValue()));
```
BenWhitehead, line 636: Similar to the getter, I think this can be simplified to
```suggestion
          return setCustomTimeBeforeOffsetDateTime(
              Utils.dateTimeCodec.nullable().decode(customTimeBefore));
```

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -646,8 +692,23 @@ public Builder setDaysSinceNoncurrentTime(Integer daysSinceNoncurrentTime) {
...
          * truncated. This condition is satisfied when the noncurrent time on an object is before
          * this date. This condition is relevant only for versioned objects.
+         *
+         * @deprecated Use {@link #setNoncurrentTimeBeforeOffsetDateTime(OffsetDateTime)}
          */
+        @Deprecated
         public Builder setNoncurrentTimeBefore(DateTime noncurrentTimeBefore) {
+          if (noncurrentTimeBefore == null) return this;
+          return setNoncurrentTimeBeforeOffsetDateTime(
+              millisOffsetDateTimeCodec.encode(noncurrentTimeBefore.getValue()));
```
BenWhitehead, line 702: Same as `#setCustomDateTime`

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -657,8 +718,23 @@ public Builder setNoncurrentTimeBefore(DateTime noncurrentTimeBefore) {
...
          * truncated. This condition is satisfied when the custom time on an object is before this
          * date in UTC.
+         *
+         * @deprecated {@link #setCustomTimeBeforeOffsetDateTime(OffsetDateTime)}
          */
+        @Deprecated
         public Builder setCustomTimeBefore(DateTime customTimeBefore) {
+          if (customTimeBefore == null) return this;
+          return setCustomTimeBeforeOffsetDateTime(
+              millisOffsetDateTimeCodec.encode(customTimeBefore.getValue()));
```
BenWhitehead, line 728: Same as `#setCustomDateTime`

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -594,9 +625,24 @@ public Builder setAge(Integer age) {
...
+         * UTC.
+         *
+         * @deprecated Use {@link #setCreateBeforeOffsetDateTime(OffsetDateTime)}
          */
+        @Deprecated
         public Builder setCreatedBefore(DateTime createdBefore) {
+          if (createdBefore == null) return this;
+          return setCreateBeforeOffsetDateTime(
+              millisOffsetDateTimeCodec.encode(createdBefore.getValue()));
+        }
```
BenWhitehead, line 637: Additionally, when using the `Utils.dateTimeCodec.nullable()` you'll want to put that in a variable so it doesn't construct a new codec instance each time.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -165,26 +190,276 @@ private BucketInfo bucketInfoDecode(Bucket from) {
...
     ifNonNull(from.getVersioning(), Bucket.Versioning::getEnabled, to::setVersioningEnabled);
     ifNonNull(from.getDefaultEventBasedHold(), to::setDefaultEventBasedHold);
     ifNonNull(from.getLabels(), to::setLabels);
-    ifNonNull(from.getBilling(), Bucket.Billing::getRequesterPays, to::setRequesterPays);
+    if (from.getWebsite() != null
+        || from.getWebsite().getMainPageSuffix() != null
+        || from.getWebsite().getNotFoundPage() != null) {
+      to.setIndexPage(from.getWebsite().getMainPageSuffix());
+      to.setNotFoundPage(from.getWebsite().getNotFoundPage());
+    }
```
BenWhitehead, line 203: According to the proto definition, both of these properties are non-optional.
https://github.com/googleapis/java-storage/blob/1f14acd53d481f0a290b33cb207c79be8cc2cc1c/proto-google-cloud-storage-v2/src/main/proto/google/storage/v2/storage.proto#L1520-L1533

This if condition can be simplified to:
```
if (from.hasWebsite())
```

google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
frankyn: I thought it was 100 but i can't seem to find a limit. I can remove the upperbound but 100 feels okay to me. What do you think?

BenWhitehead: 100's probably fine in that case. We probably won't want to somehow end up slowing down tests because the generator needs to construct like 700k of these things.

google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
```diff
@@ -140,14 +139,21 @@ public static final class Buckets {
...
+          .as(
+              (a, s) -> {
+                Bucket.Lifecycle.Rule.Action.Builder actionBuilder =
+                    Bucket.Lifecycle.Rule.Action.newBuilder();
+                actionBuilder.setType(a);
+                if (a.equals("SetStorageClass")) {
+                  actionBuilder.setStorageClass(s);
+                }
+                return actionBuilder.build();
+              });
```
BenWhitehead, line 153: I found the previous version easier to read, but if you feel this is easier feel free to keep it.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -657,8 +717,21 @@ public Builder setNoncurrentTimeBefore(DateTime noncurrentTimeBefore) {
          * Note that only date part will be considered, if the time is specified it will be
          * truncated. This condition is satisfied when the custom time on an object is before this
          * date in UTC.
+         *
+         * @deprecated {@link #setCustomTimeBeforeOffsetDateTime(OffsetDateTime)}
```
BenWhitehead, line 721: Missing the `Use` on this statement like the others

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -549,13 +559,37 @@ public Integer getDaysSinceNoncurrentTime() {
 
       /**
        * Returns the date in RFC 3339 format with only the date part (for instance, "2013-01-15").
+       *
+       * @deprecated {@link #getNoncurrentTimeBeforeOffsetDateTime()}
```
BenWhitehead, line 563: Missing the `Use` on this statement like the others

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -646,8 +693,21 @@ public Builder setDaysSinceNoncurrentTime(Integer daysSinceNoncurrentTime) {
...
         public Builder setNoncurrentTimeBefore(DateTime noncurrentTimeBefore) {
+          return setNoncurrentTimeBeforeOffsetDateTime(Utils.dateTimeCodec.nullable().decode(noncurrentTimeBefore));
+        }
+
+        /**
+         * Sets the date in RFC 3339 format with only the date part (for instance, "2013-01-15").
+         * Note that only date part will be considered, if the time is specified it will be
+         * truncated. This condition is satisfied when the noncurrent time on an object is before
+         * this date. This condition is relevant only for versioned objects.
+         */
```
BenWhitehead, line 709: Can we drop the RFC 3339 part of this statement?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -657,8 +717,21 @@ public Builder setNoncurrentTimeBefore(DateTime noncurrentTimeBefore) {
...
+         *
+         * @deprecated {@link #setCustomTimeBeforeOffsetDateTime(OffsetDateTime)}
          */
+        @Deprecated
         public Builder setCustomTimeBefore(DateTime customTimeBefore) {
+          return setCustomTimeBeforeOffsetDateTime(Utils.dateTimeCodec.nullable().decode(customTimeBefore));
+        }
+
+        /**
+         * Sets the date in RFC 3339 format with only the date part (for instance, "2013-01-15").
```
BenWhitehead, line 729: Can we drop the RFC 3339 part of this statement.

BenWhitehead (approved):
A few minor requests to update some of the javadocs. I'll approve so you can merge after updating those items.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -165,26 +192,279 @@ private BucketInfo bucketInfoDecode(Bucket from) {
...
+    ifNonNull(from.getDefaultEventBasedHold(), to::setDefaultEventBasedHold);
+    ifNonNull(from.getLabels(), to::putAllLabels);
+    // preserve mapping to deprecated property
+    Bucket.Lifecycle.Builder lifecycleBuilder = Bucket.Lifecycle.newBuilder();
+    ifNonNull(
+        from.getLifecycleRules(),
+        Utils.toImmutableListOf(lifecycleRule()::encode),
+        lifecycleBuilder::addAllRule);
+    /*
+    // Corruption occurs by having deleteRule() values added
```
frankyn, line 273: @BenWhitehead deleteRule() corrupts the policy and I commented it out. I can do a follow-up PR to better handle this case as this PR is long. WDYT?

BenWhitehead: SGTM

PR #1432, opened 2022-06-02
chore: add initial skeleton of Retrying for Grpc

BenWhitehead:
* rename RetryAlgorithmManager -> HttpRetryAlgorithmManager
* Add GrpcRetryAlgorithmManager
* Update Retrying.run(GrpcStorageOptions, ...) to take a Decoder instead
  of a Function (this helps narrow things a bit to our decoders)
* Add GrpcStorageImpl#SyntaxDecoders
  * For our "Syntax" types - Blob, Bucket, etc - Decoding an instance is
    always of the form message -> info -> syntax. This new class gives
    us a decoder that can use simply for message -> syntax by binding
    to the enclosing storage instance.
  * update "Syntax" return points to use new decoders

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -31,6 +31,7 @@ import com.google.cloud.Policy;
 import com.google.cloud.ReadChannel;
 import com.google.cloud.WriteChannel;
 import com.google.cloud.storage.Acl.Entity;
+import com.google.cloud.storage.Conversions.Decoder;
 import com.google.cloud.storage.HmacKey.HmacKeyMetadata;
 import com.google.cloud.storage.HmacKey.HmacKeyState;
 import com.google.cloud.storage.PostPolicyV4.PostConditionsV4;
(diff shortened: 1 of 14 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -572,8 +582,14 @@ public void deleteHmacKey(HmacKeyMetadata hmacKeyMetadata, DeleteHmacKeyOption..
...
-    // TODO retries
-    grpcStorageStub.deleteHmacKeyCallable().call(req);
+    Retrying.run(
+        getOptions(),
+        retryAlgorithmManager.getFor(req),
+        () -> {
+          grpcStorageStub.deleteHmacKeyCallable().call(req);
+          return null;
+        },
+        Decoder.identity());
```
frankyn, line 592: Instead of identity, could we use an overload which makes Decoder optional?

frankyn (changes requested):
Overall lgtm, just have one nit

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
BenWhitehead: In StorageImpl we're also using `identity()` for the few methods where necessary. These methods are almost exclusively delete methods (which arguably should return something better than void or boolean). So I'm not sure they are enough motivation to overload `Retrying.run`

frankyn (approved):

Issue #1425, opened 2022-05-31
Can not download ~200GB file using storage blob


rdubitsky-syberry:
Execution fails when ~128gb is downloaded and code fail does not depend on time spent on downloading this 128 gb of file.

#### Environment details

OS type and version: Linux
Java version: 11

#### Code example

```java
JsonObject jsonObject = new JsonObject();
jsonObject.addProperty(GoogleCredentialsConstants.CLIENT_ID, CLIENT_ID);
jsonObject.addProperty(GoogleCredentialsConstants.CLIENT_EMAIL, CLIENT_EMAIL);
jsonObject.addProperty(GoogleCredentialsConstants.ACCOUNT_TYPE, ACCOUNT_TYPE);
jsonObject.addProperty(GoogleCredentialsConstants.PRIVATE_KEY_ID, PRIVATE_KEY_ID);
jsonObject.addProperty(GoogleCredentialsConstants.PRIVATE_KEY, PRIVATE_KEY);
String credentials = jsonObject.toString();

GoogleCredentials googleCredentials;
try (InputStream inputStream = new ByteArrayInputStream(credentials.getBytes(StandardCharsets.UTF_8))) {
    googleCredentials = GoogleCredentials.fromStream(inputStream);
} catch (IOException ex) {
    throw new CustomException(ex);
}

Storage storage = StorageOptions.newBuilder()
      .setProjectId(projectId)
      .setCredentials(googleCredentials )
      .build()
      .getService();

Blob blob = storage.get(BlobId.of(bucketName, remoteFilePath));
blob.downloadTo(localFilePath); // line where code fails
```

#### Stack trace
```
Caused by: java.net.SocketException: Connection reset by peer (Write failed)
	at java.base/java.net.SocketOutputStream.socketWrite0(Native Method)
	at java.base/java.net.SocketOutputStream.socketWrite(Unknown Source)
	at java.base/java.net.SocketOutputStream.write(Unknown Source)
	at java.base/sun.security.ssl.SSLSocketOutputRecord.flush(Unknown Source)
	at java.base/sun.security.ssl.OutputRecord.changeWriteCiphers(Unknown Source)
	at java.base/sun.security.ssl.KeyUpdate$KeyUpdateProducer.produce(Unknown Source)
	at java.base/sun.security.ssl.KeyUpdate$KeyUpdateKickstartProducer.produce(Unknown Source)
	at java.base/sun.security.ssl.SSLHandshake.kickstart(Unknown Source)
[...]

PR #1434, opened 2022-06-06
fix: update request method of HttpStorageRpc to properly configure offset on requests

BenWhitehead:
When invoking downloadTo(..., OutputStream) if a retry was attempted the proper byte
offset was not being sent in the retried request. Update logic of HttpStorageRpc.read
to manually set the range header rather than trying to rely on MediaDownloader to do
it along with not automatically decompressing the byte stream.

Update ITRetryConformanceTest to run Scenario 8 test cases, which cover resuming a
download which could have caught this error sooner.

Update StorageException.translate(IOException) to classify `IOException: Premature EOF`
as the existing retryable reason `connectionClosedPrematurely`. Add case to
DefaultRetryHandlingBehaviorTest to ensure conformance to this categorization.

Break downloadTo integration test out into their own class, and separate
the multiple scenarios being tested in the same method.

Related to https://github.com/googleapis/java-storage/issues/1425

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -24,6 +24,7 @@ import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 import com.google.api.client.googleapis.batch.BatchRequest;
 import com.google.api.client.googleapis.batch.json.JsonBatchCallback;
 import com.google.api.client.googleapis.json.GoogleJsonError;
+import com.google.api.client.googleapis.media.MediaHttpDownloader;
 import com.google.api.client.http.ByteArrayContent;
 import com.google.api.client.http.EmptyContent;
 import com.google.api.client.http.GenericUrl;
@@ -283,7 +284,7 @@ public class HttpStorageRpc implements StorageRpc {
   }
 
   private static StorageException translate(IOException exception) {
-    return new StorageException(exception);
+    return StorageException.translate(exception);
   }
 
   private static StorageException translate(GoogleJsonError exception) {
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultRetryHandlingBehaviorTest.java
```diff
@@ -149,7 +152,7 @@ public static Collection<Object[]> testCases() {
 
     // ensure all permutations are accounted for, reporting any that haven't been and providing
     // a stub which can be used to easily define them.
-    assertWithMessage("Missing mappings for tokens").that(difference).isEmpty();
+    // assertWithMessage("Missing mappings for tokens").that(difference).isEmpty();
```
frankyn, line 155: are you planning on removing this line?

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -750,10 +751,11 @@ public long read(
...
       }
-      req.getMediaHttpDownloader().setBytesDownloaded(position);
-      req.getMediaHttpDownloader().setDirectDownloadEnabled(true);
-      req.executeMedia().download(outputStream);
-      return req.getMediaHttpDownloader().getNumBytesDownloaded();
+      MediaHttpDownloader mediaHttpDownloader = req.getMediaHttpDownloader();
+      mediaHttpDownloader.setBytesDownloaded(position);
+      mediaHttpDownloader.setDirectDownloadEnabled(true);
+      req.executeMediaAndDownloadTo(outputStream);
+      return mediaHttpDownloader.getNumBytesDownloaded();
```
frankyn, line 758: Is this a refactor or is there magical Apiary functionality hidden in this?

frankyn (changes requested):

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultRetryHandlingBehaviorTest.java
BenWhitehead: This was an accident, I'll revert this line.

frankyn (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
BenWhitehead: I have since moved away from the previous executeAndDownload method change. This was not properly respecting `returnRawInputStream` which would break gzip automatic transcoding if you wanted to actually download the zipped bytes. Instead, we are now setting the range header ourselves.

PR #1435, opened 2022-06-06
feat: Bucket metadata grpc

frankyn:
Add remaining fields for Bucket metadata (defaultObjectAcls, Owner, Logging, Cors, and IamConfiguration) 
Fix OLM deleteRules encode/decode to only function when Rules are exactly what's expected otherwise return null.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
@@ -29,6 +29,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.io.BaseEncoding;
 import com.google.common.primitives.Ints;
 import com.google.protobuf.ByteString;
+import com.google.protobuf.Descriptors;
 import com.google.protobuf.Timestamp;
 import com.google.storage.v2.Bucket;
 import com.google.storage.v2.Bucket.Billing;
@@ -42,8 +43,7 @@ import java.time.Instant;
 import java.time.LocalDate;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
-import java.util.logging.Level;
-import java.util.logging.LogRecord;
+import java.util.List;
 
 final class GrpcConversions {
   static final GrpcConversions INSTANCE = new GrpcConversions();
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -263,52 +272,179 @@ private Bucket bucketInfoEncode(BucketInfo from) {
...
+        from.getDefaultAcl(), toImmutableListOf(objectAcl()::encode), to::addAllDefaultObjectAcl);
+    ifNonNull(from.getIamConfiguration(), iamConfiguration()::encode, to::setIamConfig);
     // TODO(frankyn): Add SelfLink when the field is available
-    // TODO(frankyn): Add decoder for iamConfig
     // TODO(frankyn): Add Etag when support is avialable
     return to.build();
   }
 
+  private Bucket.Lifecycle buildLifecyclePolicy(BucketInfo from) {
+    // Handle duplicate rules introduced by deleteRules using a backing ImmutableSet
```
frankyn, line 291: Handling duplicates, what are your thoughts @BenWhitehead; I'm deduping proto Rule message types, but not sure if equals() will handle unknown fields as different. Looking into this next; but thought you may already know.

frankyn: Looks like this should be okay, tested an unknown field on Rule (thankfully action and condition are both messages as well): Let me know if I goofed.

```
  @Test
  public void assertNotEqualWithUnknownField() {
    com.google.storage.v2.Bucket.Lifecycle.Rule rule =
        com.google.storage.v2.Bucket.Lifecycle.Rule.getDefaultInstance();

    assertEquals(
        com.google.storage.v2.Bucket.Lifecycle.Rule.getDefaultInstance()
            .toBuilder()
            .setUnknownFields(
                UnknownFieldSet.newBuilder()
                    .addField(100, UnknownFieldSet.Field.newBuilder().addFixed32(10).build())
                    .build()).build(), com.google.storage.v2.Bucket.Lifecycle.Rule.getDefaultInstance());
  }
```

BenWhitehead: Hmm, duplicates is an interesting case that we haven't thus far accounted for.

In fact the existing BucketInfo.fromPb[1] is duplicating and not deduplicating later.

I think we're going to want to start another PR after merging this one, where we actually go through and remove all internal usage of DeleteRule and instead translate the respective delete rule operations into the correct operations for lifecycle rules. Then, we can remove this duplication from both the apiary and grpc converters.

Regarding handling of unknown fields, we only have to account for them when decoding to BlobInfo. Once the BlobInfo is decoded, the unknown fields will effectively be abandoned. 

[1] https://github.com/googleapis/java-storage/blob/main/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java#L2200-L2219

BenWhitehead (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
BenWhitehead: We spoke offline, we're going to open a separate PR to remove the duplication in BucketInfo and instead shim the old DeleteRules over exclusively LifecycleRules

PR #1439, opened 2022-06-07
feat: TransformPageDecorator 

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -268,7 +268,7 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
     Function<com.google.storage.v2.Bucket, BucketInfo> decode = codecs.bucketInfo()::decode;
 
     Function<com.google.storage.v2.Bucket, Bucket> translator =
-        decode.andThen(bi -> bi.asBucket(this));
+        decode.andThen(bucketInfo -> bucketInfo.asBucket(this));
     return new TransformPageDecorator<>(page, translator);
   }
 
@@ -281,7 +281,8 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
     ListObjectsPagedResponse call = listObjectsPagedResponseUnaryCallable.call(req);
     ListObjectsPage page = call.getPage();
     Function<com.google.storage.v2.Object, BlobInfo> decode = codecs.blobInfo()::decode;
-    Function<com.google.storage.v2.Object, Blob> translator = decode.andThen(bi -> bi.asBlob(this));
+    Function<com.google.storage.v2.Object, Blob> translator =
+        decode.andThen(blobInfo -> blobInfo.asBlob(this));
     return new TransformPageDecorator<>(page, translator);
   }
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -79,7 +79,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
 
 final class GrpcStorageImpl extends BaseService<StorageOptions> implements Storage {
 
@@ -257,9 +256,8 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
 
   @Override
   public Page<Bucket> list(BucketListOption... options) {
-    UnaryCallable<ListBucketsRequest, ListBucketsPagedResponse>
-        listBucketsCallable =
-            grpcStorageStub.listBucketsPagedCallable();
+    UnaryCallable<ListBucketsRequest, ListBucketsPagedResponse> listBucketsCallable =
+        grpcStorageStub.listBucketsPagedCallable();
     final Map<StorageRpc.Option, ?> optionsMap = StorageImpl.optionMap(options);
     ListBucketsRequest.Builder builder = ListBucketsRequest.newBuilder();
     ifNonNull(
@@ -268,20 +266,15 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
     ifNonNull((String) optionsMap.get(StorageRpc.Option.PREFIX), builder::setPrefix);
     // TODO(sydmunro): StorageRpc.Option.Fields
     // TODO(sydmunro): User Project
-    ListBucketsPagedResponse call =
-        listBucketsRequestListBucketsPagedResponseUnaryCallable.call(builder.build());
+    ListBucketsPagedResponse call = listBucketsCallable.call(builder.build());
     ListBucketsPage page = call.getPage();
-    Function<com.google.storage.v2.Bucket, BucketInfo> decode = codecs.bucketInfo()::decode;
-
-    Function<com.google.storage.v2.Bucket, Bucket> translator =
-        decode.andThen(bucketInfo -> bucketInfo.asBucket(this));
-    return new TransformPageDecorator<>(page, translator);
+    return new TransformingPageDecorator<>(page, syntaxDecoders.bucket);
   }
 
   @Override
   public Page<Blob> list(String bucket, BlobListOption... options) {
-    UnaryCallable<ListObjectsRequest, ListObjectsPagedResponse>
-        listObjectsCallable = grpcStorageStub.listObjectsPagedCallable();
+    UnaryCallable<ListObjectsRequest, ListObjectsPagedResponse> listObjectsCallable =
+        grpcStorageStub.listObjectsPagedCallable();
     final Map<StorageRpc.Option, ?> optionsMap = StorageImpl.optionMap(options);
     ListObjectsRequest.Builder builder = ListObjectsRequest.newBuilder().setParent(bucket);
     ifNonNull(
@@ -293,12 +286,9 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
         (String) optionsMap.get(StorageRpc.Option.START_OFF_SET), builder::setLexicographicStart);
     ifNonNull((String) optionsMap.get(StorageRpc.Option.END_OFF_SET), builder::setLexicographicEnd);
     // TODO(sydmunro) StorageRpc.Option.Fields
-    ListObjectsPagedResponse call = listObjectsPagedResponseUnaryCallable.call(builder.build());
+    ListObjectsPagedResponse call = listObjectsCallable.call(builder.build());
     ListObjectsPage page = call.getPage();
-    Function<com.google.storage.v2.Object, BlobInfo> decode = codecs.blobInfo()::decode;
-    Function<com.google.storage.v2.Object, Blob> translator =
-        decode.andThen(blobInfo -> blobInfo.asBlob(this));
-    return new TransformPageDecorator<>(page, translator);
+    return new TransformingPageDecorator<>(page, syntaxDecoders.blob);
   }
 
   @Override
@@ -615,9 +605,8 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
 
   @Override
   public Page<HmacKeyMetadata> listHmacKeys(ListHmacKeysOption... options) {
-    UnaryCallable<ListHmacKeysRequest, ListHmacKeysPagedResponse>
-        listHmacKeysCallable =
-            grpcStorageStub.listHmacKeysPagedCallable();
+    UnaryCallable<ListHmacKeysRequest, ListHmacKeysPagedResponse> listHmacKeysCallable =
+        grpcStorageStub.listHmacKeysPagedCallable();
     final Map<StorageRpc.Option, ?> optionsMap = StorageImpl.optionMap(options);
     // TODO: Project is required?
     ListHmacKeysRequest.Builder builder = ListHmacKeysRequest.newBuilder();
@@ -630,12 +619,9 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
     ifNonNull(
         (Boolean) optionsMap.get(StorageRpc.Option.SHOW_DELETED_KEYS), builder::setShowDeletedKeys);
     ifNonNull((String) optionsMap.get(StorageRpc.Option.PROJECT_ID), builder::setProject);
-    ListHmacKeysPagedResponse call =
-        listBucketsRequestListHmacKeysPagedResponseUnaryCallable.call(builder.build());
+    ListHmacKeysPagedResponse call = listHmacKeysCallable.call(builder.build());
     ListHmacKeysPage page = call.getPage();
-    Function<com.google.storage.v2.HmacKeyMetadata, HmacKey.HmacKeyMetadata> decode =
-        codecs.hmacKeyMetadata()::decode;
-    return new TransformPageDecorator<>(page, decode);
+    return new TransformingPageDecorator<>(page, codecs.hmacKeyMetadata());
   }
 
   @Override
@@ -757,7 +743,7 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
         b -> codecs.bucketInfo().decode(b).asBucket(GrpcStorageImpl.this);
   }
 
-  private static class TransformPageDecorator<
+  private static class TransformingPageDecorator<
           RequestT,
           ResponseT,
           ResourceT,
@@ -766,9 +752,9 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
       implements Page<ModelT> {
 
     private final PageT page;
-    private final Function<ResourceT, ModelT> translator;
+    private final Decoder<ResourceT, ModelT> translator;
 
-    public TransformPageDecorator(PageT page, Function<ResourceT, ModelT> translator) {
+    public TransformingPageDecorator(PageT page, Decoder<ResourceT, ModelT> translator) {
       this.page = page;
       this.translator = translator;
     }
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -682,9 +750,78 @@ private Blob internalCreate(
...
+  private static class TransformPageDecorator<
+          RequestT,
+          ResponseT,
+          ResourceT,
+          PageT extends AbstractPage<RequestT, ResponseT, ResourceT, PageT>,
+          ModelT>
+      implements Page<ModelT> {
+
+    private final PageT page;
+    private final Function<ResourceT, ModelT> translator;
```
BenWhitehead, line 769: If we change this field type from `Function<ResourceT, ModelT>` to `Decoder<ResourceT, ModelT>` then we can simplify each of the list methods a bit by directly pointing to the decoders, including the new `syntaxDecoders` that were added during my retry refactor last week.

```
    return new TransformPageDecorator<>(page, syntaxDecoders.bucket);
    return new TransformPageDecorator<>(page, syntaxDecoders.blob);
    return new TransformPageDecorator<>(page, codecs.hmacKeyMetadata());
```

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -567,7 +615,27 @@ public HmacKey createHmacKey(ServiceAccount serviceAccount, CreateHmacKeyOption.
 
   @Override
   public Page<HmacKeyMetadata> listHmacKeys(ListHmacKeysOption... options) {
-    return todo();
+    UnaryCallable<ListHmacKeysRequest, ListHmacKeysPagedResponse>
+        listBucketsRequestListHmacKeysPagedResponseUnaryCallable =
+            grpcStorageStub.listHmacKeysPagedCallable();
+    final Map<StorageRpc.Option, ?> optionsMap = StorageImpl.optionMap(options);
+    // TODO: Project is required?
```
BenWhitehead, line 622: Project is required.

For the JSON implementation, we first attempt to read the supplied option for project_id, if that is null fall back to using the project id from options. We can do the same thing here.

https://github.com/googleapis/java-storage/blob/feat/grpc-storage/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L1447-L1450

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -245,12 +257,48 @@ public Blob get(BlobId blob) {
...
+        listBucketsRequestListBucketsPagedResponseUnaryCallable =
+            grpcStorageStub.listBucketsPagedCallable();
+    final Map<StorageRpc.Option, ?> optionsMap = StorageImpl.optionMap(options);
+    ListBucketsRequest.Builder builder = ListBucketsRequest.newBuilder();
+    ifNonNull(
+        (Long) optionsMap.get(StorageRpc.Option.MAX_RESULTS), Long::intValue, builder::setPageSize);
+    ifNonNull((String) optionsMap.get(StorageRpc.Option.PAGE_TOKEN), builder::setPageToken);
+    ifNonNull((String) optionsMap.get(StorageRpc.Option.PREFIX), builder::setPrefix);
+    // TODO(sydmunro): StorageRpc.Option.Fields
+    // TODO(sydmunro): User Project
```
BenWhitehead, line 270: According to the protos, project is required to be able to list buckets.

We'll probably need to do a similar thing here as for listHmacKeys with the fallback to options.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -245,12 +257,48 @@ public Blob get(BlobId blob) {
 
   @Override
   public Page<Bucket> list(BucketListOption... options) {
-    return todo();
+    UnaryCallable<ListBucketsRequest, ListBucketsPagedResponse>
+        listBucketsRequestListBucketsPagedResponseUnaryCallable =
```
BenWhitehead, line 261: nit
```suggestion
        listBucketsCallable =
```

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -245,12 +257,48 @@ public Blob get(BlobId blob) {
...
+    Function<com.google.storage.v2.Bucket, Bucket> translator =
+        decode.andThen(bucketInfo -> bucketInfo.asBucket(this));
+    return new TransformPageDecorator<>(page, translator);
   }
 
   @Override
   public Page<Blob> list(String bucket, BlobListOption... options) {
-    return todo();
+    UnaryCallable<ListObjectsRequest, ListObjectsPagedResponse>
+        listObjectsPagedResponseUnaryCallable = grpcStorageStub.listObjectsPagedCallable();
```
BenWhitehead, line 284: nit
```suggestion
        listObjectsCallable = grpcStorageStub.listObjectsPagedCallable();
```

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -567,7 +615,27 @@ public HmacKey createHmacKey(ServiceAccount serviceAccount, CreateHmacKeyOption.
 
   @Override
   public Page<HmacKeyMetadata> listHmacKeys(ListHmacKeysOption... options) {
-    return todo();
+    UnaryCallable<ListHmacKeysRequest, ListHmacKeysPagedResponse>
+        listBucketsRequestListHmacKeysPagedResponseUnaryCallable =
```
BenWhitehead, line 619: copy-paste name
```suggestion
        listHmacKeysCallable =
```

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -682,9 +750,78 @@ private Blob internalCreate(
...
   /** Bind some decoders for our "Syntax" classes to this instance of GrpcStorageImpl */
   private final class SyntaxDecoders {
+
     final Decoder<Object, Blob> blob =
         o -> codecs.blobInfo().decode(o).asBlob(GrpcStorageImpl.this);
     final Decoder<com.google.storage.v2.Bucket, Bucket> bucket =
         b -> codecs.bucketInfo().decode(b).asBucket(GrpcStorageImpl.this);
   }
+
+  private static class TransformPageDecorator<
```
BenWhitehead, line 760: nit: `s/Transform/Transforming/g`

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -682,9 +750,78 @@ private Blob internalCreate(
...
+
+    @Override
+    public Iterable<ModelT> getValues() {
+      return () -> {
+        final Iterator<ResourceT> inter = page.getValues().iterator();
+        return new ResourceIterator(inter);
+      };
+    }
+
+    private class ResourceIterator implements Iterator<ModelT> {
```
BenWhitehead, line 807: Nit: `s/ResourceIterator/TransformingIterator/g`

The interesting work this iterator is doing is the transformation, not really anything Resource related.

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
sydney-munro: User project is different though because it is used for requester pays to refer to the billing project, correct? the required field that you are referring to is the projectid right?

frankyn: +1 to @sydney-munro, you're right, UserProject is meant to assign operation cost to a specified project. It's usually used with RequesterPays but may still be used even if requesterPays is not enabled.

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class GrpcTransformPageDecoratorTest {
+
+  @Test
+  public void valueTranslationTest() {
```
frankyn, line 41: Will you be adding a property based test in a follow-up PR?

frankyn (commented):
Overall LGTM, just have one q on property tests

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -743,7 +743,7 @@ private final class SyntaxDecoders {
         b -> codecs.bucketInfo().decode(b).asBucket(GrpcStorageImpl.this);
   }
 
-  private static class TransformingPageDecorator<
+  protected static class TransformingPageDecorator<
```
BenWhitehead, line 752: We can drop protected all together, along with making this class final since we don't intend for it to ever be extended.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
BenWhitehead: According to https://github.com/googleapis/java-storage/blob/8a7755cc8352b3ab21c252885fb86576474d7f09/proto-google-cloud-storage-v2/src/main/proto/google/storage/v2/storage.proto#L336 a project must be specified in a list buckets request.

If that is incorrect, we should request that the protos be updated.

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
sydney-munro: I can do that, I wanted to make a general test because I felt a property test would kind of just be duplicating the conversion tests, the only difference being wrapping the BlobInfo, BucketInfo etc in a Page but the translation logic is the same. Do you think there would be additional value to doing this?

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+    // Initial values for the first page
+    List<String> initialValues = Arrays.asList("string1", "string2", "string3");
```
BenWhitehead, line 43: Guava has an ImmutableList implementation with a convenience factory `ImmutableList.of(...)`.

Unless mutability is specifically necessary, we prefer to use `ImmutableList`s most of the time - either directly, or by converting a mutable list to immutable before leaving the scope.

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+    // Request which will be appended to the second page
+    String request = "string4";
+
+    // Expected values after the translation
+    List<String> expectedValuesPageOne = Arrays.asList("STRING1", "STRING2", "STRING3");
+    List<String> expectedValuesPageTwo = Arrays.asList("STRING4");
```
BenWhitehead, line 49: Arrays.asList -> ImmutableList.of

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+    List<String> expectedPagesValuesMerged =
+        Stream.of(expectedValuesPageOne, expectedValuesPageTwo)
+            .flatMap(list -> list.stream())
+            .collect(Collectors.toList());
```
BenWhitehead, line 53: We'll want to collect to an ImmutableList `.collect(ImmutableList.toImmutableList())`

Other possible means of concatenating these two lists if interested:
```
Streams.concat(expectedValuesPageOne.stream(), expectedValuesPageTwo.stream())
    .collect(ImmutableList.toImmutableList());
```

```
ImmutableList.<String>builder()
    .addAll(expectedValuesPageOne)
    .addAll(expectedValuesPageTwo)
    .build()
```

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    StringPagedListDescriptor descriptor = new StringPagedListDescriptor();
+    UnaryCallable callable = new StringPagedCallable();
```
BenWhitehead, line 56: Missing type parameters

```suggestion
    UnaryCallable<String, List<String>> callable = new StringPagedCallable();
```

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+    PageContext<String, List<String>, String> context =
+        PageContext.create(callable, descriptor, request, Mockito.mock(ApiCallContext.class));
+    ListStringPage page = new ListStringPage(context, initialValues);
+    TransformingPageDecorator decorator =
```
BenWhitehead, line 60: Missing type parameters
```suggestion
    TransformingPageDecorator<String, List<String>, String, ListStringPage, String> decorator =
```

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
frankyn: I was thinking you could use property based testing to generate the list of object names of different sizes (for initial Values, pageOne, pageTwo) then make then the test could do the rest that you're doing in this test. That was my only point in this comment.

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+        new TransformingPageDecorator<>(page, String::toUpperCase);
+
+    assertThat(ImmutableList.copyOf(decorator.getValues().iterator()))
+        .containsExactlyElementsIn(expectedValuesPageOne);
+    assertThat(ImmutableList.copyOf(decorator.iterateAll().iterator()))
+        .containsExactlyElementsIn(expectedPagesValuesMerged);
+  }
+
+  private class ListStringPage extends AbstractPage<String, List<String>, String, ListStringPage> {
```
BenWhitehead, line 69: Class can be `static`

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+      return payload.size() > 0 ? payload.get(0) : emptyToken();
+    }
+
+    @Override
+    public Iterable<String> extractResources(List<String> payload) {
+      return payload;
+    }
+  }
+
+  private class StringPagedCallable extends UnaryCallable<String, List<String>> {
```
BenWhitehead, line 116: Class can be `static`

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+    // We only want to add one additional page with the same value as the request,
+    // this is kind of hacky, but I wanted to validate we are performing iterate all
+    // properly.
+    private int numberOfPages = 1;
+
+    @Override
+    public ApiFuture<List<String>> futureCall(String request, ApiCallContext context) {
+      if (numberOfPages-- > 0) {
+        return ApiFutures.immediateFuture(Arrays.asList(request));
```
BenWhitehead, line 125: Arrays.asList -> ImmutableList.of

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
BenWhitehead: This test is specifically for the PageDecorator.

As far as testing the list calls themselves is concerned, I think those can be added in a later PR once we are also testing with the testbench.

Given that pages all have the same "shape" the things we care about testing there are:
1. Do we construct the request accurately
2. Do we wire in the correct decoder to the constructed TransformingPageDecorator

The decoder wiring part is pretty much verified for us at compile type via type checking.

The request construction part is probably more of a unit test, given how unfriendly the `*Options` stuff is.

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
sydney-munro: yeah, what I am referring to in this comment is that 
https://github.com/googleapis/java-storage/blob/main/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L1015-L1017
is an option but not supported by the proto. The project parent field for ht ListBucketsRequest is the project id which is something else and should be fixed now with https://github.com/googleapis/java-storage/pull/1439/files#diff-8d6230c7cebeacb950a63216b0068065a15d4c2943ac7f18405906a38b3d5f66R263

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcTransformPageDecoratorTest.java
sydney-munro: oh nice, the first suggestion looks cleaner 👍

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
BenWhitehead: Ahh, missed that in the diff-diffs.

BenWhitehead (approved):

PR #1452, opened 2022-06-13
test: add property test for DefaultBufferedReadableByteChannel

BenWhitehead:
New property test against arbitrarily generated data along with buffer size, and app read size.

Ensure the expected number of reads and their sizes are returned for a specific data size.

Ensure channel is closed after full read, and throws a ClosedChannelException when attempted to read after close

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultBufferedReadableByteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultBufferedReadableByteChannel.java
@@ -28,6 +28,9 @@ final class DefaultBufferedReadableByteChannel implements BufferedReadableByteCh
 
   private final UnbufferedReadableByteChannel channel;
 
+  private boolean flipped = false;
+  private boolean retEOF = false;
+
   DefaultBufferedReadableByteChannel(ByteBuffer buffer, UnbufferedReadableByteChannel channel) {
     this.buffer = buffer;
     this.channel = channel;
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultBufferedReadableByteChannel.java
```diff
@@ -109,13 +124,13 @@ public int read(ByteBuffer dst) throws IOException {
             // determine the position buffer needs to be set to
             long bytesReadIntoBuffer = read - dstRemaining;
             Buffers.position(buffer, Math.toIntExact(bytesReadIntoBuffer));
-            tmpBytesCopied = Math.toIntExact(read);
+            flipped = false;
```
frankyn, line 127: The `flipped` variable is confusing for tracking purposes I mean.

google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultBufferedReadableByteChannel.java
```diff
@@ -56,35 +60,40 @@ public int read(ByteBuffer dst) throws IOException {
 
       final int tmpBytesCopied;
       if (enqueuedBytes()) {
+        if (!flipped) {
+          buffer.flip();
+          flipped = true;
+        }
         long copy = Buffers.copy(buffer, new ByteBuffer[] {dst});
         if (buffer.remaining() == 0) {
           Buffers.clear(buffer);
```
frankyn, line 69: Shouldn't this set `flipped = false`?

frankyn (approved):

frankyn:
Chatted over GVC with @BenWhitehead and cleared up the `flipped` variable.

PR #1454, opened 2022-06-15
feat: support formatted bucket name with projectID

frankyn:
1. Added Bucket+ProjectID model in BucketInfo and update setters and getters for Bucket name to use this new model.
1. Added setters and getters for projectID in BucketInfo and Bucket (using BucketName helper)
1. Added unformatted bucket name into property based test "bucket" vs. "projects/project-id/buckets/bucket-id"
1. Support default value when presented with unformatted bucket name to  use "_" for project id
1. Added fall for unformatted into log bucket parser in grpc conversions.
1. Tested unformatted bucket name only for log bucket name (do not do it for BucketInfo)

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -71,7 +71,7 @@ public class BucketInfo implements Serializable {
 
   private static final long serialVersionUID = -4712013629621638459L;
   private final String generatedId;
-  private final String name;
+  private final BucketWithProject bucketWithProject;
   private final Acl.Entity owner;
   private final String selfLink;
   private final Boolean requesterPays;
@@ -332,8 +332,8 @@ public class BucketInfo implements Serializable {
   public static class Logging implements Serializable {
 
     private static final long serialVersionUID = -708892101216778492L;
-    private String logBucket;
-    private String logObjectPrefix;
+    private final BucketWithProject logBucketWithProject;
+    private final String logObjectPrefix;
 
     @Override
     public boolean equals(Object o) {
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
```diff
@@ -84,13 +85,39 @@ public static Arbitrary<String> randomString() {
     return Arbitraries.strings().all().ofMinLength(1).ofMaxLength(1024);
   }
 
+  public static Arbitrary<ProjectID> projectID() {
```
BenWhitehead, line 88: Was it intentional to move this method up here? It's not showing if there is/what the change is from before

google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
```diff
@@ -84,13 +85,39 @@ public static Arbitrary<String> randomString() {
...
+    return Arbitraries.oneOf(
```
BenWhitehead, line 89: There's only a single argument to `oneOf` here. We can remove it.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1900,7 +1925,7 @@ public Builder toBuilder() {
 
   @Override
   public int hashCode() {
-    return Objects.hash(name);
+    return Objects.hash(bucketWithProject.getBucketName());
```
BenWhitehead, line 1928: I don't think we should reach down into the bucketWithProject here. That could result in having two buckets with differing projects but the same name generating the same hashCode.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1920,7 +1945,9 @@ public boolean equals(Object o) {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this).add("name", name).toString();
+    return MoreObjects.toStringHelper(this)
+        .add("name", bucketWithProject.getBucketName())
```
BenWhitehead, line 1949: I think we'll need to add a nullcheck and include the project id if non-null and non `_`

BenWhitehead (changes requested):

google-cloud-storage/src/test/java/com/google/cloud/storage/jqwik/StorageArbitraries.java
frankyn: I wanted to add it before bucketName because it's now part of the bucket proto. /shrug total nit. Other than that there should be no change.

BenWhitehead (approved):

PR #1458, opened 2022-06-16
test: add property test for DefaultBufferedWritableByteChannel

BenWhitehead:
New property test against arbitrarily generated data along with buffer size, and app write size.

Ensure the expected number of writes and their sizes are performed for a specific data size.

Ensure the expected number of flushes are performed, and at the correct intervals.

Ensure channel is closed after full read, and throws a ClosedChannelException when attempted to read after close

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
@@ -139,11 +139,11 @@ public final class DefaultBufferedWritableByteChannelTest {
    * writes:  |----|--------|--|---------|
    *                   10
    * flush 1: |---------|
-   *                     12
+   *            2        12
    * flush 2:   |---------|
-   *                               22
+   *                     12        22
    * flush 3:             |---------|
-   *                                    27
+   *                              19    27
    * flush 4:                     |------|
    * </pre>
    */
@@ -196,11 +196,11 @@ public final class DefaultBufferedWritableByteChannelTest {
    * writes:  |--|--|--|--|
    *             3
    * flush 1: |--|
-   *                6
+   *             3  6
    * flush 2:    |--|
-   *                   10
+   *               5   10
    * flush 3:      |----|
-   *                     12
+   *                   10 12
    * flush 4:           |-|
    * </pre>
    */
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
```diff
@@ -0,0 +1,549 @@
...
+        CountingWritableByteChannelAdapter adapter =
+            new CountingWritableByteChannelAdapter(Channels.newChannel(baos));
+        BufferedWritableByteChannel c = new DefaultBufferedWritableByteChannel(buffer, adapter)) {
+
+      c.write(data1);
+      adapter.nextWriteMaxConsumptionLimit = 2L;
+      c.write(data2);
+
+      c.write(data3);
+      adapter.nextWriteMaxConsumptionLimit = 7L;
```
frankyn, line 176: Where does this `7L` come in from the data pipeline above?

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
```diff
@@ -0,0 +1,549 @@
...
+   *
+   * <pre>
+   *           0                        27
+   * data:    |--------------------------|
+   *               5       14 17        27
+   * writes:  |----|--------|--|---------|
+   *                   10
+   * flush 1: |---------|
+   *                     12
+   * flush 2:   |---------|
```
frankyn, line 143: Could you add the starting index for subsequent flushes?

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
```diff
@@ -0,0 +1,549 @@
...
+      adapter.nextWriteMaxConsumptionLimit = 7L;
+      c.write(data4);
+
+      c.close();
+      assertThrows(ClosedChannelException.class, () -> c.write(null));
+
+      assertWithMessage("Unexpected total flushed length")
+          .that(adapter.writeEndPoints)
+          .isEqualTo(ImmutableList.of(2L, 12L, 19L, 27L));
+      assertThat(baos.toByteArray()).isEqualTo(allData);
```
frankyn, line 185: I thought this was a partial flush, shouldn't you assert that it's equal to a subset of allData?

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
```diff
@@ -0,0 +1,549 @@
...
+    try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        CountingWritableByteChannelAdapter adapter =
+            new CountingWritableByteChannelAdapter(Channels.newChannel(baos));
+        BufferedWritableByteChannel c = new DefaultBufferedWritableByteChannel(buffer, adapter)) {
+
+      c.write(data1);
+      c.flush();
+
+      c.write(data2);
+      adapter.nextWriteMaxConsumptionLimit = 2L;
```
frankyn, line 232: IIUC this is telling the implementation to flush on the next write because data2 (3) is larger than 2 bytes?

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
```diff
@@ -0,0 +1,549 @@
...
+            ByteBuffer.wrap(bytes, 16, 16),
+            ByteBuffer.wrap(bytes, 32, 16),
+            ByteBuffer.wrap(bytes, 48, 13));
+    ImmutableList<Long> flushes =
+        ImmutableList.of(
+            3L, 6L, 9L, 12L, 15L, 18L, 21L, 24L, 27L, 30L, 33L, 36L, 39L, 42L, 45L, 48L, 51L, 54L,
+            57L, 60L, 61L);
+    String z = "[0x00000010 * 0x00000003, 0x0000000d]";
+    WriteOps expected = new WriteOps(bytes, 3, 16, writes, flushes, z);
+    WriteOps actual = WriteOps.of(61, 3, 16);
```
frankyn, line 322: I'm going to ask for variable for these `constants` at least; the others are clear from the data pipeline above.

It would help to know that `61` is data size, `3` is buffer size, and `16` is writeSize without having to pull up definition of WriteOps.

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
BenWhitehead: Close, this is actually limiting how many bytes will be consumed when the flush happens. The flush threshold itself is still determined based upon the size of the buffer, in this case 5.

I've added several comments around these operations to hopefully help follow the flow.

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
BenWhitehead: The overlap between `flush 3` and `flush 4`. I've added some more comments around this flow to help illustrate.

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
BenWhitehead: There are individual partial flushes, but we still end up with all the data. For a dataSize of 27 and buffer size of 10, we should be able to consume everything in three flushes, however due to two of the flushes being partial we end up need 4 flushes, with some overlap of ranges.

frankyn (approved):

PR #1467, opened 2022-06-22
chore: refactor Buffered Channels such that buffer allocation is now lazy

BenWhitehead:
Eagar allocation of large buffers can cause significant memory pressure for users
who are processing many objects in parallel or processing small objects. Since our
Buffered Channels are present to smooth out rpc sizes which can have high latencies
we do not want to unnecessarily allocate a buffer.

Buffered Read & Write Channel already minimize unnecessary copies into our buffer,
coupled with this change, if a user is reading/writing buffers the same size as the
configured buffer size we will never actually allocate a buffer thereby saving on
unnecessary memory consumption.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
@@ -56,10 +56,11 @@ public final class DefaultBufferedWritableByteChannelTest {
   @Property
   void bufferingEagerlyFlushesWhenFull(@ForAll("WriteOps") WriteOps writeOps) throws IOException {
     ByteBuffer buffer = ByteBuffer.allocate(writeOps.bufferSize);
+    AuditingBufferHandle handle = new AuditingBufferHandle(BufferHandle.handleOf(buffer));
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
```diff
@@ -56,10 +56,11 @@ void edgeCases() {
   @Property
   void bufferingEagerlyFlushesWhenFull(@ForAll("WriteOps") WriteOps writeOps) throws IOException {
     ByteBuffer buffer = ByteBuffer.allocate(writeOps.bufferSize);
+    AuditingBufferHandle handle = new AuditingBufferHandle(BufferHandle.handleOf(buffer));
```
sydney-munro, line 59: nit:could we just allocate the BufferHandle directly instead of doing a handleOf?

sydney-munro (approved):

google-cloud-storage/src/test/java/com/google/cloud/storage/DefaultBufferedWritableByteChannelTest.java
BenWhitehead: Using the factory method rather than the constructor directly affords us an easy path to changing the actual implementing class if/when we need to.

PR #1470, opened 2022-06-23
feat: Custom Placement Config Dual Region Support

### LINKED ISSUE #1428: Add Support for Updated Dual-Region Bucket API
[Internal]
- For details and coordination: [go/gcs-dpe-dual-region-milestones-api-refactor](http://go/gcs-dpe-dual-region-milestones-api-refactor)
- Implementation Example: https://github.com/googleapis/nodejs-storage/pull/1977

sydney-munro:

Fixes #1428  ☕️

Will follow up with sample update.


danielbankhead (approved):
Thanks! LGTM

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -103,6 +103,7 @@ public com.google.api.services.storage.model.Bucket apply(BucketInfo bucketInfo)
   private final IamConfiguration iamConfiguration;
   private final String locationType;
   private final Logging logging;
+  private final CustomPlacementConfig customPlacementConfig;
```
BenWhitehead, line 106: How does custom placement differ from `location` above?

Should we refactor away location to always use this multi-region type thing with a single element?

BenWhitehead (changes requested):
Looks like some tests are failing and need to be fixed before merging.

I had one question above for BucketInfo.

@danielbankhead Do you know if this same change is going to be made for the gRPC protos?

We need to keep the protos in mind when designing our new model classes.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
danielbankhead: Good catch, both `location` and `customPlacementConfig` will need to be specified for (custom) dual-regions. Here's an example:
- `location` = 'US'
- `customPlacementConfig` = 'US-EAST-1' and 'US-WEST-1'

danielbankhead:
@sydney-munro, @BenWhitehead this PR should be testable/verifiable after a config push deployment (ETA July 1).

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
danielbankhead: This should be made for the gRPC protos as well.

BenWhitehead: @sydney-munro and @frankyn 

Given the information from Dan above: Do we want to introduce an actual higher level data type which can model the `location` vs `location + customPlacementConfig`, where our BucketInfo would then use this new type? Or, do we hope any possible error message from the backend helps folks figure out they need to be set together?

sydney-munro: I spoke with Dan offline and the backend should be handling this. I think the main reason to not do this is that using location does not necessarily mean you need customPlacementConfig.

BenWhitehead: I was thinking of it from the other direction. If I'm using custom placement config I would guess that I choose either `location` or `custom placement config`, not choose either `location` or `location and custom placement config`.

sydney-munro: As I was working on coding this I do have a concern around this as a breaking change, do we want people that don't care about multiple regions that are using location to have to worry about changing their code just to be able to use the latest client?

BenWhitehead: sync'd offline. we're going to go ahead with the initial proposal, deferring this new type to a later time if we need to.

BenWhitehead (approved):
Feel free to merge when tests are all passing.

PR #1472, opened 2022-06-24
chore: make GapicUnbufferedReadableByteChannel lazily open the server stream

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
```diff
@@ -92,7 +87,12 @@ public long read(ByteBuffer[] dsts) throws IOException {
   @Override
   public long read(ByteBuffer[] dsts, int offset, int length) throws IOException {
     if (complete && open) {
-      close();
+      if (blobOffset != totalSize) {
```
frankyn, line 90: When would this happen?

google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
```diff
@@ -199,7 +210,63 @@ public long read() {
...
+
+    @Override
+    public void close() {
+      if (serverStream != null) {
+        // todo: do we need to "drain" anything?
+        serverStream.cancel();
+      }
+    }
+
+    private void checkOpen() {
```
frankyn, line 243: this is named `checkOpen` but it mutates state, should it be something like `openStream()`

frankyn (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
BenWhitehead: If there were some sort of corruption, or stream truncation that didn't result in a premature eof error.

Seems like a fairly simple check we can do regardless of whether crc32c validation is enabled or not.

google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
BenWhitehead: It will only open the stream if the stream isn't already open. Most of the time, the calls to this method will end up immediately returning.

What do you think of `ensureOpen()` instead of `checkOpen()`?

frankyn: `ensureOpen()` sounds good to me

google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
frankyn: I'm assuming we would retry this failure, because it throws an IOException?

frankyn: Nah it's fine, as long as we retry it then it's okay.

BenWhitehead: 412 Isn't retriable right now. So, I'm going to remove this for now and we can add it back in when we come back and do retries.

frankyn (approved):

PR #1474, opened 2022-06-27
chore: refactor GrpcStorageImpl#read* methods to use more specific channels

BenWhitehead:
Update readAllBytes and downloadTo methods to use unbuffered read channel, instead of reader

Update *ReadableByteChannelSessions to take a ReadObjectRequest instead of Object

Update Hasher#validate to be lazy with regard to the ByteBuffer, this ensures no new byte buffer is created if it isn't needed in the case of noop.

Add StorageArbitraries.Objects#name()
Move StorageArbitraries.bucketName() -> StorageArbitraries.Buckets#name()
Add StorageArbitraries.alnum() and cleanup usages in Buckets.name()

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -30,7 +30,6 @@ import com.google.api.gax.rpc.ApiExceptions;
 import com.google.api.gax.rpc.UnaryCallable;
 import com.google.cloud.BaseService;
 import com.google.cloud.Policy;
-import com.google.cloud.ReadChannel;
 import com.google.cloud.WriteChannel;
 import com.google.cloud.storage.Acl.Entity;
 import com.google.cloud.storage.Conversions.Decoder;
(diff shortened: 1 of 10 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/StorageV2ProtoUtils.java
```diff
@@ -0,0 +1,95 @@
...
+      checkArgument(0 <= limit, VALIDATION_TEMPLATE, offset, limit);
+    }
+
+    if (!offsetNull && !limitNull) {
+      checkArgument(offset <= limit, VALIDATION_TEMPLATE, offset, limit);
+    }
+  }
+
+  @NonNull
+  static String fmtProto(@NonNull final MessageOrBuilder msg) {
```
frankyn, line 79: What's wrong with using proto helper: https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/util/JsonFormat.Printer

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -936,4 +983,45 @@ private static String fmtMethodName(String name, Class<?>... args) {
...
+    ifNonNull(
+        (Long) optionsMap.get(StorageRpc.Option.IF_GENERATION_MATCH),
+        builder::setIfGenerationMatch);
+    ifNonNull(
+        (Long) optionsMap.get(StorageRpc.Option.IF_GENERATION_NOT_MATCH),
+        builder::setIfGenerationNotMatch);
+    String userProject = (String) optionsMap.get(StorageRpc.Option.USER_PROJECT);
+    if (userProject != null) {
+      // TODO: user project
+      // ReadObjectRequest does not have a user_project field
```
frankyn, line 1013: IIUC, we need to use gRPC Metadata for this now per cl/448634966

I see: https://g3doc.corp.google.com/net/grpc/g3doc/java/life_of_a_stream_java.md?cl=head#clientinterceptor 

Looks like we need to use a client side interceptor but couldn't figure out where to attach it. Maybe you'll have better luck.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -445,27 +463,56 @@ public GrpcBlobReadChannel reader(String bucket, String blob, BlobSourceOption..
...
 
   @Override
   public void downloadTo(BlobId blob, OutputStream outputStream, BlobSourceOption... options) {
-    try (WritableByteChannel out = Channels.newChannel(outputStream);
-        GrpcBlobReadChannel in = reader(blob, options)) {
-      ByteStreams.copy(in, out);
+
+    final Map<StorageRpc.Option, ?> optionsMap = StorageImpl.optionMap(options);
+
+    // TODO: handle StorageRpc.Option.RETURN_RAW_INPUT_STREAM impacts
```
frankyn, line 500: gRPC will always provide compressed data; maybe we can decompress data for the user to match up with the existing JSON API experience

frankyn (approved):

PR #1483, opened 2022-06-30
feat: implement update(BucketInfo, BucketTargetOption...): Bucket

frankyn:
Implement `update(BucketInfo, BucketTargetOption...): Bucket` and also fix a gotcha in create bucket not passing along metadata from bucketInfo.

PR #1484, opened 2022-07-01
feat: implement GrpcStorageImpl#update(BlobInfo): Blob

sydney-munro:


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -422,12 +424,45 @@ public Bucket update(BucketInfo bucketInfo, BucketTargetOption... options) {
...
+    ifNonNull(
+        (Long) optionsMap.get(StorageRpc.Option.IF_METAGENERATION_MATCH),
+        updateRequestBuilder::setIfMetagenerationMatch);
+    ifNonNull(
+        (Long) optionsMap.get(StorageRpc.Option.IF_METAGENERATION_NOT_MATCH),
+        updateRequestBuilder::setIfMetagenerationNotMatch);
+    ifNonNull(
+        (String) optionsMap.get(StorageRpc.Option.PREDEFINED_ACL),
+        updateRequestBuilder::setPredefinedAcl);
+    updateRequestBuilder.setUpdateMask(
```
sydney-munro, line 447: grabbed this from https://github.com/googleapis/java-storage/pull/1483/files so @frankyn comment is valid here as well. not sure if this will work as expected. We will need to generate these masks for quite a few operations so maybe we can make this a helper method?

sydney-munro: done.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -1154,4 +1183,15 @@ private CommonObjectRequestParams commonRequestParams(String key) {
...
+  private FieldMask fieldMaskGenerator(
+      Map<FieldDescriptor, java.lang.Object> fieldDescriptorObjectMap) {
+    return FieldMask.newBuilder()
+        .addAllPaths(
+            fieldDescriptorObjectMap.entrySet().stream()
+                .filter(x -> x.getValue() != null)
+                .map(e -> e.getKey().getName())
+                .collect(Collectors.toList()))
+        .build();
+  }
```
BenWhitehead, line 1196: Adding this helper is great, since we'll need to do the same thing for each of the update methods.

We can encapsulate a step further though, to simplify the usage of it.

In protobuf, all of our generated classes extends from `com.google.protobuf.Message`, meaning we can change the method signature to take in any `Message` and yield the FieldMask.

```suggestion
  private FieldMask fieldMaskGenerator(Message message) {
    return FieldMask.newBuilder()
        .addAllPaths(
            message.getAllFields().entrySet().stream()
                .filter(x -> x.getValue() != null)
                .map(e -> e.getKey().getName())
                .collect(Collectors.toList()))
        .build();
  }
```

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -422,12 +425,38 @@ public Bucket update(BucketInfo bucketInfo, BucketTargetOption... options) {
...
+    ifNonNull(
+        (Long) optionsMap.get(StorageRpc.Option.IF_METAGENERATION_MATCH),
+        updateRequestBuilder::setIfMetagenerationMatch);
+    ifNonNull(
+        (Long) optionsMap.get(StorageRpc.Option.IF_METAGENERATION_NOT_MATCH),
+        updateRequestBuilder::setIfMetagenerationNotMatch);
+    ifNonNull(
+        (String) optionsMap.get(StorageRpc.Option.PREDEFINED_ACL),
+        updateRequestBuilder::setPredefinedAcl);
+    updateRequestBuilder.setUpdateMask(fieldMaskGenerator(object.getAllFields()));
```
BenWhitehead, line 448: This line will need to change after the small refactor to the helper method.

```suggestion
    updateRequestBuilder.setUpdateMask(fieldMaskGenerator(object));
```

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
sydney-munro: thanks! yep agreed this is simpler.

BenWhitehead (approved):
Thanks!

frankyn (commented):
Thanks for making this easier @sydney-munro!

PR #1498, opened 2022-07-08
feat: implement GrpcStorageImpl#compose(ComposeRequest): Blob

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -44,12 +44,16 @@ import com.google.cloud.storage.spi.v1.StorageRpc;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import com.google.common.hash.HashCode;
 import com.google.common.hash.Hashing;
 import com.google.common.io.ByteStreams;
 import com.google.protobuf.ByteString;
 import com.google.storage.v2.BucketName;
 import com.google.storage.v2.CommonObjectRequestParams;
+import com.google.storage.v2.ComposeObjectRequest;
+import com.google.storage.v2.ComposeObjectRequest.SourceObject;
 import com.google.storage.v2.CreateBucketRequest;
 import com.google.storage.v2.DeleteBucketRequest;
 import com.google.storage.v2.DeleteHmacKeyRequest;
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -502,7 +506,40 @@ public boolean delete(BlobId blob) {
...
   @Override
   public Blob compose(ComposeRequest composeRequest) {
-    return todo();
+    final Map<StorageRpc.Option, ?> optionsMap =
+        StorageImpl.optionMap(Iterables.toArray(composeRequest.getTargetOptions(), Option.class));
+    GrpcCallContext grpcCallContext = GrpcRequestMetadataSupport.create(optionsMap);
+    ComposeObjectRequest.Builder composeObjectReqBuilder = ComposeObjectRequest.newBuilder();
+    final List<SourceObject> sources =
+        Lists.newArrayListWithCapacity(composeRequest.getSourceBlobs().size());
+    for (ComposeRequest.SourceBlob sourceBlob : composeRequest.getSourceBlobs()) {
```
sydney-munro, line 515: Concern here:
ComposeRequest does not have options for the source object, yet the ComposeObjectRequest has a field for ObjectPrecondition that apply to the source, the only condition listed is if_generation_match.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -502,7 +506,40 @@ public boolean delete(BlobId blob) {
...
+      sources.add(
+          SourceObject.newBuilder()
```
sydney-munro, line 517: I did not think this was worth writing a codec but could be convinced otherwise.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -502,7 +506,40 @@ public boolean delete(BlobId blob) {
...
+        composeObjectReqBuilder::setIfGenerationMatch);
+    ifNonNull(
+        (Long) optionsMap.get(StorageRpc.Option.IF_METAGENERATION_MATCH),
+        composeObjectReqBuilder::setIfMetagenerationMatch);
+    ifNonNull(
+        (String) optionsMap.get(StorageRpc.Option.PREDEFINED_ACL),
+        composeObjectReqBuilder::setDestinationPredefinedAcl);
+    ifNonNull(
+        (String) optionsMap.get(StorageRpc.Option.KMS_KEY_NAME),
+        composeObjectReqBuilder::setKmsKey);
```
BenWhitehead, line 536: I think we also need to account for encryption options here, similar to https://github.com/googleapis/java-storage/blob/dc78526343c30ea85fd05c6977696309d687d958/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java#L1181-L1184

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -502,7 +506,40 @@ public boolean delete(BlobId blob) {
 
   @Override
   public Blob compose(ComposeRequest composeRequest) {
```
BenWhitehead, line 508: Can you also update the annotation in Storage.java to include `Transport.GRPC` in the supported transports for compose? https://github.com/googleapis/java-storage/blob/dc78526343c30ea85fd05c6977696309d687d958/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L2499

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
sydney-munro: works for me 👍

BenWhitehead (approved):

PR #1505, opened 2022-07-12
chore: introduce ZOpt

BenWhitehead:
StorageRpc.Option is a difficult and often error prone to use, ZOpt tries to make it easier and safer to work with.

ZOpt defines a set of constants matching those from StorageRpc.Option, except that ZOpt carries the generic type information with it rather than dropping it. This makes usage of these options easier and type safe.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/ZOptTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/ZOptTest.java
@@ -20,8 +20,13 @@ import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.fail;
 
 import com.google.cloud.storage.spi.v1.StorageRpc;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import java.lang.reflect.Field;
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/ZOptTest.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Collections;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+import org.junit.Test;
+
+public final class ZOptTest {
+
+  @Test
+  public void happyPath() {
```
frankyn, line 33: What do you think about adding a test to check that all StorageRpc.Option.* values are handled by ZOpt? 

It will help when we introduce new options in the future; It's probably going to be a "duh" moment, but it will be helpful to make sure there are no orphens introduced.

frankyn (commented):
Overall, LGTM and thanks for making it easier. I just had one question

frankyn (approved):

PR #1510, opened 2022-07-14
test: update testbench to support grpc endpoint

frankyn:
Support grpc api testing with testbench

blocked until https://github.com/googleapis/storage-testbench/pull/376 is released.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
@@ -36,7 +36,9 @@ import org.junit.Test;
 public final class ITGrpcTest {
   private static final Logger LOGGER = Logger.getLogger(ITGrpcTest.class.getName());
 
-  @ClassRule public static final TestBench TEST_BENCH = TestBench.newBuilder().build();
+  @ClassRule
+  public static final TestBench TEST_BENCH =
+      TestBench.newBuilder().setContainerName("it-grpc").build();
 
   @Rule
   public final StorageFixture storageFixture =
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
@@ -27,8 +27,6 @@ import com.google.cloud.storage.StorageOptions;
 import com.google.cloud.storage.conformance.retry.TestBench;
 import com.google.cloud.storage.testing.RemoteStorageHelper;
 import java.util.logging.Logger;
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.logging.Logger;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+public final class ITGrpcTest {
+  private static final Logger LOGGER = Logger.getLogger(ITGrpcTest.class.getName());
+  private static final String TESTBENCH_GRPC_URI = "http://localhost:9001";
```
BenWhitehead, line 38: I think this is unnecessary, since it matches the default

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
```diff
@@ -0,0 +1,76 @@
...
+  @ClassRule
+  public static final TestBench TEST_BENCH =
+      TestBench.newBuilder().setGRPCBaseUri(TESTBENCH_GRPC_URI).build();
+
+  @Rule
+  public final StorageFixture storageFixture =
+      StorageFixture.from(
+          () ->
+              StorageOptions.grpc()
+                  .setHost(TESTBENCH_GRPC_URI)
```
BenWhitehead, line 49: This can be `TEST_BENCH.getGRPCBaseUri()`. The actual reading of this value will happen after TestBench has already booted.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
```diff
@@ -0,0 +1,76 @@
...
+    LOGGER.fine("Running setup complete");
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    LOGGER.fine("Running teardown...");
+    storage.close();
+    LOGGER.fine("Running teardown complete");
+  }
+
```
BenWhitehead, line 69: This setup and teardown is already taken care of within the `StorageFixture` we can remove it.

BenWhitehead (changes requested):
The integration tests appear to be failing due to not being able to connect to the grpc endpoint.

In the past we've seen this because the port was explicitly bound to `localhost` which prevents it being forwarded to from outside docker. Can you double check that the grpc endpoint is listening on `0.0.0.0`?

BenWhitehead (approved):
Synced offline. Clear to merge after cleanup of left over debug changes.

PR #1521, opened 2022-07-21
test: add BucketFixture

BenWhitehead:
Create a JUnit Rule which can manage the lifecycle of a temporary bucket. A bucket will be created by the rule, and then emptied and deleted upon test completion.

Update some of the IT*Test classes to use the new fixture when applicable.

Update bucket name handling in some GrpcStorageImpl methods.

Add simplistic listBlobs and listBuckets tests to ITGrpcTest

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBlobReadChannelTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBlobReadChannelTest.java
@@ -22,11 +22,9 @@ import com.google.cloud.ReadChannel;
 import com.google.cloud.WriteChannel;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
-import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.BucketFixture;
 import com.google.cloud.storage.DataGeneration;
-import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageFixture;
-import com.google.cloud.storage.testing.RemoteStorageHelper;
 import com.google.common.io.ByteStreams;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBlobReadChannelTest.java
```diff
@@ -54,26 +52,23 @@ public final class ITBlobReadChannelTest {
...
+  public static final BucketFixture bucketFixture =
+      BucketFixture.newBuilder().setHandle(storageFixture::getInstance).build();
 
-  private Storage storage;
   private String bucketName;
   private String blobName;
 
   @Before
   public void setUp() throws Exception {
-    storage = storageFixture.getInstance();
```
sydney-munro, line 65: nit: I don't see the value of offloading this to all places that require an instance of storage.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
```diff
@@ -18,24 +18,29 @@
...
 
 public final class ITGrpcTest {
-  @ClassRule
+  @ClassRule(order = 1)
   public static final TestBench TEST_BENCH =
       TestBench.newBuilder().setContainerName("it-grpc").setDockerImageTag("v0.26.0").build();
 
-  @Rule
-  public final StorageFixture storageFixture =
+  @ClassRule(order = 2)
```
sydney-munro, line 42: this is more a question than a comment, what is the benefit to using ClassRule over Rule here?

BenWhitehead: `@ClassRule` will bind to the lifecycle of the class, whereas `@Rule` will bind to the lifecycle of the test.

In this case, the storage client can live as long as the testbench since we're not testing anything on the construction/setup of the storage instance.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBlobReadChannelTest.java
BenWhitehead: I can back this change out then.

sydney-munro (approved):

PR #1531, opened 2022-08-01
test: add regression tests for each Option on our public api

BenWhitehead:
*  add tests for metageneration{Not,}Match(long)
*  add tests for {en,de}cryptionKey({Key,String})
*  add tests for userProject

   * Bucket.BucketSourceOption#userProject is somehow busted, and doesn't work for exists/refresh/delete due to some enum exception

*  add tests for predefined{DefaultObject,}Acl
*  add tests for disableGzipContent()
*  add tests for generation{Not,}Match and doesNotExist
*  add tests for detectContentType
*  add tests for kmsKeyName
*  add tests for projection
*  add tests for requestedPolicyVersion
*  add tests for projectId
*  add tests for list operations
*  add tests for fields(*) options
*  add tests for BlobWriteOption#{md5,crc32c}Match
*  add tests for {meta,}generation{Not,}Match extractors

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -324,6 +324,7 @@ public class HttpStorageRpc implements StorageRpc {
           .setProjection(DEFAULT_PROJECTION)
           .setPredefinedAcl(Option.PREDEFINED_ACL.getString(options))
           .setPredefinedDefaultObjectAcl(Option.PREDEFINED_DEFAULT_OBJECT_ACL.getString(options))
+          .setUserProject(Option.USER_PROJECT.getString(options))
           .execute();
     } catch (IOException ex) {
       span.setStatus(Status.UNKNOWN.withDescription(ex.getMessage()));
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -324,6 +324,7 @@ public Bucket create(Bucket bucket, Map<Option, ?> options) {
           .setProjection(DEFAULT_PROJECTION)
           .setPredefinedAcl(Option.PREDEFINED_ACL.getString(options))
           .setPredefinedDefaultObjectAcl(Option.PREDEFINED_DEFAULT_OBJECT_ACL.getString(options))
+          .setUserProject(Option.USER_PROJECT.getString(options))
```
frankyn, line 327: No way, the auditing caught that userProject wasn't set for this calls?

frankyn (approved):
LGTM, love the auditing helper!

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
BenWhitehead: Indeed it did.

PR #1546, opened 2022-08-04
chore: introduce new UnifiedOpt internal api

BenWhitehead:
Attempt to create a single unified api for the various types of options
to remove the need for cross translation, duplication and extraneous
conflation.

*NOTE* For now this api is internal only, however when a binary
compatibility boundary is able to be crossed we could make parts of it
public and replace the existing Option classes.

## New Approach

The general idea of the new approach is one of encapsulated
transformation to specific types of containers.

When handling an option, we need to take into consideration at least
each of the following:
1. Apiary client
2. com.google.cloud.storage.* Model classes
3. gRPC client
4. Maintainability of Options
5. Cognitive load keeping track of options
6. Lifecycle of instances of Options

In particular, 4 and 5 above have been non-trivial to us and customers
both in the past. Specifically, there are three different classes
 related to reading a blob (`Storage.BlobSourceOption`,
`Storage.BlobGetOption`, `Blob.BlobSourceOption`) and three different
classes related to writing/updating a blob. Often highly duplicated, but
not actually commonly defined/maintained. Additionally, each kind of
list operation has it's own Option, duplicated again.

This new change leverages the fact that multiple inheritance of java
interfaces is possible. As such, we can define a set of interfaces which
relate to different resource types (Bucket, Blob, HmacKey) and interfaces
which relate to operation types (Source, Target, List) and applies these
interfaces to those options for which they apply.

Then, we can define the necessary transformations an Option needs to
provide in order to operate in the above circumstances.

### Implementation
Each parameter from https://cloud.google.com/storage/docs/json_api/v1/parameters
has a corresponding concrete class definition in UnifiedOpts. Each class
binds a provided value to the corresponding StorageRpc.Option when applicable.

Some "Options" are actually mutations of BlobInfo before it is translated
to its transport specific representation. Several of these are confusing,
and are present for the sake of backward compatibility but should be
removed/excluded from any future public API modification.

Some "Options" are actually value extractors from a BlobId/BlobInfo/BucketInfo
several of these are confusing, and can even be mutually exclusive with
the other behavior of the operation (performing an object get which extracts
the generation value adding `ifGenerationNotMatch` sending a request to GCS
which can never succeed).

#### Serialization
All existing Options are Serializable meaning our new Opt classes also need
to be Serializable. Each Opt property defines equals/hashCode along with
serialVersionUID in order to provide compatibility.

#### ProjectAsSource
Some options - Generation, Metageneration - have corresponding "SOURCE"
versions. These are important in the case of operations like rewrite. As such,
an interface `ProjectAsSource` has been introduced which can be applied to an
Opt to generate the corresponding "SOURCE" Opt when necessary.

#### Extractors
Parameter extractors are implemented as an Opt which produces another Opt
when provided with an instance of some Model class. For example,
`MetagenerationMatchExtractor` will produce an instance of
`MetagenerationMatch` when resolved against a `BlobInfo` or `BucketInfo`.

Each extractor is itself a singleton, as they are "pure" and always produce
a value relative to the input provided to one of its method invocations.
To ensure each extractor is a singleton, integration with Java Serialization
has been done via `readResolve()` which will return the singleton instance
when attempting to deserialize.

### Binary Compatibility
Leave all existing Option classes exposed as they are, however each class
will function as a facade for one of the Opt instance for the
corresponding new Unified Opt.

### Implementation Compatibility
Several of the existing Option implementations actually perform extractions
or mutations to instances of BucketInfo or BlobInfo between method invocation.
Some of these have been preserved, but are also marked `@Deprecated` where
these Options can lead to confusing semantics (BlobId with generation set,
then also specifying generationNotMatch()). Removing these confusing Options
will require a binary break and updated docs/samples to show customers how
they can accomplish a clear semantic operation.

There is a rough new multi-phase pattern used with the new options and flows
like this:
1. Create an new instance of Opts from the existing legacy Option classes
2. With the result from 1, resolve any extractor Opt to it's corresponding
   concrete instance
3. If applicable, apply any modification back to the "Info" class
   (This should be removed in the future, but is the current way someone can
   define md5, crc32c or automatic content type based on object name)
4. Have the result from 2 create a Map<StorageRpc.Option, ?> which can be
   passed to StorageRpc methods
5. For gRPC implementation, instead of using the map like in 4, we can
   define mapping methods to operate on the respective request builder

## Impacted Code changes

1. Completely gut com.google.cloud.storage.Option, this class is left only
   for binary compatibility reasons.
2. Move behavior of clearing md5 and crc32c from a BlobInfo out of translate
   and into each applicable method. This is done for behavioral compatibility
   and should be removed at such time as we can release breaking changes.
   * If a customer sets their md5/crc32c before attempting a create or write
     the value will be removed, only added back if a corresponding `md5Match`
     or `crc32cMatch` is provided.
3. Many many test were reaching down into Option fields. Each test doing this
   has been updated to instead provide the key/value they are validating
   explicitly.
4. In `Bucket` and `Blob` some operations were mutating the respective Info
   before passing it to `Storage`, this has been removed so that any Option
   processing logic which can modify an Info is centralized in the `Storage`
   instance.
5. In `Bucket` and `Blob` some level of Option validation was being performed
   before calling the `Storage` operation, this has been removed so that
   all Option validation can be performed in the `Storage` instance.

### Test Cleanup

1. Cleanup Bucket related tests which were hardcoded to expect modifying
   generation/metageneration of BlobInfo before passed to storage
2. Delete several test from BucketTest which were performing early validation
   unnecessarily.
3. Delete MANY unused constants from various test classes


frankyn (commented):
@BenWhitehead could you write a design doc so we can review the change as a team? 

It's a significantly sized change but I'm guessing the crux of the change is relatively short.
It would help review this PR and future us understand the change.

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -0,0 +1,1438 @@
...
+
+  static final class ServiceAccount extends RpcOptVal<String> implements HmacKeyListOpt {
+    private static final long serialVersionUID = 1630581690347694016L;
+
+    private ServiceAccount(String val) {
+      super(StorageRpc.Option.SERVICE_ACCOUNT_EMAIL, val);
+    }
+  }
+
+  static final class SetContentType implements ObjectTargetOpt {
```
frankyn, line 804: We have a lot of boiler plate in these classes; would it be reasonable to introduce Autoclass?

google-cloud-storage/src/main/java/com/google/cloud/storage/Option.java
```diff
@@ -16,53 +16,17 @@
...
-
-  StorageRpc.Option getRpcOption() {
-    return rpcOption;
-  }
-
-  Object getValue() {
-    return value;
-  }
+@Deprecated
+public abstract class Option<O extends Opt> extends UnifiedOpts.OptionShim<O>
```
frankyn, line 24: I was wondering how Option<T> was actually going to work; lucky for us it's a class in our library.

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -182,9 +176,11 @@ public Blob create(
   @Override
   @Deprecated
   public Blob create(BlobInfo blobInfo, InputStream content, BlobWriteOption... options) {
-    Tuple<BlobInfo, BlobTargetOption[]> targetOptions = BlobTargetOption.convert(blobInfo, options);
-    StorageObject blobPb = codecs.blobInfo().encode(targetOptions.x());
-    Map<StorageRpc.Option, ?> optionsMap = optionMap(targetOptions.x(), targetOptions.y());
+    Opts<ObjectTargetOpt> opts = Opts.unwrap(options).resolveFrom(blobInfo);
+    Map<StorageRpc.Option, ?> optionsMap = opts.getRpcOptions();
+    BlobInfo.Builder builder = blobInfo.toBuilder().setMd5(null).setCrc32c(null);
+    BlobInfo updated = opts.blobInfoMapper().apply(builder).build();
```
frankyn, line 182: This is a bit nuanced for me; could you clarify what this apply is doing the BlobInfo builder?

BenWhitehead: Agreed, I was surprised when I ran across the need for it too.

Previously, when converting between certain options (for example[1]) there were actually mutations applied to the BlobInfo.

Instead of doing the same mutation during conversion, we now have a two phase approach. 1) Resolve any extracting Opts against the Info type using `resolveFrom(info)` 2) then applying any mutations to the Info before encoding via the codec.

My hope is that we'll be able to move away from these extractor types of options as they are prone to confusion (Construct/Load a BlobInfo with generation, then specify `generationNotMatch()` resulting in a url like `?generation=3&ifGenerationNotMatch=3` which can't ever succeed). But, for the time being we need them in order to preserve the existing behavior.

[1] https://github.com/googleapis/java-storage/blob/2db5f4a4ebf001248462553a0eb7d4f4399ddfe1/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L571-L589

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
BenWhitehead: Using AutoValue would prevent us from defining final classes which I really don't want to do on a (likely future) public api.

The majority of our `Opt`s extend off `RpcOptVal` so they only need a constructor, it's mainly these more esoteric extractor/mutation `Opt`s which are needing their own `equals` and `hashCode`.

google-cloud-storage/src/main/java/com/google/cloud/storage/Option.java
BenWhitehead: Yeah, thankfully it's not in core

frankyn (approved):

PR #1558, opened 2022-08-08
chore: update double check lazy initialization idiom

BenWhitehead:
Update location of double check lazy initialization idiom to follow the flow
outlined in Effective Java, 3rd Edition Item 83


google-cloud-storage/src/main/java/com/google/cloud/storage/BufferHandle.java
```diff
@@ -53,6 +53,9 @@ static final class LazyBufferHandle extends BufferHandle {
     private final int capacity;
     private final Function<Integer, ByteBuffer> factory;
 
+    // It is theoretically possible for this value to be null for any of the methods, while
+    // get is invoked. When ever reading this value, always read into a local variable and then
```
JesseLovelace, line 57: I think you should use "whenever" instead of "when ever here"

JesseLovelace (approved):

PR #1563, opened 2022-08-09
chore: preliminary implementation of gRPC status code to HTTP Status code translation

BenWhitehead:
Update StorageException#coalesce to account for subclasses of ApiException returned by the gapic client.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageException.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageException.java
@@ -25,9 +25,11 @@ import com.google.cloud.BaseServiceException;
 import com.google.cloud.RetryHelper.RetryHelperException;
 import com.google.cloud.http.BaseHttpServiceException;
 import com.google.common.collect.ImmutableSet;
+import io.grpc.StatusException;
+import io.grpc.StatusRuntimeException;
 import java.io.IOException;
 import java.util.Set;
-import java.util.function.Supplier;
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/StorageException.java
```diff
@@ -91,13 +95,30 @@ private static StorageException getStorageException(Throwable t) {
...
+      // https://cloud.google.com/apis/design/errors#error_payloads
+      // TODO: flush this out more to wire through "error" and "details"
+      int httpStatusCode = 0;
+      StatusCode statusCode = apiEx.getStatusCode();
+      if (statusCode instanceof GrpcStatusCode) {
+        GrpcStatusCode gsc = (GrpcStatusCode) statusCode;
+        httpStatusCode =
+            BackwardCompatibilityUtils.grpcCodeToHttpStatusCode(gsc.getTransportCode());
+      }
+      String message = firstNonNull(() -> getCauseMessage(apiEx), apiEx::getMessage);
```
sydney-munro, line 118: one of these is guaranteed to be non null? if not we could risk throwing an illegal state exception instead of the StorageException we plan on constructing.

BenWhitehead: You're correct, I've changed it so our base case is a generic message we construct.

sydney-munro (approved):

PR #1566, opened 2022-08-11
docs: Document thread safety of library

JesseLovelace:
Internal bug 216304555, documents the thread safety of this library


BenWhitehead (approved):
I'm hoping to address the lack of thread safety for Read/Write Channel along with the gRPC work. If that works, we can try to backport the same approach to the JSON implementation.

PR #1568, opened 2022-08-11
chore: remove ZOpt replacing with UnifiedOpts

BenWhitehead:
Update all usage of UnifiedOpts in GrpcStorageImpl to instead use Request mappers rather than ZOpt.

Add suite of unit tests for new gRPC request mappers.

Follow up of #1546


frankyn (approved):
@BenWhitehead and I discussed this PR at length due the extensive redesign.

A few key points that weren't clear to me at first that help:

1. Options are consolidated into a cross product of Options and Resource Types
2. Reduction of the problem to resource types instead of requests allows for association per request + resource type; because requests (action on a resource) can have different options enabled.
3. Now that the options are consolidated, introduction of new options is much cleaner and less error prone because options are maintained in a Unified map.
4. Options are executed as a pipeline modifying request payload and headers (proto , metadata)
5. UnifiedOpts also rely on compile time checking

PR #1571, opened 2022-08-12
chore: implement retries for List operations in GrpcStorageImpl

BenWhitehead:
Update TransformPageDecorator to apply automatic retires to individual page fetches.

To accomplish page granularity for retries I've replaced TransformingIterator with
a page producing Stream, where each page can then be expected into a single Iterator.

Add new interface Retrying.RetryingDependencies to provide a marker for the need
of RetrySettings and an ApiClock so we're not coupled to full StorageOptions
instances.

Renamed Retrying.run(HttpStorageOptions, ...) to Retrying.runHttp(HttpStorageOptions, ...)
to remove ambiguity. Ideally we'll be able to remove this in the future once grpc
and http are both able to use the same approach for configuring gccl-invocation-id.

Refactor much of GrpcTransformPageDecoratorTest.java to provide a data driven
implementation of UnaryCallable removing the need to track individual request counts.
Rather than using String for RequestT and ResponseT I've created Req and Resp container
classes to be used and to remove any ambiguity about the directionality of various
operations.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -27,6 +27,7 @@ import com.google.api.core.ApiFuture;
 import com.google.api.gax.grpc.GrpcCallContext;
 import com.google.api.gax.paging.AbstractPage;
 import com.google.api.gax.paging.Page;
+import com.google.api.gax.retrying.ResultRetryAlgorithm;
 import com.google.api.gax.rpc.ApiException;
 import com.google.api.gax.rpc.ApiExceptions;
 import com.google.api.gax.rpc.UnaryCallable;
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -987,43 +1007,84 @@ public String getNextPageToken() {
...
+      return () ->
+          streamIterate(
+                  page,
+                  p -> p != null && p.hasNextPage(),
+                  prev -> {
+                    // explicitly define this callable rather than using the method reference to
+                    // prevent a javac 1.8 exception
+                    // https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8056984
+                    Callable<PageT> c = () -> prev.getNextPage();
+                    return Retrying.run(deps, resultRetryAlgorithm, c, Decoder.identity());
```
frankyn, line 1030: only question i have is; for exceptions thrown from `Retrying.run`, what happens to them? I ask because I don't see `throws Exception` in the method signatures.

BenWhitehead: Exceptions which aren't retryable and/or exhaust the `Retrying.run` will be caught and coalesced to a StorageException.

https://github.com/googleapis/java-storage/pull/1571/files#diff-16ae9534e54393cc0a423731519677b6d8d1dae39e70afe603579abff529860cR85-R96

frankyn: My question is more about checked exception vs. non-checked exception. In this case it would be unchecked exception but not sure why we wouldn't.

BenWhitehead: The retrying code under the hood will catch any checked exception and wrap it.

frankyn (approved):

PR #1585, opened 2022-08-24
test: add first two low level tests for grpc upload

BenWhitehead:
two new tests that validate correct request packing and response handling when
performing a direct or resumable upload.

#### ByteSizeConstants
Create a constants class to keep track of the common byte size values used
throughout the library.

#### GapicUnbufferedWritableByteChannel
Reduce the surface area on the constructor, moving Write callable, Hasher,
ByteStringStrategy, etc. out since GapicUnbufferedWritableByteChannel itself
doesn't use them.

Change ChunkSegmenter to no longer be static, instead taking a handle the
Hasher, ByteStringStrategy and maxSegmentSize which are necessary for later
computations of segmentBuffers.

Change FlusherFactory to no longer take the write callable, instead any
dependency is expected to be bound into the state of the factory. Flushers will
now maintain a handle to the write callable when they are constructed.

#### Hasher
Annotate Hasher#hash(*) methods as nullable to ensure the return value is
checked before dereferencing (via intellij warning, and eventually error prone
when we can wire that in).

#### TestUtils#subDivide(byte[], int)
Create utility method to subdivide a byte array into N ByteBuffers, each with a
window over the provided byte[] with a maximum limit of int bytes.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ByteSizeConstants.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ByteSizeConstants.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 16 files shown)


frankyn (approved):
LGTM

PR #1588, opened 2022-08-26
test: add grpc integration tests for simplistic object read/write

BenWhitehead:
Update BucketFixture to provide its own implementation of bulk object delete.
The method in RemoteStorageHelper relies upon batch methods which aren't
supported for grpc.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketFixture.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/BucketFixture.java
@@ -18,13 +18,21 @@ package com.google.cloud.storage;
 
 import static java.util.Objects.requireNonNull;
 
+import com.google.api.gax.paging.Page;
 import com.google.cloud.storage.conformance.retry.CleanupStrategy;
-import com.google.cloud.storage.testing.RemoteStorageHelper;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
+import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
```diff
@@ -74,13 +89,30 @@ public void testCreateBucket() {
...
+    String prefix = testName.getMethodName();
+    List<Blob> blobs =
+        IntStream.rangeClosed(1, 10)
+            .mapToObj(i -> String.format("%s/%02d", prefix, i))
+            .map(n -> BlobInfo.newBuilder(bucketInfo, n).build())
+            .map(
+                info ->
+                    storageFixture
+                        .getInstance()
+                        .create(info, content, BlobTargetOption.doesNotExist()))
```
frankyn, line 102: It's possbile that this can fail with a 412, can you ignore 412 failures in this lambda?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
```diff
@@ -156,4 +188,53 @@ public void deleteHmacKey() {
     storageFixture.getInstance().updateHmacKeyState(hmacKey.getMetadata(), HmacKeyState.INACTIVE);
     storageFixture.getInstance().deleteHmacKey(hmacKey.getMetadata());
   }
+
+  @Test
+  public void object_writeGetRead() {
```
frankyn, line 193: Are you using snakecase to clearly group test names while the refactoring is on-going?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
```diff
@@ -156,4 +188,53 @@ public void deleteHmacKey() {
...
+    s.create(info, content, BlobTargetOption.doesNotExist());
+
+    Blob blob = s.get(info.getBlobId());
+
+    byte[] actualContent = blob.getContent();
+    assertThat(actualContent).isEqualTo(content);
+  }
+
+  @Test
+  public void objectCreate_storage_create() {
```
frankyn, line 206: do you want to create a consistent prefix string for:
`objectWriteStorageCreate()`
`objectWriteStorageCreateStream()

google-cloud-storage/src/test/java/com/google/cloud/storage/BucketFixture.java
```diff
@@ -94,7 +102,30 @@ public static Builder newBuilder() {
   private void doCleanup(String bucketName, Storage s) {
     LOGGER.info("Starting bucket cleanup...");
     try {
-      RemoteStorageHelper.forceDelete(s, bucketName, 5, TimeUnit.MINUTES);
+      // TODO: probe bucket existence, a bad test could have deleted the bucket
+      Page<Blob> page1 = s.list(bucketName);
```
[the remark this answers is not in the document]
BenWhitehead: Yeah, each `Page` is the first pages results, plus a handle to be able to create the next page (and transparently perform the RPC)

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
BenWhitehead: The only way these preconditions could fail is if the container from TestBench lives longer than it's supposed to. Not sure if it's worth attempting to swallow that kind of failure. WDYT?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
BenWhitehead: Not for the refactoring, simply to group the conceptually related tests.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
frankyn: Fair, could you add a TODO to remember that this can happen when we start hitting the live service?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
BenWhitehead: The intent here was to try and communicate "testing objectCreate through Storage#create"

objectCreate was due to following the fact the current methods are named create.

We can change to `objectWrite` while keeping the other hints if that sounds good to you.

frankyn: Any objection to splitting uploads / downloads IT into separate test files? They're the two main operations we are going to have many IT for.

BenWhitehead: I've been doing that for most of the more detailed read/write tests of the new channels, these tests are mainly more of a "are things plumbed up in GrpcStorageImpl". I think, after the IT refactoring is more settled we can decided if we want to move these.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
BenWhitehead: I also forgot, for this class we're always operating on a temporary bucket, so these objects can't escape to impact another run (unless the same UUID is somehow generated).

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
frankyn: Can we go with your original suggestion of using `objectWrite` while keeping the other hints, which I assume is `storage_create` to qualify that this is for `storage.create()`?

frankyn (approved):

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
BenWhitehead: will do

PR #1590, opened 2022-08-26
fix: update StorageSettings configuration to allow ReadObject streams to use idleTimeout rather than totalTimeout

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -97,35 +99,51 @@ StorageSettings getStorageSettings() throws IOException {
...
+        getRetrySettings()
+            .toBuilder()
+            // when performing a read via ReadObject, the ServerStream will have a default relative
+            // deadline set of `requestStartTime() + totalTimeout`, meaning if the specified
+            // RetrySettings have a totalTimeout of 10 seconds -- which should be plenty for
+            // metadata RPCs -- the entire ReadObject stream would need to complete within 10
+            // seconds.
+            // To allow read streams to have longer lifespans, crank up their timeouts, instead rely
+            // on idleTimeout below.
+            .setLogicalTimeout(Duration.ofDays(28))
```
frankyn, line 122: Why 28 days and not 24 hours? Don't mean to bikeshed, but i don't follow what makes a sensible default here.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITGrpcReadTimeoutTest.java
```diff
@@ -0,0 +1,219 @@
...
+ * <p>Instead of relying on total stream timeout, we rely on idleTimeout for the stream via {@link
+ * com.google.api.gax.rpc.ServerStreamingCallSettings.Builder#setIdleTimeout(Duration)}.
+ *
+ * <p>These tests force specific timeout scenarios to happen against an in-process grpc server to
+ * ensure our configuration of the StorageClient properly translates to the behavior we want.
+ *
+ * <p><i>NOTE:</i>Unfortunately, these tests are slow as they are waiting on wall clock time in
+ * several circumstances.
+ */
+public final class ITGrpcReadTimeoutTest {
```
frankyn, line 59: Nit this looks more like a conformance test or unit test and not an integration test similar to our existing IT which hit the live service.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
BenWhitehead: 🤷🏻 24 hours felt like it could bump up against some slow workload while 28 days is longer than most anyone would be willing to wait for even a slow system.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITGrpcReadTimeoutTest.java
BenWhitehead: I can drop the IT prefix. I think it's probably dangling there from the first attempt trying to use testbench.

frankyn: Remove the dangling `IT`

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
frankyn: fair enough; it's the "turbo" long-time for me is all. Please ignore.

frankyn (approved):

PR #1592, opened 2022-08-26
chore: update WriteFlushStrategy to inject bucket id

BenWhitehead:
Update WriteFlushStrategy to inject bucket id into x-goog-request-params upon opening the stream.

Fixes [b/238792234](http://b/238792234)

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
@@ -17,9 +17,11 @@
 package com.google.cloud.storage;
 
 import com.google.api.core.SettableApiFuture;
+import com.google.api.gax.grpc.GrpcCallContext;
 import com.google.api.gax.rpc.ApiStreamObserver;
 import com.google.api.gax.rpc.ClientStreamingCallable;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.storage.v2.WriteObjectRequest;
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
```diff
@@ -51,22 +53,34 @@ private WriteFlushStrategy() {}
...
-        new FsyncOnClose(write, committedTotalBytesCallback, onSuccessCallback);
+        new FsyncOnClose(write, bucketName, committedTotalBytesCallback, onSuccessCallback);
+  }
+
+  private static GrpcCallContext contextWithBucketName(String bucketName) {
+    GrpcCallContext ret = GrpcCallContext.createDefault();
+    if (bucketName != null && !bucketName.isEmpty()) {
+      return ret.withExtraHeaders(
+          ImmutableMap.of(
+              "x-goog-request-params", ImmutableList.of(String.format("bucket=%s", bucketName))));
```
frankyn, line 81: oh, i thought GAPIC would do this for us?

frankyn (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
BenWhitehead: Unfortunately it can't do it for us on the ClientStream, it's technically possible to open a client stream without a request at open time. Due to this, we need to pass it along when we first create the stream.

PR #1594, opened 2022-08-29
test: ITStorage converted to ITObjectTest

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadBlobWithoutAuth.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadBlobWithoutAuth.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.google.cloud.storage.it;
 
 import static com.google.common.truth.Truth.assertThat;
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadBlobWithoutAuth.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadBlobWithoutAuth.java
@@ -1,0 +1,92 @@
+package com.google.cloud.storage.it;
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.fail;
+import static org.junit.Assume.assumeFalse;
+
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.BlobInfo;
+import com.google.cloud.storage.BucketFixture;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
+import com.google.cloud.storage.StorageFixture;
+import com.google.cloud.storage.StorageOptions;
+import java.util.Iterator;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class ITDownloadBlobWithoutAuth {
+  @ClassRule(order = 1)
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
```diff
@@ -1069,28 +1008,6 @@ public void testCopyBlobFail() {
     }
   }
 
-  @Test
-  public void testReadAndWriteChannels() throws IOException {
```
sydney-munro, line 1073: I feel like this is covered via the various other Read/Write channel tests

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
```diff
@@ -1069,28 +1008,6 @@ public void testCopyBlobFail() {
...
-      readStringBytes = ByteBuffer.allocate(stringBytes.length);
-      reader.read(readBytes);
-      reader.read(readStringBytes);
-    }
-    assertArrayEquals(BLOB_BYTE_CONTENT, readBytes.array());
-    assertEquals(BLOB_STRING_CONTENT, new String(readStringBytes.array(), UTF_8));
-  }
-
   @Test
   public void testReadAndWriteChannelWithEncryptionKey() throws IOException {
```
sydney-munro, line 1012: Asking for opinions here, there are 3 tests that don't seem to fit nicely into other Read/Write channel tests to move.
- testReadAndWriteChannelWithEncryptionKey
- testReadAndWriteChannelsWithDifferentFileSize
- testReadAndWriteCaptureChannels

Any opinions on what to do with these tests?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadBlobWithoutAuth.java
```diff
@@ -0,0 +1,108 @@
...
+  private static String bucketName;
+
+  @BeforeClass
+  public static void setup() {
+    storage = storageFixture.getInstance();
+    bucketName = bucketFixture.getBucketInfo().getName();
+  }
+
+  @Test
+  public void testDownloadPublicBlobWithoutAuthentication() {
```
BenWhitehead, line 57: nit: this is three tests: 1) list, 2) download, 3) create. It'd be nice to split these up, can be done in a later PR.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
BenWhitehead: Agreed, this isn't doing anything unique.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
BenWhitehead: * `testReadAndWriteChannelWithEncryptionKey` I think is okay to remove, we now have other tests which actually verify args like encryption key are wired through.
* `testReadAndWriteChannelsWithDifferentFileSize` semi useful, but written in such a way that it's painful. We can refactor this one at a later point.
* `testReadAndWriteCaptureChannels` unfortunately we need to keep this, this is verifying that a "captured" channel can be restored - despite the fact that it's often confusing.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITWriteChannelConnectionPoolTest.java
```diff
@@ -0,0 +1,94 @@
...
+import com.google.cloud.storage.testing.RemoteStorageHelper;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class ITWriteChannelConnectionPoolTest {
```
BenWhitehead, line 41: IMO this test isn't valuable for us. It's strictly changing values on an api service we don't have direct control or consumption of. It's also not actually validating any pooling happens. We can remove this class.

BenWhitehead (approved):
Some small things we can cleanup, but feel free to merge after deciding to do in this PR or another one.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITWriteChannelConnectionPoolTest.java
frankyn: The original reason I wrote this was due to a customer issue where the http body wasn't read / cleared and left connections pending when using connection pooling.

We can most likely remove it at this point; it was mainly used to validate a fix that was made at that time.

PR #1597, opened 2022-08-30
test: add new classes to SerializationTest

BenWhitehead:
* Update SerializationTest lifecycle to account for the fact that instances of
  Storage now have `.close()`
* Add each new UnifiedOpt to the list of serializable things
* Add explicit HttpStorageOptions and GrpcStorageOptions to list of serializable
  things
* Update GrpcStorageOptions to integrate with ServiceOptions caching of service
  instances
* Add new test to ensure proper integration with ServiceOptions caching of service
  instances

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/SerializationTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/SerializationTest.java
@@ -26,17 +26,20 @@ import com.google.cloud.PageImpl;
 import com.google.cloud.ReadChannel;
 import com.google.cloud.Restorable;
 import com.google.cloud.storage.Acl.Project.ProjectRole;
+import com.google.cloud.storage.Storage.PredefinedAcl;
+import com.google.cloud.storage.UnifiedOpts.Opt;
 import com.google.cloud.storage.spi.v1.StorageRpc;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.Map;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class SerializationTest extends BaseSerializationTest {
 
-  private static final Storage STORAGE =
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/SerializationTest.java
```diff
@@ -67,14 +66,96 @@ public class SerializationTest extends BaseSerializationTest {
...
+    STORAGE = storageOptions.getService();
+    BLOB = BLOB_INFO.asBlob(STORAGE);
+    BUCKET = BUCKET_INFO.asBucket(STORAGE);
+    PAGE_RESULT = new PageImpl<>(null, "c", Collections.singletonList(BLOB));
+  }
+
+  @AfterClass
+  public static void afterClass() throws Exception {
+    if (STORAGE != null) {
+      STORAGE.close();
```
frankyn, line 90: This will be a "breaking change" in behavior no when using gRPC or am I misunderstanding?
We need to call `#close()` to shut down gRPC but wasn't a requirement before.

frankyn (approved):

google-cloud-storage/src/test/java/com/google/cloud/storage/SerializationTest.java
BenWhitehead: For existing http, `#close()` is a no-op. With the gRPC resources we need a way to tell it to shutdown otherwise they may leak. For gRPC itself, it has a leak detector and will provide a stack trace of where the allocation happened and later didn't cleanup.

This addition, actually also brings us closer to the gapics by being able to use try-with-resource for the client:

```
try (Storage s = StorageOptions.grpc().build().getService()) {
  Page<Bucket> buckets = s.list();
} // automatically calls #close to clean things up
```

frankyn: Gotcha, so no breaking change for HTTP. I misudnerstood. 

Thanks Ben.

frankyn (approved):

PR #1605, opened 2022-09-06
chore: remove DeleteRule handling in GrpcConversions

BenWhitehead:
DeleteRule is deprecated, and poised for removal. BlobInfo.Builder will automatically convert any DeleteRule provided to it to the corresponding LifecycleRule with delete condition.

Our Conversions classes need only encode/decode LifecycleRule.


frankyn (approved):
:fire:

PR #1607, opened 2022-09-06
test: Adding GRPC to ITDownloadTo and ITHmacTests

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadToTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadToTest.java
@@ -39,7 +39,6 @@ import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 
 @RunWith(ParallelParameterized.class)
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadToTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadToTest.java
@@ -19,17 +19,14 @@ package com.google.cloud.storage.it;
 import static com.google.common.truth.Truth.assertThat;
 import static org.junit.Assert.fail;
 
-import com.google.cloud.NoCredentials;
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadToTest.java
```diff
@@ -19,51 +19,102 @@
...
+  public static final StorageFixture storageFixtureGrpc =
+      StorageFixture.from(
+          () ->
+              StorageOptions.grpc()
+                  .setHost(TEST_BENCH.getGRPCBaseUri())
+                  .setCredentials(NoCredentials.getInstance())
+                  .setProjectId("test-project-id")
+                  .build());
+
+  @ClassRule(order = 1)
```
frankyn, line 60: I'll look this up; but could you add a comment clarifying why you needed to add `@ClassRule(order = x)`?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadToTest.java
```diff
@@ -19,51 +19,102 @@
...
+      String clientName, StorageFixture storageFixture, BucketFixture bucketFixture) {
+    this.storageFixture = storageFixture;
+    this.bucketFixture = bucketFixture;
+    this.clientName = clientName;
+  }
+
+  @Parameters(name = "{0}")
+  public static Iterable<Object[]> data() {
+    return Arrays.asList(
+        new Object[] {"JSON/storage.googleapis.com", storageFixtureHttp, bucketFixtureHttp},
```
frankyn, line 96: For data(): wondering if you can just split up "JSON/" and "GRPC/" to be its own element.

I may have missed how parameterization works with data(). Could it be a defined type that holds these values so it doesn't need to rely on generic `Object`?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITDownloadToTest.java
sydney-munro: yep can do. The reasoning is to force the rules to be applied in a certain order. storageFixtureGrpc needs the TestBench and the BucketFixtures need their StorageFixtures etc.

frankyn (approved):

PR #1614, opened 2022-09-12
fix: Fix bugs discovered by ITBucketLifecycleTest

sydney-munro:
This PR adds the GRPC Client to ITBucketLifecycleTest as well as fixes some bugs that were uncovered by getting this test running.

1. The way we were generating the field mask was not correct for GRPC. GRPC does not allow you to specify immutable fields in the fieldmask. We must filter these values out.
2. Adding AbortMPULifecycle Action to the Lifecycle rules codec.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -1296,7 +1296,7 @@ final class GrpcStorageImpl extends BaseService<StorageOptions> implements Stora
   }
 
   private FieldMask updateMaskGenerator(Message message) {
-    //TODO: Filter based on API behavior and not a list
+    // TODO: Filter based on API behavior and not a list
     return FieldMask.newBuilder()
         .addAllPaths(
             message.getAllFields().entrySet().stream()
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -1357,4 +1359,22 @@ private static GrpcStatusCode statusCodeFor(StatusCode.Code code) {
         throw new IllegalStateException("Unrecognized status code: " + code);
     }
   }
+
+  private static ImmutableList<String> updateFields() {
```
sydney-munro, line 1363: This isn't ideal but we need a way to create the field mask with only the appropriate fields so that we are not attempting to update immutable fields causing an exception. For Json since it uses patch we do not have to worry about this but GRPC does not appear to have an equivalent.

frankyn: We tried pairing to see if there's an alternative but the available options we tried were not successful. 

Specifically we tried:

```
.filter(f -> f.getKey().toProto().getOptions().getAllFields().entrySet.contains(FieldBehavior.IMMUTABLE))
```

```
.filter(f -> f.getKey().toProto().getOptions().getAllFields().containsValue(FieldBehavior.IMMUTABLE))
```

We are probably not using the correct type to determine equality; but we shall look at this another time.

@sydney-munro could you add a todo for now?

frankyn (commented):
Request for a todo; otherwise LGTM

frankyn (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
sydney-munro: I put the todo in the updateMaskGenerator if thats alright.

frankyn (approved):

PR #1615, opened 2022-09-13
feat: add constructor parameter for JsonFactory to HttpStorageRpc

### LINKED ISSUE #1584: Jackson usage is hardcoded in HttpStorageRpc
At https://github.com/googleapis/java-storage/blob/main/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java#L125

`new JacksonFactory()` is hardcoded, making it impossible to switch it out for `GsonFactory` and making it impossible to exclude Jackson as a dependency of `google-cloud-storage`.

BenWhitehead:
Fixes #1584

### DIFF
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -33,6 +33,11 @@
       <artifactId>google-http-client-jackson2</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


frankyn (approved):
LGTM, thanks ben, looks like we are starting to hit 429's.

PR #1619, opened 2022-09-14
feat: implement GrpcStorageImpl#*Iam* methods

BenWhitehead:
* Implement GrpcStorageImpl#getIamPolicy
* Implement GrpcStorageImpl#setIamPolicy
* Implement GrpcStorageImpl#testIamPermissions

### Add Codecs for policy related messages to GrpcConversions
* New `Codec<com.google.cloud.Policy, com.google.iam.v1.Policy>`
* New `Codec<com.google.cloud.Binding, com.google.iam.v1.Binding>`
* New `Codec<com.google.cloud.Condition, com.google.type.Expr>`

### Tests
Relying upon integration tests, as the backend is better at validating the actual values passed along. Codecs have been implemented to wire through the various string values matching property names.


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -845,6 +860,61 @@ private BlobInfo blobInfoDecode(Object from) {
...
+    to.setCondition(iamConditionCodec.decode(from.getCondition()));
+    return to.build();
+  }
+
+  private Expr conditionEncode(Condition from) {
+    Expr.Builder to = Expr.newBuilder();
+    to.setExpression(from.getExpression());
+    to.setTitle(from.getTitle());
+    to.setDescription(from.getDescription());
+    // apiary doesn't have a "location" field like grpc does
```
sydney-munro, line 906: so we don't want the location field in GRPC either?

sydney-munro (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
BenWhitehead: At this point, I don't think so. GCS is using a generic representation here and I don't think they'll populate location for grpc when there is no place to populate it for apiary.

PR #1622, opened 2022-09-15
feat: implement automatic retry for resumable uploads

BenWhitehead:
Plumb through the necessary items to allow retrying to take place in WriteStrategy.FlushOnEveryFlusher#flush.
Update corresponding builders along with this to pass through the necessary items.

### Refactorings
* Move {Unbuffered,Buffered}{Direct,Resumable}UploadBuilder classes to be nested in their logical "parent builders" to allow closing over fields set at the higher level
* Move GrpcTransformPageDecoratorTest.java#retryingDeps -> TestUtils#defaultRetryingDeps

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
@@ -18,8 +18,11 @@ package com.google.cloud.storage;
 
 import com.google.api.core.SettableApiFuture;
 import com.google.api.gax.grpc.GrpcCallContext;
+import com.google.api.gax.retrying.ResultRetryAlgorithm;
 import com.google.api.gax.rpc.ApiStreamObserver;
 import com.google.api.gax.rpc.ClientStreamingCallable;
+import com.google.cloud.storage.Conversions.Decoder;
+import com.google.cloud.storage.Retrying.RetryingDependencies;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.storage.v2.WriteObjectRequest;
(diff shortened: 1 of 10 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/GapicUnbufferedWritableByteChannelTest.java
```diff
@@ -176,16 +192,113 @@ public void resumableUpload() throws IOException, InterruptedException, Executio
...
+            (obs, requests) -> {
+              writeCount.getAndIncrement();
+              if (requests.equals(ImmutableList.of(req1))) {
+                obs.onNext(resp1);
+                obs.onCompleted();
+              } else if (requests.equals(ImmutableList.of(req2))) {
+                obs.onNext(resp2);
+                if (req2SendErr.get()) {
+                  req2SendErr.set(false);
+                  obs.onError(apiException(Code.DATA_LOSS));
```
frankyn, line 212: Are you going to test other retriable gRPC codes?

google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
```diff
@@ -105,39 +111,51 @@ interface Flusher {
...
-        write.onNext(message);
-        first = false;
-      }
-      write.onCompleted();
-      observer.await();
+      Retrying.run(
+          deps,
+          alg,
+          () -> {
+            Observer observer = new Observer(sizeCallback, completeCallback);
```
frankyn, line 140: I don't think I've come across this before; what does Observer do for us?

frankyn (commented):
LGTM, just have couple questions

frankyn (approved):
My questions aren't blocking but would like to understand.

Approved.

google-cloud-storage/src/test/java/com/google/cloud/storage/GapicUnbufferedWritableByteChannelTest.java
BenWhitehead: This is specifically a non-default retryable status code to ensure our retry wiring is working.

The other grpc status codes that we do want to retry are covered via tests for the translation from grpc to StorageException.

google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
BenWhitehead: In this case, `Observer` is how our client receives events from the server. We'll get an onComplete each time the server finishes processing what we're calling a "flush". Or, an onError if some failure happens.

BenWhitehead:
Thanks for the review!

PR #1642, opened 2022-09-21
test: ITObjectTest and bug fixes

sydney-munro:


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
```diff
@@ -1328,7 +1675,7 @@ private void testAutoContentType(String method) throws IOException {
     String[] names = {"file1.txt", "dir with spaces/Pic.Jpg", "no_extension"};
```
sydney-munro, line 1675: These tests are currently failing, it is unclear to me if we need to be concerned with having content type default to "application/octet-stream" for GRPC like it is in JSON. if not I can separate the logic and turn off the ones checking the default value.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
```diff
@@ -134,13 +176,18 @@ public static void afterClass() {
...
-        BlobInfo.newBuilder(BUCKET, blobName).setCustomTime(System.currentTimeMillis()).build();
-    Blob remoteBlob = storage.create(blob, BLOB_BYTE_CONTENT);
+        BlobInfo.newBuilder(bucketFixture.getBucketInfo().getName(), blobName)
+            .setCustomTime(System.currentTimeMillis())
+            .build();
+    Blob remoteBlob = storageFixture.getInstance().create(blob, BLOB_BYTE_CONTENT);
     assertNotNull(remoteBlob);
     assertNotNull(remoteBlob.getCustomTime());
-    assertEquals(blob.getBucket(), remoteBlob.getBucket());
+    assertTrue(remoteBlob.getBucket().contains(blob.getBucket()));
```
BenWhitehead, line 185: These values should still be equal, no? Why did we need to change to `contains`?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
```diff
@@ -149,95 +196,126 @@ public void testCreateBlob() {
   public void testCreateBlobMd5Crc32cFromHexString() {
     String blobName = "test-create-blob-md5-crc32c-from-hex-string";
     BlobInfo blob =
-        BlobInfo.newBuilder(BUCKET, blobName)
+        BlobInfo.newBuilder(bucketFixture.getBucketInfo().getName(), blobName)
```
BenWhitehead, line 199: There's an overload on this method where you can pass it the `BucketInfo` and it'll extract the name.
`BlobInfo.newBuilder(bucketFixture.getBucketInfo(), blobName)`

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
```diff
@@ -87,41 +91,79 @@ public class ITObjectTest {
...
+  private final BucketFixture requesterPaysFixture;
+  private final StorageFixture storageFixture;
+  private final String clientName;
+
+  public ITObjectTest(
+      String clientName,
+      StorageFixture storageFixture,
+      BucketFixture bucketFixture,
+      BucketFixture requesterPaysFixture) {
+    this.storageFixture = storageFixture;
```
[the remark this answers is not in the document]
BenWhitehead: If you want to do this, intellj can do the refactoring for you using [Extract Field](https://www.jetbrains.com/help/idea/extract-field.html) which will save you from having to change the ~150 locations manually.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
BenWhitehead: Both JSON and grpc are using the same mechanism to automatically determine content-type based on extension. I'll try to dig in an see what's going on.

BenWhitehead (changes requested):

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
sydney-munro: for GRPC since it uses the /project/_/<bucketname> i used contains to get around this.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
BenWhitehead: I've just posted a PR with some tests for some of the common file extensions to validate our handling https://github.com/googleapis/java-storage/pull/1647

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
BenWhitehead: Ahh, we've got a bug in the Blob decoder then. The `projects/_/buckets` should be stripped.

Looks like we need to change `GrpcConversions#blobInfoDecode` to start with
```
    BlobInfo.Builder toBuilder =
        BlobInfo.newBuilder(
            BlobId.of(
                bucketNameCodec.decode(from.getBucket()), from.getName(), from.getGeneration()));
```

sydney-munro: ah I didn't think of it as a bug, ill fix up the decoder.

BenWhitehead (approved):
Thanks for all the work on this!

PR #1644, opened 2022-09-22
chore: wire automatic retries into direct uploads for GrpcStorageImpl

BenWhitehead:
* com.google.cloud.storage.GrpcStorageImpl.create(com.google.cloud.storage.BlobInfo, byte[], int, int, com.google.cloud.storage.Storage.BlobTargetOption...)
* com.google.cloud.storage.GrpcStorageImpl.createFrom(com.google.cloud.storage.BlobInfo, java.nio.file.Path, int, com.google.cloud.storage.Storage.BlobWriteOption...)

Set default chunkSize to 16MiB instead of 15MiB. While 15MiB is evenly divisible by 256KiB, 16MiB is also evenly divisible by 2MiB the max number of bytes that can be provided in a WriteObjectRequest.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -16,7 +16,7 @@
 
 package com.google.cloud.storage;
 
-import static com.google.cloud.storage.ByteSizeConstants._15MiB;
+import static com.google.cloud.storage.ByteSizeConstants._16MiB;
 import static com.google.cloud.storage.ByteSizeConstants._256KiB;
 import static com.google.cloud.storage.Utils.bucketNameCodec;
 import static com.google.cloud.storage.Utils.ifNonNull;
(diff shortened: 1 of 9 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -258,7 +260,7 @@ public Blob create(BlobInfo blobInfo, InputStream content, BlobWriteOption... op
   @Override
   public Blob createFrom(BlobInfo blobInfo, Path path, BlobWriteOption... options)
       throws IOException {
-    return createFrom(blobInfo, path, _15MiB, options);
+    return createFrom(blobInfo, path, _16MiB, options);
```
frankyn, line 263: *gasp* alignment.

google-cloud-storage/src/test/java/com/google/cloud/storage/TestUtils.java
```diff
@@ -126,4 +127,24 @@ public ApiClock getClock() {
...
     };
   }
+
+  /**
+   * Search {@code t} for an instance of {@code T} either directly or via a cause
+   *
+   * @return The found instance of T or null if not found.
+   */
+  @Nullable
+  public static <T extends Throwable> T findThrowable(Class<T> c, Throwable t) {
```
frankyn, line 137: Question: wondering why this isn't just a boolean return value if it's being used to check existence of from its usage in this PR?

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcStorageImplUploadRetryTest.java
```diff
@@ -0,0 +1,255 @@
...
+      }
+    }
+  }
+
+  private static final class Resumable {
+
+    private static final String uploadId = "upload-id";
+
+    private static final Object obj =
+        Object.newBuilder().setBucket("projects/_/buckets/buck").setName("obj").build();
```
frankyn, line 189: Looks like `"projects/_/buckets/buck"` values can be pulled up into a single variable used by the test file instead of inline literals. You may have already considered this and I would like to understand.

frankyn (commented):
Questions; nothing blocking.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
BenWhitehead: Multilevel coordinated alignment indeed

google-cloud-storage/src/test/java/com/google/cloud/storage/GrpcStorageImplUploadRetryTest.java
BenWhitehead: hehe, nope copy pasted. I'll create a common variable

google-cloud-storage/src/test/java/com/google/cloud/storage/TestUtils.java
BenWhitehead: No strong reason. When I wrote the method it felt to make more sense as a things that finds something rather the more "contains"/"caused by" type check.

I can create an overload to return the boolean if you want me to.

frankyn (approved):

PR #1675, opened 2022-09-28
feat: wire through read_mask for GrpcStorageImpl methods related to Blob and Bucket

BenWhitehead:
### Read Mask
Both JSON and gRPC provide the means of being able to specify a mask to select specific fields be returned in a response. However, there are slightly differing approaches taken between the different transports as well as outward facing impacts for existing code of our client.

Primary difference between JSON and gRPC is null vs default values. Several fields in gRPC can be excluded from the response from GCS but protobuf will still yield a value for our decoder. To address this compatibility issue, in the case of gRPC if a read_mask is defined we will clear out any non-selected field after decoding.

### Impact to existing JSON support
The new NamedField construct forces a decoupling from FieldSelector and it's associated Helper as JSON and gRPC have different formats for specifying their selectors/masks.

We are no longer using FieldSelector.Helper methods, instead we are using our own code to stitch together the values for selection. For JSON, we're not using the "slash form" presented in https://cloud.google.com/storage/docs/json_api#partial-response (i.e. `items/name,items/bucket` vs `items(name,bucket)`). Functionally these are equivalent and do not chang any of our tests which verify behavior.

### New internal only API
Both BlobInfo.Builder and BucketInfo.Builder have had `clear*` methods added to allow explicitly setting the field to `null` side stepping any possible guarding done in the respective `set*` method.

These clear methods will likely never be public, and also are no longer necessary if our decoders are field mask aware when they try to process a response.

### GrpcConversions
Fixed a couple decoder issues which present when comparing explicitly to json

### Field @TransportCompatibility
Added @TransportCompatibility too all BlobField and BucketField values

### ITReadMaskTest.java
Add a new test class which will validate expected behavior for all fields across json and grpc

### StorageImplMockitoTest.java
Remove 4 tests which weren't valuable now that we have more robust field selector and options tests

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
@@ -731,7 +731,7 @@ final class UnifiedOpts {
      * However, refactoring the whole model pipeline for both json and grpc is going to be a large
      * change.
      */
-    Decoder<Blob, Blob> clearBlobFields() {
+    Decoder<Blob, Blob> clearUnselectedBlobFields() {
       return b -> {
         if (val.isEmpty()) {
           return b;
@@ -754,7 +754,7 @@ final class UnifiedOpts {
      * However, refactoring the whole model pipeline for both json and grpc is going to be a large
      * change.
      */
-    Decoder<Bucket, Bucket> clearBucketFields() {
+    Decoder<Bucket, Bucket> clearUnselectedBucketFields() {
       return b -> {
         if (val.isEmpty()) {
           return b;
@@ -2205,14 +2205,14 @@ final class UnifiedOpts {
     Decoder<Blob, Blob> clearBlobFields() {
       return filterTo(Fields.class)
           .findFirst()
-          .map(Fields::clearBlobFields)
+          .map(Fields::clearUnselectedBlobFields)
           .orElse(Decoder.identity());
     }
 
     Decoder<Bucket, Bucket> clearBucketFields() {
       return filterTo(Fields.class)
           .findFirst()
-          .map(Fields::clearBucketFields)
+          .map(Fields::clearUnselectedBucketFields)
           .orElse(Decoder.identity());
     }
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
@@ -18,17 +18,24 @@ package com.google.cloud.storage;
 
 import static com.google.cloud.storage.Utils.projectNameCodec;
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Predicates.not;
 
 import com.google.api.gax.grpc.GrpcCallContext;
+import com.google.cloud.storage.Conversions.Decoder;
+import com.google.cloud.storage.Storage.BlobField;
+import com.google.cloud.storage.Storage.BucketField;
 import com.google.cloud.storage.spi.v1.StorageRpc;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.hash.HashCode;
 import com.google.common.hash.Hashing;
 import com.google.common.io.BaseEncoding;
 import com.google.iam.v1.GetIamPolicyRequest;
 import com.google.protobuf.ByteString;
+import com.google.protobuf.FieldMask;
 import com.google.storage.v2.CommonObjectRequestParams;
 import com.google.storage.v2.ComposeObjectRequest;
 import com.google.storage.v2.CreateBucketRequest;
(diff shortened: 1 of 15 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -661,13 +669,209 @@ public Mapper<ListObjectsRequest.Builder> listObjects() {
...
+    }
+
+    /**
+     * Define a decoder which can clear out any fields which may have not been selected.
+     *
+     * <p>This approach, isn't ideal at the backside after decoding has already taken place.
+     * However, refactoring the whole model pipeline for both json and grpc is going to be a large
+     * change.
+     */
+    Decoder<Blob, Blob> clearBlobFields() {
```
sydney-munro, line 734: nit: can we make this method name more obvious like clearUnselectedBlobFields or something just so it doesn't seem like someone can clear the FieldMask values with it.

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -661,13 +669,209 @@ public Mapper<ListObjectsRequest.Builder> listObjects() {
...
+    }
+
+    /**
+     * Define a decoder which can clear out any fields which may have not been selected.
+     *
+     * <p>This approach, isn't ideal at the backside after decoding has already taken place.
+     * However, refactoring the whole model pipeline for both json and grpc is going to be a large
+     * change.
+     */
+    Decoder<Bucket, Bucket> clearBucketFields() {
```
sydney-munro, line 757: same as above.

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -661,13 +669,209 @@ public Mapper<ListObjectsRequest.Builder> listObjects() {
...
+        }
+      };
+    }
+
+    private Set<String> getPaths() {
+      //noinspection Guava
+      return val.stream()
+          .map(NamedField::stripPrefix)
+          .map(NamedField::getGrpcName)
+          .filter(not(grpcExcludedFields::contains))
```
sydney-munro, line 778: nit: probably want to filter and then get the GRPC name because we shouldn't have GRPC names for non GRPC fields?

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -661,13 +669,209 @@ public Mapper<ListObjectsRequest.Builder> listObjects() {
...
+            .put(
+                BlobField.TIME_STORAGE_CLASS_UPDATED.getGrpcName(),
+                BlobInfo.Builder::clearTimeStorageClassUpdated)
+            .put(BlobField.UPDATED.getGrpcName(), BlobInfo.Builder::clearUpdateTime)
+            .build();
+
+    private static final ImmutableMap<String, Mapper<BucketInfo.Builder>> bucketInfoFieldClearers =
+        ImmutableMap.<String, Mapper<BucketInfo.Builder>>builder()
+            .put(BucketField.ACL.getGrpcName(), BucketInfo.Builder::clearAcl)
+            // .put(BucketField.AUTOCLASS.getGrpcName(), b -> b.clearAutoclass())
```
sydney-munro, line 837: are we leaving this out because autoclass isn't out yet?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITReadMaskTest.java
```diff
@@ -0,0 +1,446 @@
...
+            .build();
+    sh.compose(c2);
+    sh.compose(c3);
+    sh.compose(c4);
+
+    blobId = blobId1;
+  }
+
+  /*
+  public static final class UpdateMask {
```
sydney-munro, line 119: reminding you of this that you said you would remove during our pair session.

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
BenWhitehead: Will do

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
BenWhitehead: Will do

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
BenWhitehead: The exclusions are a `ImmutableSet<String>` because some fields are hard coded as strings for apiary ("nextPageToken") and don't have defined values we can reference like we can for `Storage.BlobField` and `Storage.BucketField`.

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
BenWhitehead: Yeah, autoclass hasn't been wired into BucketInfo yet.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITReadMaskTest.java
BenWhitehead: Thank you, removed.

sydney-munro (approved):
LGTM. Thanks!

PR #1681, opened 2022-09-30
chore: fix native-image.properties to allow running jqwik tests

BenWhitehead:
test(deps): bump junit-platform.version to 5.9.1

Also removed a dangling bucket deletion in ITNotificationTest which was resulting in an exception at cleanup time.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITNotificationTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITNotificationTest.java
@@ -25,7 +25,6 @@ import com.google.cloud.storage.Notification;
 import com.google.cloud.storage.NotificationInfo;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageFixture;
-import com.google.cloud.storage.testing.RemoteStorageHelper;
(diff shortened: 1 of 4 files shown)


frankyn (approved):

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITNotificationTest.java
```diff
@@ -190,8 +187,7 @@ public void testNotification() throws InterruptedException, ExecutionException {
           .isNull();
       assertThat(storage.listNotifications(bucketFixture.getBucketInfo().getName())).isEmpty();
     } finally {
-      RemoteStorageHelper.forceDelete(
-          storage, bucketFixture.getBucketInfo().getName(), 5, TimeUnit.SECONDS);
+      // delete is taken care of by BucketFixture now
```
frankyn, line 190: Are you oging to remove the finally block?

BenWhitehead: Probably not, because then the code formatter will change all the indentation for the whole test and git-blame me on those lines 😄

PR #1688, opened 2022-10-03
feat: add GrpcStorageOptions.Builder#setAttemptDirectPath

BenWhitehead:
Add new builder method to all setting attempt of direct path connection.

If enabled, the underlying gapic client will be configured to attempt connecting via direct path and using client side load balancing.

deps: add a direct dependency to io.grpc:grpc-googleapis

Rumour has it this dependency will be removed from gax, and we need to pull it up to our level to ensure it's there for storage direct path.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
@@ -39,6 +39,8 @@ import com.google.cloud.spi.ServiceRpcFactory;
 import com.google.cloud.storage.TransportCompatibility.Transport;
 import com.google.cloud.storage.spi.StorageRpcFactory;
 import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.storage.v2.ReadObjectRequest;
 import com.google.storage.v2.ReadObjectResponse;
@@ -64,6 +66,7 @@ public final class GrpcStorageOptions extends StorageOptions
 
   private final GrpcRetryAlgorithmManager retryAlgorithmManager;
   private final Duration terminationAwaitDuration;
+  private final boolean attemptDirectPath;
 
   @BetaApi
   public GrpcStorageOptions(Builder builder, GrpcStorageDefaults serviceDefaults) {
@@ -75,6 +78,7 @@ public final class GrpcStorageOptions extends StorageOptions
     this.terminationAwaitDuration =
         MoreObjects.firstNonNull(
             builder.terminationAwaitDuration, serviceDefaults.getTerminationAwaitDuration());
+    this.attemptDirectPath = builder.attemptDirectPath;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


frankyn (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -112,13 +126,28 @@ StorageSettings getStorageSettings() throws IOException {
...
+          .setDirectPathServiceConfig(
+              ImmutableMap.of(
+                  "loadBalancingConfig",
+                  ImmutableList.of(
+                      ImmutableMap.of(
+                          "grpclb",
+                          ImmutableMap.of(
+                              "childPolicy",
+                              ImmutableList.of(
+                                  ImmutableMap.of("round_robin", ImmutableMap.of())))))));
```
frankyn, line 144: This is ugly, no offence but is there a better way to do this?

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -380,6 +426,11 @@ public StorageRetryStrategy getStorageRetryStrategy() {
     public Duration getTerminationAwaitDuration() {
       return Duration.ofMinutes(1);
     }
+
+    @BetaApi
+    public boolean isAttemptDirectPath() {
+      return false;
```
frankyn, line 432: wait shouldn't this be a class property?

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -236,6 +266,22 @@ public Builder setTerminationAwaitDuration(Duration terminationAwaitDuration) {
...
+     * <p>In order to use direct path, both this option must be true and the environment variable
+     * (not system property) {@code GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS} must be true.
+     *
+     * <p><i>NOTE</i>There is no need to specify a new endpoint via {@link #setHost(String)} as the
+     * underlying code will translate the normal {@code https://storage.googleapis.com:443} into the
+     * proper Direct Path URI for you.
+     */
+    @BetaApi
+    public GrpcStorageOptions.Builder setAttemptDirectPath(boolean attemptDirectPath) {
+      this.attemptDirectPath = attemptDirectPath;
```
frankyn, line 281: Should this be enabled by default?

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
BenWhitehead: This method is under `GrpcStorageDefaults`. And since it's the literal value of a boolean, doesn't need a constant to create a single instance.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
BenWhitehead: With the latest version of gax-java an environment variable still needs to be set in order for direct path to actually be attempted. I think this type of thing we can change the default on in the future without there being an issue as long as there isn't any overhead to the attempt.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
BenWhitehead: Ended up not being needed. I removed it.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -94,10 +98,20 @@ Duration getTerminationAwaitDuration() {
...
-    String endpoint = String.format("%s:%d", uri.getHost(), port);
+    int port = uri.getPort();
+    // Gax routes the endpoint into a method which can't handle schemes, unless for direct path
+    // try and strip here if we can
+    switch (scheme) {
+      case "http":
+        endpoint = String.format("%s:%s", uri.getHost(), port > 0 ? port : 80);
+        break;
+      case "https":
+        endpoint = String.format("%s:%s", uri.getHost(), port > 0 ? port : 443);
```
burkedavison, line 110: Implementer's choice, just looking at alternative syntax to avoid the implicit-fallthrough `default` case:
```java
private String resolveEndpoint(String originalEndpoint) {
  URI uri = URI.create(originalEndpoint);

  int defaultPort = scheme.equals("http") ? 80 : 443;
  int explicitPort = uri.getPort();
  int port = explicitPort > 0 ? explicitPort : defaultPort;

  // Gax routes the endpoint into a method which can't handle schemes.
  // Unless for direct path, try and strip here if we can.
  return (schema.equals("http") || schema.equals("https"))
    ? String.format("%s:%s", uri.getHost(), port)
    : originalEndpoint;
}

@InternalApi
StorageSettings getStorageSettings() throws IOException {
  String endpoint = resolveEndpoint(getHost());

PR #1696, opened 2022-10-06
fix: update Object Write handling to be able to specify an explicit crc32c value

BenWhitehead:
Storage#createFrom and Storage#writer allow specifying an explicit crc32c value which should be passed to GCS for validation. Update our uploads to allow these explicitly defined values to be passed in the `finish_write` message to GCS.

Update UnifiedOpts.Crc32cValue to represent its value as an int rather than the b64 string from apiary. Internally when the value is necessary for apiary it will be converted to the b64 string representation.

Map gRPC Status `DATA_LOSS` to HTTP 400 Bad Request to mirror the behavior of JSON when a checksum validation fails.

Add a new Hasher.constant instant which will function to specify a constant value for the cumulative value of a write.

Various plumbing tweaks to allow passing this value all the way down and also attempting to "nullSafeConcat"

Add new integration tests to try the negative case to ensue we are getting checksum failures when we pass mismatch values and content.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BackwardCompatibilityUtils.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BackwardCompatibilityUtils.java
@@ -93,6 +93,7 @@ final class BackwardCompatibilityUtils {
       case OK:
         return 200;
         // 400 Bad Request
+      case DATA_LOSS:
       case INVALID_ARGUMENT:
       case OUT_OF_RANGE:
         return 400;
@@ -133,7 +134,6 @@ final class BackwardCompatibilityUtils {
       case ABORTED: // ?
       case CANCELLED: // ?
       case UNKNOWN: // ?
-      case DATA_LOSS: // ?
       default:
         return 0;
     }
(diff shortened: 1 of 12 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BackwardCompatibilityUtils.java
```diff
@@ -133,7 +134,6 @@ static int grpcCodeToHttpStatusCode(Code code) {
       case ABORTED: // ?
       case CANCELLED: // ?
       case UNKNOWN: // ?
-      case DATA_LOSS: // ?
```
frankyn, line 136: Are the `?` TODO's that need to be finished before cutting a relaese?

frankyn (approved):
LGTM, just have one question about error code mapping to HTTP 1.1 errors

google-cloud-storage/src/main/java/com/google/cloud/storage/BackwardCompatibilityUtils.java
BenWhitehead: Hopefully, they're being tracked in the tracker. However, `ABORTED`, `CANCELLED`, `UNKNOWN` probably fall into the "connection errors" bucket we've got for http.

PR #1707, opened 2022-10-13
chore: plumb through md5 request validate options

BenWhitehead:
### Refactoring
Remove Hasher.Constant. I was made aware that GCS allows specifying the expected ObjectChecksums in the first WriteObjectRequest sent for a resumable session. Thereby we don't need to push the value all the way down to the WriteFlusher for the final write.

This also allows us to specify an md5 value upfront, since we only compute crc32c's due to their left distributed property allowing incremental calculation.

Move ITOptionRegressionTest#Content out into it's only class in tests to allow other classes to use it.


BenWhitehead:
Not quite ready to merge, need to validate some semantics questions.

BenWhitehead:
Verified semantic expectations, this is ready for review.

BenWhitehead:
Revised, previous verification was incomplete. This isn't going to work. Closing until such time as a new approach can be implemented.

PR #1713, opened 2022-10-17
chore: refactor and cleanup translation between gRPC status codes and http codes

BenWhitehead:
Move all mapping from GrpcStorageImpl and BackwardCompatibilityUtils to GrpcToHttpStatusCodeTranslation.

Define new GrpcToHttpStatusCodeTranslation.StatusCodeMapping to explicitly define a mapping between io.grpc.Status.Code and our http code

Update tests to be explicit and not depend on utility methods which can also be used by the code under test.

Add new tests to ensure expected status codes are returned when evaluating DefaultStorageRetryStrategy

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -18,6 +18,7 @@ package com.google.cloud.storage;
 
 import static com.google.cloud.storage.ByteSizeConstants._16MiB;
 import static com.google.cloud.storage.ByteSizeConstants._256KiB;
+import static com.google.cloud.storage.GrpcToHttpStatusCodeTranslation.resultRetryAlgorithmToCodes;
 import static com.google.cloud.storage.Utils.bucketNameCodec;
 import static com.google.cloud.storage.Utils.ifNonNull;
 import static com.google.cloud.storage.Utils.projectNameCodec;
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -1394,80 +1376,6 @@ ApiFuture<ResumableWrite> startResumableWrite(
...
-      case DATA_LOSS:
-        return GrpcStatusCode.of(Code.DATA_LOSS);
-      case UNAUTHENTICATED:
-        return GrpcStatusCode.of(Code.UNAUTHENTICATED);
-      default:
-        throw new IllegalStateException("Unrecognized status code: " + code);
-    }
-  }
-
-  private static ImmutableList<String> updateFields() {
```
frankyn, line 1451: Did we move away from using this list?

frankyn (approved):
Over LGTM, just had one question that I pinged you offline about. 
I also like the change in principle of writing tests without relying on the same utilities as the implementation.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
BenWhitehead: Yeah, in https://github.com/googleapis/java-storage/pull/1637 we added individual field level modification tracking so we don't guess.

PR #1722, opened 2022-10-19
test: add more scenarios to be validated for checksum support

BenWhitehead:
* Define 3 new content size scenarios to be validated
  1. data is small enough to fit within a single message of a single stream
  2. data requires multiple messages within a single stream
  3. data requires multiple messages within multiple streams
* Refactor checksum validation centric tests out of ITObjectTest into new ITObjectChecksumSupportTest
* Move ITOptionRegressionTest.Content to top level as ChecksummedTestContent and flush it out a bit more for cross usage by ITObjectChecksumSupportTest

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectChecksumSupportTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectChecksumSupportTest.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectChecksumSupportTest.java
```diff
@@ -0,0 +1,214 @@
...
+    int _2MiB = 2 * 1024 * 1024;
+    int _24MiB = 24 * 1024 * 1024;
+    ImmutableList<ChecksummedTestContent> contents =
+        ImmutableList.of(
+            // small, single message single stream when resumable
+            ChecksummedTestContent.of(gen.genBytes(15)),
+            // med, multiple messages single stream when resumable
+            ChecksummedTestContent.of(gen.genBytes(_2MiB + 3)),
+            // large, multiple messages and multiple streams when resumable
+            ChecksummedTestContent.of(gen.genBytes(_24MiB + 5)));
```
frankyn, line 106: why the 3 and 5 magic numbers being added?

frankyn (approved):
Only had 1 question; scenarios otherwise LGTM;

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectChecksumSupportTest.java
BenWhitehead: just some number to push it past the 2MiB evenly divisible values

PR #1734, opened 2022-10-25
feat: update retries for Notifications

JesseLovelace:
Adds retry algorithms for notifications to HttpRetryAlgorithmManager.java, edits StorageImpl to use those algorithms, and adds notification retry conformance tests


google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/CtxFunctions.java
```diff
@@ -133,6 +141,34 @@ static final class Rpc {
...
+                  .build();
+          return state.with(
+              ctx.getStorage().createNotification(c.getBucketName(), notificationInfo));
+        });
+
+    static final CtxFunction pubsubTopicSetup =
+        defaultSetup.andThen(pubsubTopic);
+
+    static final CtxFunction notificationSetup =
+        pubsubTopicSetup.andThen(notification);
```
BenWhitehead, line 169: I think these new CtxFunctions can simply be added to the existing `ResourceSetup` class to avoid the need for additional imports/qualified references.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/State.java
```diff
@@ -289,6 +293,10 @@ public <T> State consume(Page<T> page) {
     return newStateWith(KEY_LIST_OBJECTS, collect);
   }
 
+  public <T> State with(List<T> listResult) {
+    return newStateWith(KEY_LIST_OBJECTS, listResult);
```
BenWhitehead, line 297: Can we add a new `KEY_LIST_NOTIFICATION` (and corresponding `has`, `get` and `set` methods) so that we don't possibly overwrite objects that might already be held?

```
private static final Key<List<Notification>> KEY_LIST_NOTIFICATION = new Key<>("list<notification>");
```

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/State.java
```diff
@@ -301,6 +309,30 @@ public boolean hasComposeRequest() {
     return hasValue(KEY_COMPOSE_REQUEST);
   }
 
+  public boolean hasPubsubTopic() {
+    return hasValue(KEY_PUBSUB_TOPIC);
+  }
+
+  public TopicName getTopic() {
```
BenWhitehead, line 316: inconsistent names
```suggestion
  public boolean hasPubsubTopic() {
    return hasValue(KEY_PUBSUB_TOPIC);
  }

  public TopicName getPubsubTopic() {
```

nit: `topicName` might make more sense than `pubsubTopic`

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -1483,96 +1483,48 @@ public Notification createNotification(
...
-    } catch (RetryHelperException e) {
-      throw StorageException.translateAndThrow(e);
-    }
+    ResultRetryAlgorithm<?> algorithm =
+        retryAlgorithmManager.getForNotificationCreate(bucket, notificationPb);
+    return run(
+        algorithm,
+        () -> storageRpc.createNotification(bucket, notificationPb),
+        codecs.notificationInfo()::decode
+    ).asNotification(this);
```
BenWhitehead, line 1492: Not incorrect, but my personal preference is to keep the `.as*(this)` next to the decode. My perspective is that the transformation is more an attribute of the transformation from apiary result to our model than post process of result from `run`.

```suggestion
    return run(
        algorithm,
        () -> storageRpc.createNotification(bucket, notificationPb),
        n -> codecs.notificationInfo().decode(n).asNotification(this)
    );
```

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -1483,96 +1483,48 @@ public Notification createNotification(
...
-    } catch (RetryHelperException e) {
-      throw StorageException.translateAndThrow(e);
-    }
+    ResultRetryAlgorithm<?> algorithm
+        = retryAlgorithmManager.getForNotificationGet(bucket, notificationId);
+    return run(
+        algorithm,
+        () -> storageRpc.getNotification(bucket, notificationId),
+        codecs.notificationInfo()::decode
+    ).asNotification(this);
```
BenWhitehead, line 1503: ditto here

BenWhitehead (changes requested):

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
```diff
@@ -151,6 +151,7 @@ public static Collection<Object[]> testCases() throws IOException {
             .setHost(TEST_BENCH.getBaseUri().replaceAll("https?://", ""))
             .setTestAllowFilter(
                 RetryTestCaseResolver.includeAll()
+                    .and(RetryTestCaseResolver.specificMappings(246, 247, 248, 249))
```
BenWhitehead, line 154: Don't forget to remove this predicate before merging

BenWhitehead (approved):
For the dependencies build you'll need to declare the dependency on the pubsub protos module. When you do that, can you also add it to the renovate config[1] as a test dep (so we don't trigger a release when it's bumped)?

[1] https://github.com/googleapis/java-storage/blob/main/renovate.json#L74-L76

PR #1775, opened 2022-11-16
docs: annotate all Option factory methods with their Nullability bounds

BenWhitehead:
Update all option factory methods to explicitly state their Nullability bounds. All methods which take non-primitive parameters are @NonNull, and will be published in the javadocs.

Add new tests to explicitly verify null argument validation.

All factory methods now check their args for nullness, instead of depending on implicit checking.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
@@ -44,6 +44,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import org.checkerframework.checker.nullness.qual.NonNull;
 
 /**
  * A Google cloud storage bucket.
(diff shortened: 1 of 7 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
```diff
@@ -142,7 +143,7 @@ private BlobTargetOption(ObjectTargetOpt opt) {
 
     /** Returns an option for specifying blob's predefined ACL configuration. */
     @TransportCompatibility({Transport.HTTP, Transport.GRPC})
-    public static BlobTargetOption predefinedAcl(Storage.PredefinedAcl acl) {
+    public static BlobTargetOption predefinedAcl(Storage.@NonNull PredefinedAcl acl) {
```
sydney-munro, line 146: nit: can this not look like 
`public static BlobTargetOption predefinedAcl(@NonNull Storage.PredefinedAcl acl)`

sydney-munro (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
BenWhitehead: The `@NonNull` is applicable to a type, not the method parameter. So it needs to immediately precede the type, in this case `PredefinedAcl` which happens to be nested in `Storage`.

If the annotation is moved before Storage it will result in a compile error:
```
.../google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java:146:66
java: scoping construct cannot be annotated with type-use annotation: @org.checkerframework.checker.nullness.qual.NonNull
```

PR #1806, opened 2022-12-08
feat: implement GrpcStorageImpl#createDefaultAcl

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -936,12 +937,59 @@ public boolean deleteDefaultAcl(String bucket, Entity entity) {
...
+
+      //noinspection DataFlowIssue
+      Optional<Acl> first =
+          updateResult.getDefaultObjectAclList().stream()
+              .filter(entityPredicate)
+              .findFirst()
+              .map(codecs.objectAcl()::decode);
+
+      return first.orElseThrow(
+          () -> new StorageException(404, "Acl update call success, but not in response"));
```
frankyn, line 989: When would this occur?

BenWhitehead: Hopefully never, but since we have to search for the updated acl in the response, if the response somehow doesn't included the updated acl we need to signal that somehow. 404 here doesn't feel totally correct, but I'm not sure 500 is better. What do you think?

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -936,12 +937,59 @@ public boolean deleteDefaultAcl(String bucket, Entity entity) {
 
   @Override
   public Acl createDefaultAcl(String bucket, Acl acl) {
-    return throwNotYetImplemented(fmtMethodName("createDefaultAcl", String.class, Acl.class));
+    return updateDefaultAcl(bucket, acl);
```
sydney-munro, line 940: why do we need to call update here? do we not have a create?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITAccessTest.java
```diff
@@ -185,12 +188,108 @@ public void bucket_defaultAcl_list_bucket404() {
...
+    assertThat(storageException.getCode()).isEqualTo(404);
+  }
+
   @Test
   @CrossRun.Ignore(transports = Transport.GRPC)
   public void testBucketDefaultAcl() {
     // TODO: break this test up into each of the respective scenarios
     //   2. Delete a default ACL for a specific entity
-    //   3. Create a default ACL for specific entity
     //   4. Update default ACL to change role of a specific entity
```
sydney-munro, line 293: since update tests are done as well we can remove this too.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
BenWhitehead: The machinery of updating an acl or creating one is the same: 
1. Read Bucket metadata
2. possibly remove existing acl for entity
3. create new list from 2 + `acl`
4. Call update on bucket
5. Find new ObjectAccessControl in response
6. convert from proto to our model
7. return

Give that, I'm simply calling update

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITAccessTest.java
BenWhitehead: This whole test is removed in the `deleteDefaultAcl` PR.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
sydney-munro: understood.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITAccessTest.java
sydney-munro: LGTM, will let @frankyn ship as he still has an active comment thread.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
frankyn: Not sure it's necessary tbh, you can only have up to 100 acls associated to a bucket or object; 
image

Thinking we just rely on the service response cod in this case unless I'm misunderstanding.

BenWhitehead: I spoke with @frankyn offline, we came to the consensus that instead of returning an HTTP status code here, insead we will return undefined `0` to signal that it's purely a client side issue, not any system issue.

frankyn (approved):

PR #1809, opened 2022-12-09
chore(test): reduce the amount of extraneous logging when running the test suite

BenWhitehead:
Many informational items logged during the run of the test suite aren't useful unless debugging a specific failure. By default, the majority of informational logging is reduced unless the environment variable (or system property) `CI_VERBOSE_RUN` is set to `true`.

This also extends to things like non-applicable generated cases for ITRetryConformanceTest. The result is a much easier to skim build log.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/TimestampCodecPropertyTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/TimestampCodecPropertyTest.java
@@ -37,7 +37,7 @@ final class TimestampCodecPropertyTest {
     report(TypeUsage.of(Timestamp.class), StorageArbitraries.timestamp());
   }
 
-  @Property(tries = 50_000)
+  @Property
   void timestampCodecShouldRoundTrip(@ForAll(supplier = Supp.class) Timestamp ts) {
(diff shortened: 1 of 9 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/TimestampCodecPropertyTest.java
```diff
@@ -37,7 +37,7 @@ void edgeCases() {
     report(TypeUsage.of(Timestamp.class), StorageArbitraries.timestamp());
   }
 
-  @Property(tries = 50_000)
```
frankyn, line 40: This is a random change, what happened?

frankyn (approved):
LGTM, I just had a question about annotation removal.

google-cloud-storage/src/test/java/com/google/cloud/storage/TimestampCodecPropertyTest.java
BenWhitehead: The previous 50k was accidentally left in, and I noticed while doing the other cleanup in this PR so I removed it at the same time.

PR #1819, opened 2022-12-16
chore(test): expand StorageInstance vetoing to apply to bucket acl operations

BenWhitehead:
Simplify StorageInstance vetoing implementation. Rather than using a runtime proxy to do invocation method argument matching we are now defining a concrete class to override the methods it needs to provide enforcement for.

This is simpler in that it uses standard java language implementation approach which is helped by compiler and IDEs. It is at the expense of needing to define the new ~500 line AbstractStorageDecorator class.

Delete several obsolete tests, which have been replace by other tests or do not make sense to be written the way they are.

Cleanup of TODO from https://github.com/googleapis/java-storage/pull/1785

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/StorageInstance.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/StorageInstance.java
@@ -17,6 +17,8 @@
 package com.google.cloud.storage.it.runner.registry;
 
 import com.google.cloud.Policy;
+import com.google.cloud.storage.Acl;
+import com.google.cloud.storage.Acl.Entity;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/StorageInstance.java
```diff
@@ -103,9 +67,119 @@ private static VetoedBucketUpdateException err(String bucketName) {
...
+    }
+
+    @Override
+    public boolean deleteAcl(String bucket, Entity entity, BucketSourceOption... options) {
+      checkBucketProtected(bucket);
+      return super.deleteAcl(bucket, entity, options);
+    }
+
+    @Override
+    public boolean deleteAcl(String bucket, Entity entity) {
```
frankyn, line 135: By having this protection, I'm concerned about future changes not being caught by this mechanism. However, IIUC, whenever new methods are added to the Storage interface, the new methods will need to be implemented in the Vetoing class as well so it covers this case.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/StorageInstance.java
```diff
@@ -48,45 +47,10 @@ public Object get() {
...
-                  throw err(bucketName);
-                }
-              }
-              try {
-                return method.invoke(storage, args);
-              } catch (InvocationTargetException e) {
-                throw e.getCause();
-              }
-            }));
+    proxy = new VetoingStorageProxy();
```
frankyn, line 53: Side bar: 
I'm wondering if you're planning on exposing AbstractStorageProxy as a builder option that can be passed and potentially used for other efforts (thinking observability)?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/StorageInstance.java
BenWhitehead: Correct, by explicitly defining it as an implementation of the interface we will have to implement the method if a new one is added, whereas the runtime generated proxy before was using reflection to selectively choose which methods to augment.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/StorageInstance.java
BenWhitehead: At this point, no. I think observability stuff we're going to want/need to be more explicit and prescriptive about any extension points we provide. That said, for our internal implementation we could so something similar to allow for a single implementation that would wrap both http and grpc Stoage.

frankyn (approved):

PR #1826, opened 2022-12-22
test: add new Generator#randomObjectName

BenWhitehead:
In order to facilitate running tests in parallel, even across configurations object names need to be unique to a test. This new method encapsulates reliably generating test unique object names.

Generated tests names will look of the form: `testName[HTTP][PROD]-0001`

Replace usages of `TestName#getMethodName()` with `Generator#randomObjectName()`

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/Generator.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/Generator.java
@@ -17,18 +17,35 @@
 package com.google.cloud.storage.it.runner.registry;
 
 import java.util.UUID;
+import java.util.WeakHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.junit.runner.Description;
 
 /** Scoped instance of a class which generates common things for tests. */
 public final class Generator implements ManagedLifecycle {
 
-  /**
-   * Generate a new random bucket name
-   *
-   * @return
-   */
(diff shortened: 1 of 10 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/Generator.java
```diff
@@ -17,20 +17,37 @@
...
   }
 
+  @NonNull
+  public String randomObjectName() {
+    Description currentTest = Registry.getInstance().getCurrentTest();
+    if (currentTest == null) {
+      throw new IllegalStateException("No actively running test in registry.");
+    }
+    AtomicInteger counter = counters.computeIfAbsent(currentTest, (d) -> new AtomicInteger(1));
+    return String.format("%s-%04d", currentTest.getMethodName(), counter.getAndIncrement());
```
sydney-munro, line 48: why the counter over a UUID?

BenWhitehead: An attempt at some level of traceability on objects, in case they are left behind in a bucket we'll have some idea of where they came from. The counters is because some of the tests do create several so wanted to support that. While maintaining the traceability down to where some tests do specific operations to specific objects. Using a UUID would be opaque to most of these, so I opted for this approach.

sydney-munro: gotcha, that makes sense.

sydney-munro (approved):
LGTM

PR #1824, opened 2022-12-21
fix: update gRPC object list implementation to include synthetic directories

BenWhitehead:
Rework list centric tests in ITObjectTest to have more straightforward implementations and assertions, and in some cases a more clear name of what is being validated. (Removed all the list retry stuff, object list is strongly consistent in GCS.)

PR #1827, opened 2022-12-22
test: migrate tests which previously had hard coded object names to use new Generator#randomObjectName()

BenWhitehead:
Follow up to #1826 

Migrate tests which previously had hard coded object names to use new `Generator#randomObjectName()`.

Serval test which were previously using `@Rule Timeout` or `@Test(timeout = N)` have had their timeouts removed. The Timeout implementation in JUnit 4 is not thread safe, and can result in running test tracking issues.

The majority of these tests have had their timeouts made completely obsolete in #1824 and their change here will be short lived.

Each commit in this PR is migrating a single IT class.


frankyn (approved):
Love the random object generator!

PR #1830, opened 2022-12-28
fix: fix object metadata and bucket label change tracking to be fine-grained

BenWhitehead:
### Context
Historically, when calling storage#update(BlobInfo) or storage#update(BucketInfo) the whole info would be packed and sent to the backend as a PATCH request. The backend would then perform some server side diffing against what it already had and make modifications. Of note here, for Object#metadata and Bucket#labels the PATCH behavior meant that any value provided in the request was always treated as an append.

#### Code sample
Specifically, the following call sequence will actually append `key2: value2` to the object metadata despite nothing in the method names communicating this.

```java
try (Storage storage = StorageOptions.http().build().getService()) {
  BlobInfo info = BlobInfo.newBuilder("bucket", "object")
      .setMetadata(ImmutableMap.of("key1", "value1"))
      .build();
  Blob gen1 = storage.create(info);

  BlobInfo modified = gen1.toBuilder().setMetadata(ImmutableMap.of("key2", "value2")).build();
  Blob gen2 = storage.update(modified, BlobTargetOption.metagenerationMatch());

  // gen2.metadata = {"key1": "value1", "key2": "value2"}
}
```

Additionally, this now leaves the internal state of `modified` such that if you were to call `modified.getMetadata()` it would only contain the new `key2: value2`. However, after calling `storage.update` `gen2.getMetadata()` both keys are present.

This is confusing, and can be remedied now that we have the capability of tracking field level modification.

Another confusing thing about this: if you want to clear the metadata on an object, the intuitive `.setMetadata(Collections.emptyMap())` does not work. Instead, you would need to do the following:

```java
try (Storage storage = StorageOptions.http().build().getService()) {
  Blob gen1 = storage.get("bucket", "object");

  // Can't use ImmutableMap here, as we need null values
  HashMap<String, String> clearedKeys = new HashMap<>();
  for (String key : gen1.getMetadata().keySet()) {
    clearedKeys.put(key, null);
  }

  BlobInfo modified = gen1.toBuilder().setMetadata(clearedKeys).build();
  Blob gen2 = storage.update(modified, BlobTargetOption.metagenerationMatch());
}
```

### The Change
Update the behavior of `BlobInfo#setMetadata` and `BucketInfo#setLabels` to always explicitly set to the specified value, whether adding, removing or updating a key.

Add new methods `BlobInfo#addAllMetadata(Map)` and `BucketInfo#addAllLabels(Map)` to allow performing an upsert without the need to first query and reconcile.

##### Clearing all object metadata
```java
try (Storage storage = StorageOptions.http().build().getService()) {
  Blob gen1 = storage.get("bucket", "object");
  BlobInfo modified = gen1.toBuilder().setMetadata(Collections.emptyMap()).build();
  Blob gen2 = storage.update(modified, BlobTargetOption.metagenerationMatch());
}
```

##### Adding/modifying object metadata
```java
try (Storage storage = StorageOptions.http().build().getService()) {
  Blob gen1 = storage.get("bucket", "object");
  BlobInfo modified = gen1.toBuilder().addAllMetadata(ImmutableMap.of("key1", "value2", "key3", "value3").build();
  Blob gen2 = storage.update(modified, BlobTargetOption.metagenerationMatch());
}
```

##### Removing an individual metadata key
```java
try (Storage storage = StorageOptions.http().build().getService()) {
  Blob gen1 = storage.get("bucket", "object");
  HashMap<String, String> map = new HashMap<>();
  map.put("key1", null);
  BlobInfo modified = gen1.toBuilder().addAllMetadata(map).build();
  Blob gen2 = storage.update(modified, BlobTargetOption.metagenerationMatch());
}
```

### Implementation Details

1. Field modification now tracks nested metadata/label keys
   * New UnifiedOpts.NestedNamedField has been added to track nested keys
   * UpdateMaskTest has been updated to expect nested field names
2. `StorageImpl#update(BlobInfo)` and `StorageImpl#update(BucketInfo)` have both been updated to take field modification into consideration.
3. New test class `ITNestedUpdateMaskTest` has been added to validate many permutations of modifying metadata/labels
4. Each of the metadata/label related methods have had `@NonNull`/`@Nullable` annotations to their signatures to communicate acceptable arguments, and to set expectations.
5. All `Data.isNull`/`Data.nullOf` has been moved away from Blob metadata and Bucket labels into `ApiaryConversions` as they only apply to Apiary usage not gRPC/Protobuf.
6. Several easymock test in StorageImplTest have been wholly deleted as they weren't testing behavior, only wiring mocks to dance to the current implementation.
7. Update BucketInfoTest deep equals assertions to use `TestUtils.assertAll`

### Samples

Update samples to use new `addAll*` methods for their upsert use cases rather than `set*`.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -44,7 +44,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * An object in Google Cloud Storage. A {@code Blob} object includes the {@code BlobId} instance,
@@ -360,7 +362,14 @@ public class Blob extends BlobInfo {
     }
 
     @Override
-    public Builder setMetadata(Map<String, String> metadata) {
+    @BetaApi
+    public Builder addAllMetadata(@Nonnull Map<@NonNull String, @Nullable String> newMetadata) {
+      infoBuilder.addAllMetadata(newMetadata);
+      return this;
+    }
+
+    @Override
+    public Builder setMetadata(@Nullable Map<@NonNull String, @Nullable String> metadata) {
       infoBuilder.setMetadata(metadata);
       return this;
     }
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -19,9 +19,11 @@ package com.google.cloud.storage;
 import static com.google.cloud.storage.BackwardCompatibilityUtils.millisOffsetDateTimeCodec;
 import static com.google.cloud.storage.BackwardCompatibilityUtils.millisUtcCodec;
 import static com.google.cloud.storage.BackwardCompatibilityUtils.nullableDurationSecondsCodec;
+import static com.google.cloud.storage.Utils.diffMaps;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.Lists.newArrayList;
+import static java.util.Objects.requireNonNull;
 
 import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.api.client.util.Data;
@@ -29,12 +31,13 @@ import com.google.api.client.util.DateTime;
 import com.google.api.core.BetaApi;
 import com.google.api.services.storage.model.Bucket.Lifecycle.Rule;
 import com.google.cloud.storage.Acl.Entity;
+import com.google.cloud.storage.BlobInfo.ImmutableEmptyMap;
 import com.google.cloud.storage.Storage.BucketField;
 import com.google.cloud.storage.TransportCompatibility.Transport;
+import com.google.cloud.storage.UnifiedOpts.NamedField;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Maps;
 import com.google.common.collect.Streams;
 import java.io.IOException;
 import java.io.ObjectInputStream;
@@ -43,6 +46,7 @@ import java.io.Serializable;
 import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -100,7 +104,7 @@ public class BucketInfo implements Serializable {
   private final String location;
   private final Rpo rpo;
   private final StorageClass storageClass;
-  private final Map<String, String> labels;
+  @Nullable private final Map<String, String> labels;
   private final String defaultKmsKeyName;
   private final Boolean defaultEventBasedHold;
   private final OffsetDateTime retentionEffectiveTime;
@@ -111,7 +115,7 @@ public class BucketInfo implements Serializable {
   private final String locationType;
   private final Logging logging;
   private final CustomPlacementConfig customPlacementConfig;
-  private final transient ImmutableSet<Storage.BucketField> modifiedFields;
+  private final transient ImmutableSet<NamedField> modifiedFields;
 
   /**
    * non-private for backward compatibility on message class. log messages are now emitted from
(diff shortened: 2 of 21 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -360,7 +362,14 @@ Builder setMediaLink(String mediaLink) {
     }
 
     @Override
-    public Builder setMetadata(Map<String, String> metadata) {
+    @BetaApi
+    public Builder addAllMetadata(@Nonnull Map<@NonNull String, @Nullable String> newMetadata) {
```
BenWhitehead, line 366: I'm not tied to this name, would `upsertMetadataWith` be better?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1476,8 +1480,27 @@ Builder setUpdateTimeOffsetDateTime(OffsetDateTime updateTime) {
...
+    /**
+     * Convenience method to add new labels values to any existing present labels.
+     *
+     * <p>If a key from {@code newLabels} overlaps with any existing key in {@link #getLabels()},
+     * the new value will overwrite the old value.
+     *
+     * @see #setLabels(Map)
+     */
+    @BetaApi
+    public abstract Builder addAllLabels(@NonNull Map<@NonNull String, @Nullable String> newLabels);
```
BenWhitehead, line 1492: I'm not tied to this name, would `upsertLabelsWith` be better?

google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -100,16 +99,9 @@ public class BucketInfoTest {
...
+      TestUtils.hashMapOf("label1", "value1", "label2", null);
 
-  static {
-    BUCKET_LABELS = new HashMap<>();
-    BUCKET_LABELS.put("label1", "value1");
-    BUCKET_LABELS.put("label2", null);
-  }
-
-  private static final Map<String, String> BUCKET_LABELS_TARGET =
-      ImmutableMap.of("label1", "value1", "label2", "");
```
BenWhitehead, line 112: This was forcing `Data.nullOf` into the BucketInfo internals leaking apiary specifics. Those apiary specific have been pushed down to ApiaryConversions leaving BucketInfo able to do java standard things.

google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -210,39 +202,40 @@ public void testOf() {
...
-    assertEquals(BUCKET_LABELS_TARGET, BUCKET_INFO.getLabels());
-    assertEquals(REQUESTER_PAYS, BUCKET_INFO.requesterPays());
-    assertEquals(DEFAULT_EVENT_BASED_HOLD, BUCKET_INFO.getDefaultEventBasedHold());
-    assertEquals(RETENTION_EFFECTIVE_TIME, BUCKET_INFO.getRetentionEffectiveTime());
-    assertEquals(RETENTION_PERIOD, BUCKET_INFO.getRetentionPeriod());
-    assertEquals(RETENTION_POLICY_IS_LOCKED, BUCKET_INFO.retentionPolicyIsLocked());
-    assertTrue(LOCATION_TYPES.contains(BUCKET_INFO.getLocationType()));
-    assertEquals(LOGGING, BUCKET_INFO.getLogging());
+  public void testBuilder() throws Exception {
+    assertAll(
```
BenWhitehead, line 206: Make sure all assertions run

google-cloud-storage/src/test/java/com/google/cloud/storage/BucketInfoTest.java
```diff
@@ -260,37 +253,44 @@ public void testToPbAndFromPb() {
...
-    assertEquals(expected.getDefaultEventBasedHold(), value.getDefaultEventBasedHold());
-    assertEquals(
-        expected.getRetentionEffectiveTimeOffsetDateTime(),
-        value.getRetentionEffectiveTimeOffsetDateTime());
-    assertEquals(expected.getRetentionPeriodDuration(), value.getRetentionPeriodDuration());
-    assertEquals(expected.retentionPolicyIsLocked(), value.retentionPolicyIsLocked());
-    assertEquals(expected.getLogging(), value.getLogging());
-    assertEquals(expected, value);
+  private void compareBuckets(BucketInfo expected, BucketInfo value) throws Exception {
+    assertAll(
```
BenWhitehead, line 257: Make sure all assertions run

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITNestedUpdateMaskTest.java
```diff
@@ -0,0 +1,195 @@
...
+ * BlobInfo#getMetadata()} and {@link BucketInfo#getLabels()} and the various permutations which can
+ * be used to add and remove keys.
+ */
+@RunWith(StorageITRunner.class)
+@CrossRun(
+    backends = Backend.PROD,
+    transports = {Transport.HTTP, Transport.GRPC})
+@Parameterized(NestedUpdateMaskParametersProvider.class)
+@ParallelFriendly
+public final class ITNestedUpdateMaskTest {
```
BenWhitehead, line 61: Before adding this test class, only 1 case was being explicitly tested in the existing suite - Object add new key. Now we're testing permutations of add/update/remove for both objects and buckets.

frankyn:
Hey @BenWhitehead thanks for the summary.

Changing the behavior of setMetadata is a behavior breaking change, this can impact anyone that upgrades based on minor and patch versions. I agree that this is confusing and could be seen as a bug but it has been in the library for years. 

Are you planning on lumping this into the next major version release and just didn't call it out in the PR?

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -44,7 +44,9 @@
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
```
burkedavison, line 47: Intentional use of both `javax.annotation.Nonnull` and `org.checkerframework.checker.nullness.qual.NonNull`?

BenWhitehead:
Closing in favor of maintaining the existing `setMetadata` behavior.

PR #1832, opened 2023-01-06
chore: introduce ByteRangeSpec#endOffsetInclusive

BenWhitehead:
There are some cases where endOffset and endOffsetInclusive are both necessary and externally being able to differentiate is needed.

### Test Refactor

As part of this change, I've update the declaration of ByteRangeSpec.RangeScenario to be more verbose and easier to read, as well as have a more direct description for the test name.

This does add a fair number of lines of code, but I think it brings some clarity compared to a method with 4 different long arguments. It also reduces (and possible errors in declaration) between related scenarios.

In general the new pattern is: define the four properties of any range (beginOffset, endOffset, endOffsetInclusive, length) then specify those ByteRangeSpecs for which it should be applicable.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/ByteRangeSpecTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/ByteRangeSpecTest.java
@@ -16,8 +16,10 @@
 
 package com.google.cloud.storage;
 
+import static com.google.cloud.storage.ByteRangeSpec.EFFECTIVE_INFINITY;
 import static com.google.cloud.storage.TestUtils.assertAll;
 import static com.google.common.truth.Truth.assertThat;
+import static java.util.Objects.requireNonNull;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
@@ -27,6 +29,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.util.Arrays;
 import java.util.stream.Stream;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.junit.Test;
@@ -172,19 +175,19 @@ public final class ByteRangeSpecTest {
     @Test
     public void withNewEndOffset_sameInstanceIfNotDifferent_null() {
       ByteRangeSpec spec = ByteRangeSpec.relativeLength(null, null);
-      assertThat(spec.withNewEndOffset(RangeScenarios.INF)).isSameInstanceAs(spec);
+      assertThat(spec.withNewEndOffset(EFFECTIVE_INFINITY)).isSameInstanceAs(spec);
     }
 
     @Test
     public void withNewEndOffsetClosed_sameInstanceIfNotDifferent_null() {
       ByteRangeSpec spec = ByteRangeSpec.relativeLength(null, null);
-      assertThat(spec.withNewEndOffsetClosed(RangeScenarios.INF)).isSameInstanceAs(spec);
+      assertThat(spec.withNewEndOffsetClosed(EFFECTIVE_INFINITY)).isSameInstanceAs(spec);
     }
 
     @Test
     public void withNewRelativeLength_sameInstanceIfNotDifferent_null() {
       ByteRangeSpec spec = ByteRangeSpec.relativeLength(null, null);
-      assertThat(spec.withNewRelativeLength(RangeScenarios.INF)).isSameInstanceAs(spec);
+      assertThat(spec.withNewRelativeLength(EFFECTIVE_INFINITY)).isSameInstanceAs(spec);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/ByteRangeSpecTest.java
```diff
@@ -441,129 +444,168 @@ private static void threeWayEqual(
...
+      // expect that by default, a range should be from zero to infinity
       Stream<RangeScenario> bothNullOrEmpty =
-          Stream.of(
+          RangeScenario.expectThat()
+              .beginOffset(0L)
+              .endOffset(EFFECTIVE_INFINITY)
+              .endOffsetInclusive(EFFECTIVE_INFINITY)
+              .length(EFFECTIVE_INFINITY)
+              .httpRange(null)
+              .isApplicableTo(
```
BenWhitehead, line 490: The new builder which allows this more explicit description of the tests is the majority of the change in this PR.

google-cloud-storage/src/test/java/com/google/cloud/storage/ByteRangeSpecTest.java
```diff
@@ -572,80 +614,166 @@ public static Iterable<Object[]> testCases() {
...
+      }
+
+      private static String fmt(@Nullable Long l) {
+        if (l == null) {
+          return null;
+        } else if (l == Long.MAX_VALUE) {
+          return "Long.MAX_VALUE";
+        } else {
+          long diff = Long.MAX_VALUE - l;
+          if (diff <= 20) {
```
frankyn, line 719: Could you explain what this conditional is doing, it's looks random.

google-cloud-storage/src/test/java/com/google/cloud/storage/ByteRangeSpecTest.java
```diff
@@ -441,129 +444,168 @@ private static void threeWayEqual(
...
+                  .bounded()
+                  .isApplicableTo(
+                      ByteRangeSpec.relativeLength(null, effectiveMax_0),
+                      ByteRangeSpec.explicit(null, effectiveMax_0),
+                      ByteRangeSpec.explicitClosed(null, effectiveMax_1)),
+              // expect that, we can read from 1 up to Long.MAX_VALUE - 1
+              RangeScenario.expectThat()
+                  .beginOffset(1L)
+                  .endOffset(effectiveMax_0)
+                  .endOffsetInclusive(effectiveMax_1)
```
frankyn, line 602: Is "inclusive" in terms of math or that the endOffset is part of the range e.g. `%d-%d`? If it's the former, I would think inclusive would be `effectiveMax_0` and not `effectiveMax_1`.

frankyn (commented):
Couple of questions, otherwise ready to approve.

google-cloud-storage/src/test/java/com/google/cloud/storage/ByteRangeSpecTest.java
BenWhitehead: The inclusive value would be the second arg in the header. So it's inclusive in terms of the index being included (`i <= endOffsetInclusive`), whereas endOffset would be excluded (`i < endOffset`).

frankyn: Converting into an example because my brain turned to mush; Am I understanding this correctly?

endOffset(N):
`Range: bytes=0-(N-1)` where N is exclusive it goes from (0-N-1)

endOffsetInclusive(N-1):
`Range: bytes=0-(N+1)` where N is inclusive it goes from (0-N)

BenWhitehead: Imagine we have three ranges, A, B, C:
* A: `[0, N)` aka The set of all x such that 0 less than or equal to x and x less than N (`{x | 0 <= x < N}`)
* B: `[0, N]` aka The set of all x such that 0 less than or equal to x and x less than N (`{x | 0 <= x <= N}`)
* C: `0 + N` aka relative length

When converting A to an http range header, we know that the upper bound is exclusive, therefore we would produce `bytes=0-(N-1)`. For a range with an exclusive upper bound, `endOffset` will always equal `N` and `endOffsetInclusive` will equal `N-1`.

Contrasted by B, the upper bound of an inclusive range would produce `bytes=0-N`. Where `endOffset` will always equal `N+1` and `endOffsetInclusive` will equal `N`.

In the case of C, the upper bound is calculated relative to the beginning and would produce `bytes=0-(0+N-1)`. The `endOffset` will always be equal to `begin + length` whereas the `endOffsetInclusive` will be `begin + length - 1`.

This test is trying to say "Expect a range where the being offset is 1, the endOffset is effectiveMax_0, the endOffsetInclusive is effectiveMax_1, and the length is effectiveMax_1" and this should hold true for all three range declarations `[1 + effectiveMax_1]`, `[1, effectiveMax_0)` and `[0, effectiveMax_1]`.

frankyn (approved):

PR #1834, opened 2023-01-09
feat: update GrpcBlobReadChannel to allow seek/limit after read

BenWhitehead:
### Refactoring
Extract the majority of BlobReadChannelV2 to a new abstract base class BaseStorageReadChannel which effectively adapts a LazyReadChannel to a StorageReadChannel.

BaseStorageReadChannel now defines an abstract method newLazyReadChannel which each implementation is responsible for implementing to integrate into the lifecycle.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BaseStorageReadChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BaseStorageReadChannel.java
@@ -127,12 +127,12 @@ abstract class BaseStorageReadChannel<T> implements StorageReadChannel {
 
   protected abstract LazyReadChannel<T> newLazyReadChannel();
 
-  private void maybeResetChannel(boolean umallocBuffer) throws IOException {
+  private void maybeResetChannel(boolean freeBuffer) throws IOException {
     if (lazyReadChannel != null && lazyReadChannel.isOpen()) {
       try (BufferedReadableByteChannel ignore = lazyReadChannel.getChannel()) {
-        if (bufferHandle != null && !umallocBuffer) {
+        if (bufferHandle != null && !freeBuffer) {
           bufferHandle.get().clear();
-        } else if (umallocBuffer) {
+        } else if (freeBuffer) {
           bufferHandle = null;
         }
         lazyReadChannel = null;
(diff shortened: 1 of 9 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BaseStorageReadChannel.java
```diff
@@ -0,0 +1,149 @@
...
+        // BlobReadChannel to believe it read 0 bytes, returning -1 and leaving the channel open.
+        // Emulate that same behavior here to preserve behavior compatibility, though this should
+        // be removed in the next major version.
+        return -1;
+      } else {
+        throw new IOException(e);
+      }
+    } catch (IOException e) {
+      throw e;
+    } catch (Exception e) {
```
frankyn, line 103: This is broad; is there a specific set of Exception types that are expected?

google-cloud-storage/src/main/java/com/google/cloud/storage/BaseStorageReadChannel.java
```diff
@@ -0,0 +1,149 @@
...
+    return resolvedObject;
+  }
+
+  protected void setResolvedObject(@Nullable T resolvedObject) {
+    this.resolvedObject = resolvedObject;
+  }
+
+  protected abstract LazyReadChannel<T> newLazyReadChannel();
+
+  private void maybeResetChannel(boolean umallocBuffer) throws IOException {
```
frankyn, line 130: what does umallocBuffer mean in this case; is it just to effectively delete the old buffer and start with a new one?

If that's the case, can you name it `freeBuffer`?

frankyn (commented):
I saw that the PR is mainly a refactor while enabling seek after read in gRPC; just have 2 follow-up questions otherwise LGTM.

google-cloud-storage/src/main/java/com/google/cloud/storage/BaseStorageReadChannel.java
BenWhitehead: For grpc, the scope is more narrow, but for JSON there isn't really a common base since things can be socket issues, ssl issues or any other number of exceptions. 

This catch is primarily here to guarantee we're always doing our best to present an actionable exception to whatever is catching.

google-cloud-storage/src/main/java/com/google/cloud/storage/BaseStorageReadChannel.java
BenWhitehead: Yes, I'll rename to freeBuffer

frankyn (approved):

PR #1839, opened 2023-01-10
test: enable S.7 retry conformance tests

BenWhitehead:
Update method mappings to properly exclude retry conformance tests for Bucket#create.

Update TestRetryConformance to lazily allocate and fill a temporary file when necessary.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
@@ -30,17 +30,16 @@ import com.google.cloud.conformance.storage.v1.RetryTest;
 import com.google.cloud.conformance.storage.v1.RetryTests;
 import com.google.cloud.storage.CIUtils;
 import com.google.cloud.storage.Storage;
-import com.google.cloud.storage.TransportCompatibility.Transport;
 import com.google.cloud.storage.conformance.retry.Functions.CtxFunction;
 import com.google.cloud.storage.conformance.retry.ITRetryConformanceTest.RetryConformanceParameterProvider;
 import com.google.cloud.storage.it.runner.StorageITRunner;
 import com.google.cloud.storage.it.runner.annotations.Backend;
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
```diff
@@ -78,7 +77,8 @@
  * RpcMethodMappings}.
  */
 @RunWith(StorageITRunner.class)
-@CrossRun(transports = Transport.HTTP, backends = Backend.TEST_BENCH)
+// @CrossRun(transports = Transport.HTTP, backends = Backend.TEST_BENCH)
+@SingleBackend(Backend.TEST_BENCH)
```
sydney-munro, line 81: why are we switching this to only run against the testbench?

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/RpcMethodMappings.java
```diff
@@ -1795,7 +1796,8 @@ private static void insert(ArrayList<RpcMethodMapping> a) {
                 .build());
         a.add(
             RpcMethodMapping.newBuilder(119, objects.insert)
-                .withApplicable(TestRetryConformance::isPreconditionsProvided)
+                .withApplicable(
+                    and(TestRetryConformance::isPreconditionsProvided, not(groupIsResumableUpload)))
```
sydney-munro, line 1800: so this is here because we cannot have both preconditions and resumable uploads? sorry if this is naive I do not really get what this is doing.

google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
BenWhitehead: Previously it was only running against testbench (in fact, it can only run against test bench due to the failure instructions).

Since, the testbench doesn't yet support grpc. I replaced the `@CrossRun` with `@SingleBackend` to be more clear about how things will run.

sydney-munro (approved):

PR #1843, opened 2023-01-11
fix: update GrpcStorageImpl#update to support fine-grained update of BucketInfo.labels and BlobInfo.metadata

BenWhitehead:
Successor to #1830, whereas this preserves existing behavior rather than introducing a new method.

A new set of tests have been added to validate all permutations of modifying metadata/labels.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
@@ -17,6 +17,7 @@
 package com.google.cloud.storage;
 
 import static com.google.cloud.storage.BackwardCompatibilityUtils.millisOffsetDateTimeCodec;
+import static com.google.cloud.storage.Utils.diffMaps;
 import static com.google.common.base.MoreObjects.firstNonNull;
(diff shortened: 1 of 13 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -331,7 +335,7 @@ public Builder setTimeStorageClassUpdatedOffsetDateTime(
     }
 
     /** Sets the blob's user provided metadata. */
-    public abstract Builder setMetadata(Map<String, String> metadata);
+    public abstract Builder setMetadata(@Nullable Map<@NonNull String, @Nullable String> metadata);
```
frankyn, line 338: If a dev is extended BlobInfo and they overload; they'll need to add @Nullable as well IIUC?

BenWhitehead: The annotations are optional from the actual runtime standpoint, this is putting a label to the existing behavior that wasn't explicitly expressed anywhere.

frankyn (approved):

PR #1845, opened 2023-01-11
chore: refactor BlobReadChannel to plumb the sparse object metadata included in a read media response for both json and grpc

BenWhitehead:
Publicly exposing this is gated on how/when we are able to expose StorageReadChannel.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
@@ -237,12 +237,26 @@ final class GapicUnbufferedReadableByteChannel
 
     @Override
     public boolean hasNext() {
-      return ensureResponseIteratorOpen().hasNext();
+      try {
+        return ensureResponseIteratorOpen().hasNext();
+      } catch (RuntimeException e) {
+        if (!result.isDone()) {
+          result.setException(StorageException.coalesce(e));
+        }
+        throw e;
+      }
     }
 
     @Override
     public ReadObjectResponse next() {
-      return ensureResponseIteratorOpen().next();
+      try {
+        return ensureResponseIteratorOpen().next();
+      } catch (RuntimeException e) {
+        if (!result.isDone()) {
+          result.setException(StorageException.coalesce(e));
+        }
+        throw e;
+      }
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
```diff
@@ -237,12 +237,26 @@ private final class LazyServerStreamIterator implements Iterator<ReadObjectRespo
 
     @Override
     public boolean hasNext() {
-      return ensureResponseIteratorOpen().hasNext();
+      try {
+        return ensureResponseIteratorOpen().hasNext();
+      } catch (RuntimeException e) {
+        if (!result.isDone()) {
+          result.setException(StorageException.coalesce(e));
```
sydney-munro, line 244: why do we set the exception on the result only if the result is not complete? What case is this covering?

google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
```diff
@@ -237,12 +237,26 @@ private final class LazyServerStreamIterator implements Iterator<ReadObjectRespo
...
     }
 
     @Override
     public ReadObjectResponse next() {
-      return ensureResponseIteratorOpen().next();
+      try {
+        return ensureResponseIteratorOpen().next();
+      } catch (RuntimeException e) {
+        if (!result.isDone()) {
+          result.setException(StorageException.coalesce(e));
```
sydney-munro, line 256: same question as above.

google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
BenWhitehead: A call to `hasNext()`/`next()` could error any time. More often than not, as long as the object exists, this code will not run, but when the object doesn't exist this ensures we communicate that to the result as well as any call to `read`

google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
BenWhitehead: same answer as above

sydney-munro (approved):

PR #1846, opened 2023-01-11
test: add test to verify `lifecycle.rule.condition.age_days = 0`

BenWhitehead:
Zero (0) is a special value for some serialization configurations. Add a test to ensure it is properly passed to the backend, and is effective.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
@@ -714,8 +714,12 @@ final class GrpcConversions {
               .collect(ImmutableList.toImmutableList());
       conditionBuilder.setMatchesStorageClass(collect);
     }
-    conditionBuilder.setMatchesPrefix(condition.getMatchesPrefixList());
-    conditionBuilder.setMatchesSuffix(condition.getMatchesSuffixList());
+    if (!condition.getMatchesPrefixList().isEmpty()) {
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -714,8 +714,12 @@ private BucketInfo.LifecycleRule lifecycleRuleDecode(Bucket.Lifecycle.Rule from)
               .collect(ImmutableList.toImmutableList());
       conditionBuilder.setMatchesStorageClass(collect);
     }
-    conditionBuilder.setMatchesPrefix(condition.getMatchesPrefixList());
-    conditionBuilder.setMatchesSuffix(condition.getMatchesSuffixList());
+    if (!condition.getMatchesPrefixList().isEmpty()) {
```
frankyn, line 717: this change looks out of context, is it needed as well?

BenWhitehead: It's a difference between what is constructed via 
```
 LifecycleRule d1 =
        new LifecycleRule(
            LifecycleAction.newAbortIncompleteMPUploadAction(),
            LifecycleCondition.newBuilder().setAge(0).build());
```

And what is decoded coming back from the service via grpc. By simply setting as we were doing before the value would be empty list rather than null. Apiary on the other hand will be null by default. This change brings these both in line with the apiary behavior.

frankyn (approved):

PR #1867, opened 2023-01-20
fix: update BlobReadChannelV2 handling to correctly restart for decompressed object

BenWhitehead:
When downloading bytes from gcs and decompressing them, a restart of the stream needs to pickup from the offset within the compressed bytes not the decompressed bytes.

Prior to this change http-client was automatically applying gzip decompression to the stream it returns to us thereby causing our tracking to be off.

This change updates our interaction with http client to always request the raw bytes without any transform applied to them, we then at a higher level can handle whether gzip decompression needs to be plumbed in.

Fix up a couple channel closed state subtleties when buffering is used with decompression.

Add a new test leveraging the testbench which forces a broken stream on some compressed object bytes. _NOTE_ This test depends on the next release of testbench and will be failing until we get that release.


sydney-munro (approved):
LGTM

PR #1872, opened 2023-01-25
test: add new StorageNativeCanaryTest

BenWhitehead:
Update native-image related config to be specific for this class.

Update maven config to explicitly configure this new class as it's only class for test when run under -Pnative

Prior this change, the native image pre-submit builds will run the full integration test suite 2 times, 1 time to discover the tests which should be run during `native-test` and then again run during `native-test`.

We don't need the whole integration suite ran multiple times, or even to verify native image has built a working program. Instead, we have a specific scenario which exercises all possible request types ensuring each request type is covered. This new CanaryTest runs everything in ~30 seconds, whereas the full integration suite is ~13 min on current CI machines.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
@@ -130,8 +130,8 @@ public final class StorageNativeCanary {
           },
           () -> assertThat(actual.get(0).getContent()).isEqualTo(bytes),
           () -> assertThat(actual.get(1).getContent()).isEqualTo(bytes),
-          () -> assertThat(deletes.get(0)).isFalse(),
-          () -> assertThat(deletes.get(1)).isFalse());
+          () -> assertThat(deletes.get(0)).isTrue(),
+          () -> assertThat(deletes.get(1)).isTrue());
     }
   }
 
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
@@ -39,10 +39,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
-import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
-import java.util.function.Predicate;
 import org.junit.Test;
 
 // Intentionally avoid StorageITRunner here. It touches lots of code at a semi-static level making
@@ -130,15 +128,10 @@ public final class StorageNativeCanary {
 
             assertThat(actualNames).containsExactly(info1.getName(), info2.getName());
           },
-          () -> {
-            // if the content isn't equal carry it through
-            List<BlobWithContent> contentNotEquals =
-                actual.stream()
-                    .filter(bwc -> !Arrays.equals(bwc.getContent(), bytes))
-                    .collect(ImmutableList.toImmutableList());
-            assertThat(contentNotEquals).isEmpty();
-          },
-          () -> assertThat(deletes.stream().anyMatch(isFalse())).isFalse());
+          () -> assertThat(actual.get(0).getContent()).isEqualTo(bytes),
+          () -> assertThat(actual.get(1).getContent()).isEqualTo(bytes),
+          () -> assertThat(deletes.get(0)).isFalse(),
+          () -> assertThat(deletes.get(1)).isFalse());
     }
   }
 
@@ -159,12 +152,8 @@ public final class StorageNativeCanary {
       ByteStreams.copy(r, w);
       return new BlobWithContent(info, baos.toByteArray());
     } catch (IOException e) {
-      throw new RuntimeException(e);
+      throw new RuntimeIOException(e);
     }
-  }
-
-  private static Predicate<Boolean> isFalse() {
-    return b -> !b;
   }
 
   private static final class BlobWithContent {
@@ -182,6 +171,12 @@ public final class StorageNativeCanary {
 
     public byte[] getContent() {
       return content;
+    }
+  }
+
+  private static final class RuntimeIOException extends RuntimeException {
+    private RuntimeIOException(IOException cause) {
+      super(cause);
     }
   }
 }
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
@@ -47,18 +47,17 @@ import org.junit.Test;
 // native-test have a hard time.
 public final class StorageNativeCanary {
 
-  private static final int _512KiB = 512 * 1024;
   private static final int _256KiB = 256 * 1024;
-  private static final byte[] bytes = DataGenerator.base64Characters().genBytes(_512KiB);
+  private static final byte[] bytes = DataGenerator.base64Characters().genBytes(512 * 1024);
 
   @Test
   public void canary_happyPath_http() throws Exception {
-    canary_happyPath(StorageOptions.http().build().getService());
+    assertBehaviorOfPrimaryStorageActions(StorageOptions.http().build().getService());
   }
 
   @Test
   public void canary_happyPath_grpc() throws Exception {
-    canary_happyPath(StorageOptions.grpc().build().getService());
+    assertBehaviorOfPrimaryStorageActions(StorageOptions.grpc().build().getService());
   }
 
   /**
@@ -85,14 +84,14 @@ public final class StorageNativeCanary {
    *   <li>Delete temporary bucket (Unary)
    * </ul>
    */
-  private static void canary_happyPath(Storage storage) throws Exception {
+  private static void assertBehaviorOfPrimaryStorageActions(Storage storage) throws Exception {
     // create a temporary bucket
-    try (TemporaryBucket tempB =
+    try (TemporaryBucket temporaryBucket =
         TemporaryBucket.newBuilder()
             .setStorage(storage)
             .setBucketInfo(BucketInfo.of("java-storage-grpc-" + UUID.randomUUID()))
             .build()) {
-      String bucketName = tempB.getBucket().getName();
+      String bucketName = temporaryBucket.getBucket().getName();
       String obj1Name = UUID.randomUUID().toString();
       String obj2Name = UUID.randomUUID().toString();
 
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,187 @@
...
+          },
+          () -> {
+            // if the content isn't equal carry it through
+            List<BlobWithContent> contentNotEquals =
+                actual.stream()
+                    .filter(bwc -> !Arrays.equals(bwc.getContent(), bytes))
+                    .collect(ImmutableList.toImmutableList());
+            assertThat(contentNotEquals).isEmpty();
+          },
+          () -> assertThat(deletes.stream().anyMatch(isFalse())).isFalse());
```
mpeddada1, line 141: Same as above. Perhaps we can make these assertions more explicit? This may help us understand the expected outcome easier.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,187 @@
...
+    try (ReadChannel r = storage.reader(info.getBlobId(), BlobSourceOption.generationMatch());
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        WritableByteChannel w = Channels.newChannel(baos)) {
+      // only buffer up to half the object
+      r.setChunkSize(_256KiB);
+      ByteStreams.copy(r, w);
+      return new BlobWithContent(info, baos.toByteArray());
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
mpeddada1, line 163: Could we make the thrown exception a little more specific or perhaps add a `throws IOException` to the method signature?  This could help with debugging and also prevent the caller of this method from catching unrelated exceptions since a`RuntimeException` can happen for a number of reasons. 

https://effectivejava.googleplex.com/third-edition#page=323

mpeddada1 (commented):
Thank you so much for piloting the idea of a canary @BenWhitehead! Added the first round of comment.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,187 @@
...
+          blobs.stream()
+              .map(info -> readAll(storage, info))
+              .collect(ImmutableList.toImmutableList());
+
+      List<Boolean> deletes =
+          blobs.stream()
+              .map(b -> storage.delete(b.getBlobId(), BlobSourceOption.generationMatch()))
+              .collect(ImmutableList.toImmutableList());
+
+      assertAll(
```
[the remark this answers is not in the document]
BenWhitehead: This is a util method we've added in storage that ensures all of the assertions contained within it are ran and collected if they fail. This will report multiple failures if multiple assertions fail.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
BenWhitehead: In this case the RuntimeException wrapping is necessary because this method is called within a lambda which can have a checked exception. The fact that this wrapping is taking place for a private class in a private utility that doesn't escape this class I'm not sure a custom exception type provides much value, but I can change it.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,187 @@
...
+                    .map(BlobId::getName)
+                    .collect(ImmutableList.toImmutableList());
+
+            assertThat(actualNames).containsExactly(info1.getName(), info2.getName());
+          },
+          () -> {
+            // if the content isn't equal carry it through
+            List<BlobWithContent> contentNotEquals =
+                actual.stream()
+                    .filter(bwc -> !Arrays.equals(bwc.getContent(), bytes))
```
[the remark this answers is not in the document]
BenWhitehead: removed code, no longer necessary.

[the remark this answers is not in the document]
BenWhitehead: done

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
BenWhitehead: Done

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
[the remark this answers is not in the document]
mpeddada1: Thank you!

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
mpeddada1: Hm I see. How about we do something like `RuntimeException('couldn't read foo', e)` instead? where the message gives some context into where the try block went wrong.

BenWhitehead: We're carrying the cause through to be the cause of the runtime exception, so any exception that does pop out will have the original stacktrace with it. I've changed the thrown exception to be an internal `RuntimeIOException` to disambiguate from some other RuntimeException

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,182 @@
...
+   * <ul>
+   *   <li>Create a temporary bucket (Unary)
+   *   <li>Insert two (2) objects (Unary, ServerStream)
+   *   <li>List all objects, using a pageSize of 1 (Page over Unary)
+   *   <li>Read all bytes of each object (ServerStream)
+   *   <li>Delete each object (Unary)
+   *   <li>Delete temporary bucket (Unary)
+   * </ul>
+   */
+  private static void canary_happyPath(Storage storage) throws Exception {
```
mpeddada1, line 88: How about we rename this to something like `assertBehaviorOfPrimaryStorageActions`? Since the details of what we are testing are abstracted over by the helper method, the name can help us determine that.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,182 @@
...
+    // create a temporary bucket
+    try (TemporaryBucket tempB =
```
mpeddada1, line 90:
```suggestion
    try (TemporaryBucket temporaryBucket =
```

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,182 @@
...
+  private static void uploadUsingWriter(Storage storage, BlobInfo info) throws IOException {
+    try (WriteChannel w = storage.writer(info, BlobWriteOption.doesNotExist())) {
+      // set our size to the smallest resumable size, so we can send multiple requests
+      w.setChunkSize(_256KiB);
+      ByteStreams.copy(Channels.newChannel(new ByteArrayInputStream(bytes)), w);
+    }
+  }
+
+  private static BlobWithContent readAll(Storage storage, BlobInfo info) {
+    try (ReadChannel r = storage.reader(info.getBlobId(), BlobSourceOption.generationMatch());
```
mpeddada1, line 147:
```suggestion
    try (ReadChannel readChannel = storage.reader(info.getBlobId(), BlobSourceOption.generationMatch());
```

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,182 @@
...
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
mpeddada1, line 148: optional: how about this? 
```suggestion
        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,182 @@
...
+import java.nio.channels.WritableByteChannel;
+import java.util.List;
+import java.util.UUID;
+import org.junit.Test;
+
+// Intentionally avoid StorageITRunner here. It touches lots of code at a semi-static level making
+// native-test have a hard time.
+public final class StorageNativeCanary {
+
+  private static final int _512KiB = 512 * 1024;
```
mpeddada1, line 50: This variable is only used once in this class, how about we inline this in Line 52?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,182 @@
...
+          },
+          () -> assertThat(actual.get(0).getContent()).isEqualTo(bytes),
+          () -> assertThat(actual.get(1).getContent()).isEqualTo(bytes),
+          () -> assertThat(deletes.get(0)).isTrue(),
+          () -> assertThat(deletes.get(1)).isTrue());
+    }
+  }
+
+  private static void uploadUsingWriter(Storage storage, BlobInfo info) throws IOException {
+    try (WriteChannel w = storage.writer(info, BlobWriteOption.doesNotExist())) {
```
mpeddada1, line 139:
```suggestion
    try (WriteChannel writeChannel = storage.writer(info, BlobWriteOption.doesNotExist())) {
```

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
mpeddada1: Got it! That makes sense, thanks for the context!

mpeddada1 (commented):
This is looking great! Only minor comments remaining.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/StorageNativeCanary.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.List;
+import java.util.UUID;
+import org.junit.Test;
+
+// Intentionally avoid StorageITRunner here. It touches lots of code at a semi-static level making
+// native-test have a hard time.
+public final class StorageNativeCanary {
+
+  private static final int _512KiB = 512 * 1024;
+  private static final int _256KiB = 256 * 1024;
```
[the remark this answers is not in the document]
BenWhitehead: For gcs 256 KiB is an important number, as it is the minimum number of bytes which can be in a resumable upload chunk and still keep the resumable session active. The fact that we are setting the chunk size to this value is important and I feel the code reads more like it is important having the name it does.

`readChannel.setChunkSize(_256KiB);` is more clear about the boundary which is being set ran `readChannel.setChunkSize(CHUNK_SIZE);`.

mpeddada1 (approved):

cojenco (approved):

PR #1876, opened 2023-02-01
test: update ObjectsFixture to allow discrimination by @BucketFixture

BenWhitehead:
Sometimes it's useful to specify that the ObjectsFixture should use the RequesterPays bucket

This change is needed from a followup PR which performs some operations on a requester pays bucket, but itself does not need to create objects.


sydney-munro (approved):
LGTM

PR #1887, opened 2023-02-06
chore(docs): update README

alicejli:
README sample was missing a `</dependencies>`


BenWhitehead:
I believe this segment of the readme is actually a code snippet https://github.com/googleapis/java-storage/blob/main/samples/snippets/pom.xml#L28-L46. Do other libraries do this differently?

danielduhh:
Looks like this fix was picked up elsewhere

PR #1922, opened 2023-03-02
feat: First Pass Implementation of UploadMany

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
@@ -16,4 +16,45 @@
 
 package com.google.cloud.storage.transfermanager;
 
-public class TransferManagerImpl {}
+import com.google.cloud.storage.BlobInfo;
+import com.google.common.collect.ImmutableList;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+public final class TransferManagerImpl implements TransferManager {
+
+  private final TransferManagerConfig transferManagerConfig;
+  private final ExecutorService executor;
+
+  TransferManagerImpl(TransferManagerConfig transferManagerConfig) {
+    this.transferManagerConfig = transferManagerConfig;
+    this.executor = Executors.newFixedThreadPool(transferManagerConfig.getMaxWorkers());
+  }
+
+  @Override
+  public @NonNull UploadJob uploadFiles(List<Path> files, ParallelUploadConfig opts) {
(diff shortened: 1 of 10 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
```diff
@@ -16,4 +16,46 @@
...
+  TransferManagerImpl(TransferManagerConfig transferManagerConfig) {
+    this.transferManagerConfig = transferManagerConfig;
+    this.executor = Executors.newFixedThreadPool(transferManagerConfig.getMaxWorkers());
+  }
+
+  @Override
+  public @NonNull UploadJob uploadFiles(List<Path> files, ParallelUploadConfig opts) {
+    List<Future<UploadResult>> uploadTasks = new ArrayList<>();
+    for (Path file : files) {
+      if (Files.isDirectory(file)) throw new IllegalStateException("Directories are not supported");
```
danielduhh, line 44: just to clarify -- we plan to support Directory upload in the future?

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/UploadCallable.java
```diff
@@ -0,0 +1,87 @@
...
+    this.originalBlob = originalBlob;
+    this.sourceFile = sourceFile;
+  }
+
+  public UploadResult call() throws Exception {
+    // TODO: Check for chunking
+    return uploadWithoutChunking();
+  }
+
+  private UploadResult uploadWithoutChunking() throws IOException {
```
danielduhh, line 58: Once we support MPU, it would be great if we could determine how (single shot vs resumable vs MPU) to upload based on the object size.

danielduhh (commented):
Generally lgtm; just had a couple questions

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
sydney-munro: The idea is that the api will just accept file paths but we will "support" directories in the samples by showing the customers how to generate that list themselves from a directory.

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/UploadCallable.java
sydney-munro: yes, I think this is the goal. also throughput based. If we are uploading super fast no sense in spending the overhead to split the work.

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
JesseLovelace: Are we planning to have `TransferManagerUtils` be public to consumers? If so, should we consider a method like `TransferManagerUtils.getFileListForDirectory(Path directory)` that does it for you?

JesseLovelace (approved):

PR #1939, opened 2023-03-16
docs: add clarification on non-editable metadata

JesseLovelace:
Adds a note about non-editable metadata to all CreateBlob documentation. This addresses internal bug 273252496.


sydney-munro (approved):
LGTM

PR #1958, opened 2023-03-29
feat: implement GrpcStorageImpl#{get,list,create,delete}Notification

BenWhitehead:
Rewrite ITNotificationTest to leverage testbench for grpc and to separate different cases into their own individual tests rather than having a single large test.

For now the backend hasn't yet implemented the rpcs for grpc, so we rely on testbench. Once the backend does provide we can remove the `CrossRun.Ignore` annotation on each of the methods.


sydney-munro (approved):
LGTM

PR #1965, opened 2023-04-04
feat: Add matchGlob parameter to BlobListOption

briantruong777:
This is a new feature (b/236167515) where List Objects allows filtering results based on a provided glob pattern.

Currently only the JSON API supports the matchGlob parameter, so using the option with gRPC will throw an exception. gRPC support is planned for later.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
@@ -1084,7 +1084,8 @@ final class UnifiedOpts {
 
     @Override
     public Mapper<ListObjectsRequest.Builder> listObjects() {
-      throw new UnsupportedOperationException("Option matchGlob is not supported for gRPC yet.");
+      return GrpcStorageImpl.throwHttpJsonOnly(
+          com.google.cloud.storage.Storage.BlobListOption.class, "matchGlob(String)");
     }
   }
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
@@ -413,6 +413,11 @@ final class UnifiedOpts {
     return new KmsKeyName(kmsKeyName);
   }
 
+  static MatchGlob matchGlob(@NonNull String glob) {
+    requireNonNull(glob, "glob must be non null");
+    return new MatchGlob(glob);
+  }
+
   static Md5Match md5Match(@NonNull String md5) {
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -1070,6 +1075,19 @@ public Mapper<RewriteObjectRequest.Builder> rewriteObject() {
...
+  static final class MatchGlob extends RpcOptVal<String> implements ObjectListOpt {
+    private static final long serialVersionUID = 8819855597395473178L;
+
+    private MatchGlob(String val) {
+      super(StorageRpc.Option.MATCH_GLOB, val);
+    }
+
+    @Override
+    public Mapper<ListObjectsRequest.Builder> listObjects() {
+      throw new UnsupportedOperationException("Option matchGlob is not supported for gRPC yet.");
```
BenWhitehead, line 1087: We have a helper method to construct an appropriate exception with a pointer on how they can use the method.

```suggestion
      return GrpcStorageImpl.throwHttpJsonOnly(
          com.google.cloud.storage.Storage.BlobListOption.class, "matchGlob(String)");
```

Using this, the exception will have a message of
```
java.lang.UnsupportedOperationException: com.google.cloud.storage.Storage$BlobListOption#matchGlob(String) is only supported for HTTP_JSON transport. Please use StorageOptions.http() to construct a compatible instance.
```

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITObjectTest.java
```diff
@@ -542,6 +544,60 @@ public void testListBlobsCurrentDirectoryIncludesBothObjectsAndSyntheticDirector
...
+                  .containsExactly("foo/baz")
+                  .inOrder(),
+          () ->
+              assertThat(Iterables.transform(page6.iterateAll(), blob -> blob.getName()))
+                  .containsExactly("foo/baz", "foo/foobar", "foobar")
+                  .inOrder(),
+          () ->
+              assertThat(Iterables.transform(page7.iterateAll(), blob -> blob.getName()))
+                  .containsExactly("foo/baz", "foo/foobar")
+                  .inOrder());
```
BenWhitehead, line 597: Looks great, thank you!

BenWhitehead (changes requested):
One small change to the exception construction.

briantruong777:
Updated to use `throwHttpJsonOnly()`. PTAL.

BenWhitehead (approved):

BenWhitehead:
Change not impacted by owlbot, merging.

PR #1972, opened 2023-04-05
Add an example of how to change the oauth endpoint

DKbyo:
Fixes #1971

### DIFF
--- a/samples/snippets/src/main/java/com/example/storage/bucket/SetClientEndpoint.java
+++ b/samples/snippets/src/main/java/com/example/storage/bucket/SetClientEndpoint.java
@@ -17,11 +17,26 @@
 package com.example.storage.bucket;
 
 // [START storage_set_client_endpoint]
-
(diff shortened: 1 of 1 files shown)


samples/snippets/src/main/java/com/example/storage/bucket/SetClientEndpoint.java
```diff
@@ -17,21 +17,44 @@
...
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageOptions;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
 
 public class SetClientEndpoint {
 
+  public static ServiceAccountCredentials setOAuthEndpoint(
```
sydney-munro, line 30: Generally we would make a seperate sample (including a seperate class, test etc) if we were going to add a sample like this and not add it into the existing sample for another use case.

PR #1974, opened 2023-04-10
feat: make it possible to disable the buffer of ReadChannels returned from Storage.reader

BenWhitehead:
For some scenarios, an external client needs the ability to manage buffer blocking itself. To support this, providing 0 to ReadChannel#setChunkSize will disable buffering allowing for client to avoid the need to alight multiple levels of buffer alignments.

Because the buffering is disabled, that means reads can be much more variable in size and individually more impacted by small network latencies (with buffering these can be amortized by the buffer making followup read faster). This is considered advanced usage and will require more work from the client integrator.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageReadChannelTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageReadChannelTest.java
@@ -42,6 +42,7 @@ import com.google.cloud.storage.it.runner.annotations.Inject;
 import com.google.cloud.storage.it.runner.registry.Generator;
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.nio.channels.ReadableByteChannel;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageReadChannelTest.java
```diff
@@ -107,6 +108,35 @@ public void storageReadChannel_getObject_returns() throws Exception {
     }
   }
 
+  @Test
+  // @CrossRun.Exclude(transports = Transport.GRPC)
```
sydney-munro, line 112: nit: can we remove this line?

sydney-munro (approved):
minor comment feel free to merge once acknowledged.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITStorageReadChannelTest.java
BenWhitehead: Yeah, I'll add it's removal in my next PR (don't want to waste all the cpu time to do a full rebuild and test to remove this comment in a testclass)

PR #1977, opened 2023-04-11
fix: update GrpcStorageOptions to set allowNonDefaultServiceAccount = true

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -211,6 +211,7 @@ private Tuple<StorageSettings, Opts<UserProject>> resolveSettingsAndOpts() throw
     InstantiatingGrpcChannelProvider.Builder channelProviderBuilder =
         InstantiatingGrpcChannelProvider.newBuilder()
             .setEndpoint(endpoint)
+            .setAllowNonDefaultServiceAccount(true)
```
danielduhh, line 214: Do we think this should always be true? Should this be configurable?

danielduhh (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
BenWhitehead: My feeling, is this flag is there to support some old client which didn't have this capability. However, in the case of storage I think we always want to allow any kind of credential without requiring the customer to also know they need to flip this flag. I mean hell, I didn't even know this flag existed until dataproc tried to use the library and was surprised they weren't getting direct path and tracked it down to the credentials.

PR #1994, opened 2023-04-24
fix: make Blob and Bucket update diff aware

BenWhitehead:
If a Blob or Bucket update does not actually contain a diff, instead of sending the empty update refresh the metadata.

Add integration tests to verify behavior of updating object or bucket with no modification

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -146,6 +146,7 @@ import java.util.function.UnaryOperator;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 @BetaApi
 final class GrpcStorageImpl extends BaseService<StorageOptions> implements Storage {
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -515,7 +493,11 @@ public Bucket update(BucketInfo bucketInfo, BucketTargetOption... options) {
 
   @Override
   public Blob update(BlobInfo blobInfo, BlobTargetOption... options) {
-    Opts<ObjectTargetOpt> opts = Opts.unwrap(options).resolveFrom(blobInfo).prepend(defaultOpts);
+    Opts<ObjectTargetOpt> unwrap = Opts.unwrap(options);
+    if (blobInfo.getModifiedFields().isEmpty()) {
+      return internalBlobGet(blobInfo.getBlobId(), unwrap.constrainTo(ObjectSourceOpt.class));
```
sydney-munro, line 498: so the behavior here is if the fields are not being modified we just perform a get and return the unmodified blob?

BenWhitehead: Correct, while attempting to use those provided options that also can apply to Get

sydney-munro (approved):

PR #2022, opened 2023-05-15
fix: make GrpcBlobWriteChannel open upon construction

BenWhitehead:
Add new integration tests for both json and grpc to verify openness

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
@@ -18,12 +18,13 @@ package com.google.cloud.storage;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
-import com.google.api.services.storage.model.StorageObject;
 import com.google.cloud.ReadChannel;
 import com.google.cloud.WriteChannel;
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
```diff
@@ -59,11 +60,26 @@ public static Blob blobCopyWithStorage(Blob b, Storage s) {
     return new Blob(s, builder);
   }
 
-  public static Function<WriteChannel, Optional<StorageObject>> maybeGetStorageObjectFunction() {
+  public static Function<WriteChannel, Optional<BlobInfo>> maybeGetStorageObjectFunction() {
```
sydney-munro, line 63: nit: change the method name
```suggestion
  public static Function<WriteChannel, Optional<BlobInfo>> maybeGetBlobInfoFunction() {
```

sydney-munro (approved):

google-cloud-storage/src/test/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
BenWhitehead: Good catch, fix pushed.

PR #2027, opened 2023-05-22
chore: add new request idempotency header x-goog-gcs-idempotency-token

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/Retrying.java
```diff
@@ -115,6 +120,14 @@ static <T, U> U run(
     }
   }
 
+  @NonNull
+  static GrpcCallContext newCallContext() {
```
tritone, line 124: Does this need a comment or more descriptive name clarifying that it adds the idempotency header? I think it could be confusing otherwise.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcIdempotencyTokenTest.java
```diff
@@ -0,0 +1,168 @@
...
+    IterableSubject subject = requestAuditing.assertRequestHeader(X_GOOG_GCS_IDEMPOTENCY_TOKEN);
+
+    byte[] actual = storage.readAllBytes(info.getBlobId());
+    String actualXxd = xxd(actual);
+
+    // We expect 4 distinct requests:
+    //  1. start resumable session
+    //  2. PUT first 256KiB
+    //  3. PUT second 256KiB
+    //  4. Finalize session and put final 45B
```
tritone, line 165: Just figured I'd ask because I am not great at following some of this testing code -- does the test assert that each of these 4 gets its own request ID, and that all retries for each request have the same ID?

tritone (approved):
Overall looks great, couple small questions based on my weak understanding of some java stuff...

google-cloud-storage/src/main/java/com/google/cloud/storage/Retrying.java
BenWhitehead: Since this method is attached to the `Retrying` class, I feel it carries the context with it that the `GrpcCallContext` returned from this method contains things related to retries. I think the idempotency token is an implementation detail of what goes into the context.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcIdempotencyTokenTest.java
BenWhitehead: Mostly correct, we are asserting there is a distinct token for each of the 4 requests that are sent, however we are not inducing any failures in these tests to verify tokens are maintained for the same request across multiple attempts. Instead we are waiting on https://github.com/googleapis/storage-testbench/issues/486 for the retry enforcement piece.

PR #2043, opened 2023-06-02
feat: Error handling for Downloads

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
@@ -86,17 +86,6 @@ public final class PackagePrivateMethodWorkarounds {
     };
   }
 
-  public static Function<WriteChannel, Optional<BlobInfo>> maybeGetBlobInfoFunction() {
-    return writeChannel -> {
-      Optional<StorageObject> so = maybeGetStorageObjectFunction().apply(writeChannel);
-      if (so.isPresent()) {
-        return Optional.of(Conversions.apiary().blobInfo().decode(so.get()));
-      } else {
-        return Optional.empty();
-      }
-    };
-  }
-
   public static ApiFuture<BlobInfo> getBlobInfoFromReadChannelFunction(ReadChannel c) {
     if (c instanceof StorageReadChannel) {
       StorageReadChannel src = (StorageReadChannel) c;
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/DirectDownloadCallable.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/DirectDownloadCallable.java
@@ -49,7 +49,7 @@ final class DirectDownloadCallable implements Callable<DownloadResult> {
   @Override
   public DownloadResult call() {
     Path path = TransferManagerUtils.createDestPath(parallelDownloadConfig, originalBlob);
-    long bytesCopied = 0L;
+    long bytesCopied = -1L;
     try (ReadChannel rc =
         storage.reader(
             BlobId.of(parallelDownloadConfig.getBucketName(), originalBlob.getName()), opts)) {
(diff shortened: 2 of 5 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/DirectDownloadCallable.java
```diff
@@ -50,16 +49,26 @@ final class DirectDownloadCallable implements Callable<DownloadResult> {
   @Override
   public DownloadResult call() {
     Path path = TransferManagerUtils.createDestPath(parallelDownloadConfig, originalBlob);
-    try (ReadChannel rc = storage.reader(originalBlob.getBlobId(), opts)) {
+    long bytesCopied = 0L;
```
BenWhitehead, line 52: It would be better to initialize this to `-1` or something less than zero instead of `0`. There can be objects that are `0` bytes, so that would actually be a success, as long as the file is created on disk (something else to double check).

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITTransferManagerTest.java
```diff
@@ -253,6 +253,23 @@ public void uploadNonexistentFile() throws Exception {
     }
   }
 
+  @Test
+  public void downloadNonexistentBucket() throws Exception {
+    TransferManagerConfig config =
+        TransferManagerConfigTestingInstances.defaults(storage.getOptions());
+    try (TransferManager transferManager = config.getService()) {
+      String bucketName = "this-bucket-does-not-exist";
```
BenWhitehead, line 261: Maybe include a uuid in this name, to prevent the case of someone creating this bucket and breaking our tests.

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
```diff
@@ -86,17 +86,6 @@ public static Function<WriteChannel, Optional<BlobInfo>> maybeGetBlobInfoFunctio
     };
   }
 
-  public static Function<WriteChannel, Optional<BlobInfo>> maybeGetBlobInfoFunction() {
```
BenWhitehead, line 89: In `main` this method is still used in two locations. How is it's removal here not causing a failure?

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITTransferManagerTest.java
```diff
@@ -253,6 +253,49 @@ public void uploadNonexistentFile() throws Exception {
...
+      ParallelDownloadConfig parallelDownloadConfig =
+          ParallelDownloadConfig.newBuilder()
+              .setBucketName(bucketName)
+              .setDownloadDirectory(baseDir)
+              .build();
+      DownloadJob job = transferManager.downloadBlobs(blobs, parallelDownloadConfig);
+      List<DownloadResult> downloadResults = ApiFutures.allAsList(job.getDownloadResults()).get();
+      assertThat(downloadResults).hasSize(3);
+      try {
+        assertThat(downloadResults.get(0).getStatus()).isEqualTo(TransferStatus.FAILED_TO_START);
```
BenWhitehead, line 292: What status do we expect for the other 2 results?

BenWhitehead (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
sydney-munro: its a duplicate that was not removed during the merge up from main, we don't need two implementations of the same method.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITTransferManagerTest.java
sydney-munro: same status, I can go ahead and check for those as well.

PR #2047, opened 2023-06-05
fix: include invocation-id for resumable PUTs

BenWhitehead:
Change BlobWriteChannel to us Retrying.run so it will receive an invocation id

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.storage;
 
-import static com.google.cloud.RetryHelper.runWithRetries;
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java
```diff
@@ -186,7 +186,9 @@ private static StorageException unrecoverableState(
   @Override
   protected void flushBuffer(final int length, final boolean lastChunk) {
     try {
-      runWithRetries(
+      Retrying.run(
```
frankyn, line 189: Am IIUC runWithRetries did not use InvocationId support and using Retrying.run() will address that?

BenWhitehead: correct

frankyn (approved):

PR #2059, opened 2023-06-09
feat: update Storage.createFrom(BlobInfo, Path) to have 150% higher throughput

BenWhitehead:
When uploading a file where we are able to rewind to an arbitrary offset, we can be more optimistic in the way we send requests to GCS.

Add new code middleware to allow PUTing an entire file to GCS in a single request, and using query resumable session to recover from the specific offset in the case of retryable error.

#### Methodology

Generate a random file on disk of size `128KiB..2GiB` from `/dev/urandom`, then upload the generated file using `Storage.createFrom(BlobInfo, Path)`.

Perform each 4096 times.

Run on a c2-standard-60 instance is us-central1 against a regional bucket located in us-central1.

#### Results

The following summary of throughput in MiB/s as observed between the existing implementation, and the new implementation proposed in this PR.

```
                                count     mean     std    min      50%      75%      90%      99%      max
runId                 ApiName                                                                             
createFrom - existing JSON     4096.0   66.754  10.988  3.249   67.317   73.476   78.961   91.197  107.247
createFrom - new      JSON     4096.0  158.769  67.105  4.600  170.680  218.618  240.992  266.297  305.205
```

#### Comparison

When comparing the new implementation to the existing implementation we get the following improvement to throughput (higher is better):
```
stat      pct
mean  137.841
 50%  153.547
 90%  205.204
 99%  192.003
```

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
@@ -186,7 +186,6 @@ enum JsonResumableSessionFailureScenario {
     if (suppress != null) {
       storageException.addSuppressed(suppress);
     }
-    storageException.printStackTrace(); // todo: remove this
     return storageException;
   }
 
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
@@ -1,0 +1,224 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage;
+
+import com.google.api.client.http.HttpHeaders;
+import com.google.api.client.http.HttpResponse;
+import com.google.api.client.http.HttpResponseException;
+import com.google.cloud.BaseServiceException;
+import com.google.cloud.storage.StorageException.IOExceptionCallable;
+import com.google.common.io.CharStreams;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Predicate;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+@ParametersAreNonnullByDefault
+enum JsonResumableSessionFailureScenario {
+  // TODO: send more bytes than are in the Content-Range header
+  SCENARIO_0(BaseServiceException.UNKNOWN_CODE, null, "Unknown Error"),
+  SCENARIO_0_1(BaseServiceException.UNKNOWN_CODE, null, "Response not application/json."),
+  SCENARIO_1(400, "invalid", "Attempt to append to already finalized resumable session."),
+  SCENARIO_2(
+      400,
+      "invalid",
+      "Attempt to finalize resumable session with fewer bytes than the backend has received."),
+  SCENARIO_3(
+      400,
+      "dataLoss",
+      "Attempt to finalize resumable session with more bytes than the backend has received."),
+  SCENARIO_4(200, "ok", "Attempt to finalize an already finalized session with same object size"),
+  SCENARIO_4_1(400, "dataLoss", "Finalized resumable session, but object size less than expected."),
+  SCENARIO_4_2(
+      400, "dataLoss", "Finalized resumable session, but object size greater than expected."),
+  SCENARIO_5(
+      400,
+      "dataLoss",
+      "Client side data loss detected. Attempt to append to a resumable session with an offset higher than the backend has"),
+  SCENARIO_7(
+      400, "dataLoss", "Client side data loss detected. Bytes acked is more than client sent."),
+  SCENARIO_9(503, "backendNotConnected", "Ack less than bytes sent"),
+  QUERY_SCENARIO_1(503, "", "Missing Range header in response");
+
+  private static final String PREFIX_I = "\t|< ";
+  private static final String PREFIX_O = "\t|> ";
+  private static final String PREFIX_X = "\t|  ";
+
+  private static final Predicate<String> includedHeaders =
+      matches("Content-Length")
+          .or(matches("Content-Encoding"))
+          .or(matches("Content-Range"))
+          .or(matches("Content-Type"))
+          .or(matches("Range"))
+          .or(startsWith("X-Goog-Stored-"))
+          .or(matches("X-GUploader-UploadID"));
+
+  private static final Predicate<Map.Entry<String, ?>> includeHeader =
+      e -> includedHeaders.test(e.getKey());
+
+  private final int code;
+  @Nullable private final String reason;
+  private final String message;
+
+  JsonResumableSessionFailureScenario(int code, @Nullable String reason, String message) {
+    this.code = code;
+    this.reason = reason;
+    this.message = message;
+  }
+
+  StorageException toStorageException(String uploadId, HttpResponse resp) {
+    return toStorageException(
+        uploadId, resp, null, () -> CharStreams.toString(new InputStreamReader(resp.getContent())));
+  }
+
+  StorageException toStorageException(
+      String uploadId, @Nullable HttpResponse resp, @Nullable Throwable cause) {
+    if (resp != null) {
+      // an exception caused this, do not try to read the content from the response.
+      return toStorageException(uploadId, resp, cause, () -> null);
+    } else {
+      return new StorageException(code, message, reason, cause);
+    }
+  }
+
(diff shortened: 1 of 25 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Predicate;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+@ParametersAreNonnullByDefault
+enum JsonResumableSessionFailureScenario {
+  // TODO: send more bytes than are in the Content-Range header
+  SCENARIO_0(BaseServiceException.UNKNOWN_CODE, null, "Unknown Error"),
```
frankyn, line 37: Could we label these with human readable error names?

google-cloud-storage/src/test/java/com/google/cloud/storage/FakeHttpServer.java
```diff
@@ -0,0 +1,144 @@
...
+import java.net.InetSocketAddress;
+import java.net.URI;
+
+final class FakeHttpServer implements AutoCloseable {
+
+  private final URI endpoint;
+  private final Channel channel;
+  private final Runnable shutdown;
+
+  private FakeHttpServer(URI endpoint, Channel channel, Runnable shutdown) {
```
frankyn, line 53: Should this eventually move into testbench?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
```diff
@@ -0,0 +1,807 @@
...
+
+      JsonResumableSessionPutTask task =
+          new JsonResumableSessionPutTask(
+              httpClientContext,
+              uploadUrl,
+              RewindableHttpContent.empty(),
+              HttpContentRange.of(ByteRangeSpec.explicitClosed(0L, 10L)));
+
+      StorageException se = assertThrows(StorageException.class, () -> task.call());
+      assertThat(se.getCode()).isEqualTo(503);
```
frankyn, line 151: Why is this 503 instead of 408?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
```diff
@@ -0,0 +1,807 @@
...
+      JsonResumableSessionPutTask task =
+          new JsonResumableSessionPutTask(
+              httpClientContext,
+              uploadUrl,
+              RewindableHttpContent.empty(),
+              HttpContentRange.of(ByteRangeSpec.explicitClosed(0L, 10L)));
+
+      StorageException se = assertThrows(StorageException.class, task::call);
+      assertThat(se.getCode()).isEqualTo(400);
+      assertThat(se.getReason()).isEqualTo("dataLoss");
```
frankyn, line 199: Is this an error through gRPC or JSON or both?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
```diff
@@ -0,0 +1,807 @@
...
+   * Content-Range: bytes 0-262143/*
+   *     </pre></td>
+   *   </tr>
+   *   <tr>
+   *     <td>response</td>
+   *     <td><pre>
+   * 200 OK
+   * Content-Type: application/json; charset=utf-8
+   *
+   * {"name": "obj", "size": 524288}
```
frankyn, line 237: Why is this OK when the test expected a 400?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
```diff
@@ -0,0 +1,807 @@
...
+      assertThat(confirmedBytes.get()).isEqualTo(-1L);
+    }
+  }
+
+  /**
+   *
+   *
+   * <h2>S.3</h2>
+   *
+   * Attempting to finalize a session with more bytes than GCS acknowledges.
```
frankyn, line 357: GCS expects instead of acknowledges?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
```diff
@@ -0,0 +1,807 @@
...
+      JsonResumableSessionPutTask task =
+          new JsonResumableSessionPutTask(
+              httpClientContext,
+              uploadUrl,
+              RewindableHttpContent.empty(),
+              HttpContentRange.of(_512KiBL));
+
+      StorageException se = assertThrows(StorageException.class, () -> task.call());
+      assertThat(se.getCode()).isEqualTo(400);
+      assertThat(se.getReason()).isEqualTo("dataLoss");
```
frankyn, line 573: Looking through these; looks like this error is emitted by client side logic.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
```diff
@@ -0,0 +1,807 @@
...
+      assertThat(confirmedBytes.get()).isEqualTo(-1);
+    }
+  }
+
+  /**
+   *
+   *
+   * <h2>S.5</h2>
+   *
+   * Attempt to append to a resumable session with an offset higher than GCS has
```
frankyn, line 662: than GCS expects

google-cloud-storage/src/main/java/com/google/cloud/storage/ResumableSession.java
```diff
@@ -0,0 +1,41 @@
...
+abstract class ResumableSession<T> {
+
+  ResumableSession() {}
+
+  abstract ResumableOperationResult<@Nullable T> put(
+      RewindableHttpContent content, HttpContentRange contentRange);
+
+  abstract ResumableOperationResult<@Nullable T> query();
+
+  // TODO: abort()
```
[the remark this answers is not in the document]
BenWhitehead: removing for now

google-cloud-storage/src/test/java/com/google/cloud/storage/FakeHttpServer.java
BenWhitehead: I think it's an interesting idea, but this fake server is better for forcing erroneous state rather than individual request/status code modificaiton.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
BenWhitehead: The reason field can be any string. In this case, from the perspective of the client there is data that has been lost, so the reason is set to such.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
BenWhitehead: The OK here is from the backend. (A session being finalized by another client, or delayed packets, etc) However, from the client it was performing an incremental PUT so it is impossible for the session to finalize, so we raise a client side error.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
BenWhitehead: Not expects, acknowledges. The response headers is `Range: bytes=0-262143`, which is the number of bytes that have been acknowledged.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
BenWhitehead: fixed

google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
BenWhitehead: possibly later, though I feel like the enum names will likely become unwieldy. For example, the messages in them are already pretty long.

google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Predicate;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+@ParametersAreNonnullByDefault
+enum JsonResumableSessionFailureScenario {
+  // TODO: send more bytes than are in the Content-Range header
```
frankyn, line 36: wouldn't the server just stop reading bytes after content-range is satisfied? Guessing this would be more of a dataloss scenario client side?

BenWhitehead: I spent some time thinking about this, and this would actually be a client side bug. Since the client computes both Content-Range and Content-Length if those don't sync up, we have a bug so not a server response failure we need to handle. I'll remove this TODO in the follow up PR I have queued up on my workstation to follow this one.

frankyn: Client side issue makes sense; if anything tests would serve to protect client changes that introduce such a bug.

PR #2062, opened 2023-06-12
feat: Exceptions for Chunked Downloads and Precondition Failure tests

sydney-munro:
Plumbing of exceptions for chunked downloads and snuck in a standard precondition failure test without skipIfExists set to true.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ChunkedDownloadCallable.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ChunkedDownloadCallable.java
@@ -20,9 +20,7 @@ import com.google.cloud.ReadChannel;
 import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.Storage.BlobSourceOption;
-import com.google.cloud.storage.StorageException;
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ChunkedDownloadCallable.java
```diff
@@ -57,17 +55,24 @@ final class ChunkedDownloadCallable implements Callable<DownloadSegment> {
...
     try (ReadChannel rc = storage.reader(originalBlob.getBlobId(), opts)) {
       FileChannel wc =
           FileChannel.open(destPath, StandardOpenOption.WRITE, StandardOpenOption.CREATE);
       rc.seek(startPosition);
       rc.limit(endPosition);
       wc.position(startPosition);
-      ByteStreams.copy(rc, wc);
-    } catch (IOException e) {
-      throw new StorageException(e);
+      bytesCopied = ByteStreams.copy(rc, wc);
```
BenWhitehead, line 66: Since we know how many bytes we expect we should validate we receive them here.

There are circumstances where we've seen a connection get truncated without throwing an error, and are not retried.

BenWhitehead (approved):
One additional failure case, but looks good otherwise.

PR #2063, opened 2023-06-13
feat: add new dedupe utility method to Option classes

BenWhitehead:
When providing options duplicates are not allowed, however there are times when you might end up providing two options which would collide. These new utility methods encapsulate the logic necessary to deduplicate any overlap wil returing an array which can be passed directly to the respective operation method.

`dedupe` utility methods added to each of the following Option classes:
* Storage.BlobGetOption.dedupe();
* Storage.BlobListOption.dedupe();
* Storage.BlobSourceOption.dedupe();
* Storage.BlobTargetOption.dedupe();
* Storage.BlobWriteOption.dedupe();
* Storage.BucketGetOption.dedupe();
* Storage.BucketListOption.dedupe();
* Storage.BucketSourceOption.dedupe();
* Storage.BucketTargetOption.dedupe();
* Storage.CreateHmacKeyOption.dedupe();
* Storage.DeleteHmacKeyOption.dedupe();
* Storage.GetHmacKeyOption.dedupe();
* Storage.ListHmacKeysOption.dedupe();
* Storage.UpdateHmacKeyOption.dedupe();
* Bucket.BlobTargetOption.dedupe();
* Bucket.BlobWriteOption.dedupe();
* Bucket.BucketSourceOption.dedupe();
* Blob.BlobSourceOption.dedupe();

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
@@ -40,6 +40,7 @@ import java.nio.file.Path;
 import java.security.Key;
 import java.time.OffsetDateTime;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


sydney-munro (approved):
LGTM, Thanks!

google-cloud-storage/src/main/java/com/google/cloud/storage/Option.java
```diff
@@ -29,4 +34,25 @@
...
+  static <O extends Option<?>> O[] dedupe(IntFunction<O[]> gen, Collection<O> collection, O... os) {
+    return dedupe(gen, Stream.of(collection.stream(), Arrays.stream(os)).flatMap(s -> s));
+  }
+
+  @SafeVarargs
+  static <O extends Option<?>> O[] dedupe(IntFunction<O[]> gen, O[] array, O... os) {
+    return dedupe(gen, Stream.of(Arrays.stream(array), Arrays.stream(os)).flatMap(s -> s));
+  }
+
+  private static <O extends Option<?>> O[] dedupe(IntFunction<O[]> gen, Stream<O> s) {
```
frankyn, line 53: IIUC this method will 1) group by option class type 2) get last instance from group then 3) return array of this filtering.

Please add a comment describing what each of these helpers does. Change LGTM.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
```diff
@@ -163,6 +164,40 @@ public static BlobSourceOption shouldReturnRawInputStream(boolean shouldReturnRa
       return new BlobSourceOption(UnifiedOpts.returnRawInputStream(shouldReturnRawInputStream));
     }
 
+    /**
+     * Deduplicate any options which are the same parameter. The value which comes last in {@code
+     * os} will be the value included in the return.
+     */
+    @BetaApi
+    public static BlobSourceOption[] dedupe(BlobSourceOption... os) {
```
frankyn, line 172: I see these helpers are public; I understand the purpose just not the experience. Please reply with a snippet of how code will look when using one of these?

google-cloud-storage/src/main/java/com/google/cloud/storage/Option.java
BenWhitehead: All `Option`s contain an `Opt`, `Opt`s are distinct classes allowing us to group based on those classes. Once grouped, we select the last element to provide last wins.

Each of these helpers is an internal implementation detail, primarily due to the fact that generic arrays can not be instantiated in Java and requires a factory to be passed in.

google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java
BenWhitehead: Imagine the case where you have some defaults you want to specify, but allow overriding in specific usage.

```java
final class X {
  private static final BlobWriteOption[] defaultWriteOptions = new BlobWriteOption[] {
      BlobWriteOption.doesNotExist(), 
      BlobWriteOption.userProject("proj")
  };

  public static void main(String[] args) throws Exception {
    try (Storage s = StorageOptions.http().build().getService()) {

      BlobInfo info = BlobInfo.newBuilder("bucket", "object").build();

      try (WriteChannel w = s.writer(
          info,
          BlobWriteOption.dedupe(
              defaultWriteOptions, 
              BlobWriteOption.generationMatch(3)
          )
      )) {
        w.write(ByteBuffer.wrap("abc".getBytes(StandardCharsets.UTF_8)));
      }
    }
  }
}
```

google-cloud-storage/src/main/java/com/google/cloud/storage/Option.java
frankyn: Thanks Ben, please add this context as comments in the code for future us.

PR #2064, opened 2023-06-13
feat: Resolve API futures on getDownload/UploadResults calls

sydney-munro:
For the public API we will resolve the API Futures and return the result (ie. await the resolution and return the result).


google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/DownloadJob.java
```diff
@@ -39,8 +41,8 @@ private DownloadJob(
     this.parallelDownloadConfig = parallelDownloadConfig;
   }
 
-  public @NonNull List<ApiFuture<DownloadResult>> getDownloadResults() {
-    return downloadResults;
+  public @NonNull List<DownloadResult> getDownloadResults() {
```
BenWhitehead, line 44: When we add javadocs, we will want to be sure to mention these methods are blocking.

BenWhitehead (approved):

PR #2086, opened 2023-06-21
fix: update UploadCallable to use createFrom to avoid NPE trying to resolve resulting object

sydney-munro:
This PR has a few parts to it, all small enough to where I did not feel the need to separate.

### Major Addition
There was a bug where the storageObject was being accessed before the WriteChannel closes, when the upload is not finished WriteChannel returns a null storageObject resulting in an NPE when we go to decode the storageObject. see [BlobWriteChannel#storageObject](https://github.com/googleapis/java-storage/blob/976b9ba367d5eb5bcd4a2b50ef6f530b3a6dff94/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java#L44) for where this is documented.

First Pass (up to commit [76462df](https://github.com/googleapis/java-storage/pull/2086/commits/76462df4649db6764043f1ca011e5d2d62a68ddd)): addressed this by trying to process the storageObject outside of the scope of the try block, which would ensure the write has completed. All we have to do is manually ensure that the channel is closed.

Refactor (commit [976b9ba](https://github.com/googleapis/java-storage/pull/2086/commits/976b9ba367d5eb5bcd4a2b50ef6f530b3a6dff94)): utilizes createFrom, this addition is to both relieve the upload call of handling the Write/FileChannels directly and also allows us to take advantage of any performance improvements being made. The drawback of this is we will regress on our TransferStatus reporting as we no longer have direct byte access to track upload progress. The plan is to short term handle this via Status code checks and long term add in progress monitoring.

### Minor Additions
- Add additional assertions to tests to prevent a bug like this from lurking about.
- Add the skipIfExists functionality

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
@@ -64,7 +64,6 @@ public final class PackagePrivateMethodWorkarounds {
     return (w) -> {
       BlobWriteChannel blobWriteChannel;
       if (w instanceof BlobWriteChannel) {
-
         blobWriteChannel = (BlobWriteChannel) w;
         return Optional.of(blobWriteChannel.getStorageObject())
             .map(Conversions.apiary().blobInfo()::decode);
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
@@ -64,6 +64,7 @@ public final class PackagePrivateMethodWorkarounds {
     return (w) -> {
       BlobWriteChannel blobWriteChannel;
       if (w instanceof BlobWriteChannel) {
+
         blobWriteChannel = (BlobWriteChannel) w;
         return Optional.of(blobWriteChannel.getStorageObject())
             .map(Conversions.apiary().blobInfo()::decode);
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/UploadCallable.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/UploadCallable.java
@@ -16,16 +16,12 @@
 
 package com.google.cloud.storage.transfermanager;
 
-import com.google.cloud.WriteChannel;
+import com.google.cloud.storage.Blob;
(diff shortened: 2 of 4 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
```diff
@@ -64,6 +64,7 @@ public static Function<WriteChannel, Optional<BlobInfo>> maybeGetBlobInfoFunctio
     return (w) -> {
       BlobWriteChannel blobWriteChannel;
       if (w instanceof BlobWriteChannel) {
+
```
sydney-munro, line 67: accidental commit. will remove.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITTransferManagerTest.java
```diff
@@ -217,6 +232,28 @@ public void uploadNonexistentFile() throws Exception {
...
+      assertThat(uploadResults.get(0).getStatus()).isEqualTo(TransferStatus.SUCCESS);
+      UploadJob failedSecondUpload =
+          transferManager.uploadFiles(ImmutableList.of(tmpFile.getPath()), parallelUploadConfig);
+      List<UploadResult> failedResult = failedSecondUpload.getUploadResults();
+      assertThat(failedResult.get(0).getStatus()).isEqualTo(TransferStatus.FAILED_TO_FINISH);
+      assertThat(failedResult.get(0).getException()).isInstanceOf(StorageException.class);
+      assertThat(failedResult.get(0).getException().getMessage()).contains("Precondition Failed");
+    }
+  }
+
```
sydney-munro, line 256: add test for skipIfExists with generation in options and add test for skipIfExists with any other option

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/UploadCallable.java
```diff
@@ -64,12 +64,14 @@ private UploadResult uploadWithoutChunking() {
     long bytesCopied = -1L;
     try {
       Optional<BlobInfo> newBlob;
-      try (FileChannel r = FileChannel.open(sourceFile, StandardOpenOption.READ);
-          WriteChannel w = storage.writer(originalBlob, opts)) {
+      WriteChannel w = storage.writer(originalBlob, opts);
```
sydney-munro, line 67: replace with storage.createFrom

BenWhitehead (approved):

BenWhitehead (approved):

PR #2094, opened 2023-06-23
docs: Javadocs for TransferManager interface and ParallelUploadConfig

sydney-munro:
First commit for Javadocs. Will continue to add files as I write them but wanted to get initial feedback to ensure this is the direction we would like the javadocs to go and they are clear.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ParallelDownloadConfig.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ParallelDownloadConfig.java
@@ -53,7 +53,7 @@ public final class ParallelDownloadConfig {
   }
 
   /**
-   * A common prefix that is removed from downloaded object's name before written to the filesystem.
+   * A common prefix removed from an object's name before being written to the filesystem.
    */
   @BetaApi
   public @NonNull String getStripPrefix() {
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ParallelDownloadConfig.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ParallelDownloadConfig.java
@@ -72,7 +72,10 @@ public final class ParallelDownloadConfig {
     return bucketName;
   }
 
-  /** A list of common BlobSourceOptions that are used for each download request. Note this list of options will be applied to every single request. */
+  /**
+   * A list of common BlobSourceOptions that are used for each download request. Note this list of
+   * options will be applied to every single request.
+   */
   @BetaApi
   public @NonNull List<BlobSourceOption> getOptionsPerRequest() {
     return optionsPerRequest;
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ParallelUploadConfig.java
```diff
@@ -124,61 +125,82 @@ private static List<BlobWriteOption> applySkipIfExists(
     return writeOptsPerRequest;
   }
 
+  /**
+   * Builds an instance of ParallelUploadConfig.
+   *
+   * @see ParallelUploadConfig
```
frankyn, line 131: will this link to parallel upload config in javadoc?

frankyn (approved):
LGTM, just had one q which is from me not remember javadoc;

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ParallelUploadConfig.java
sydney-munro: yep, it will render as `See Also: ParallelUploadConfig` and the ParallelUploadConfig bit will link to those javadocs. Similar to using link just not inline.

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManager.java
```diff
@@ -18,19 +18,75 @@
...
+   * ParallelUploadConfig parallelUploadConfig =
+   *           ParallelUploadConfig.newBuilder()
+   *               .setBucketName(bucketName)
+   *               .build();
+   *
+   * UploadJob uploadedFiles = transferManager.uploadFiles(files, config);
+   *
+   * }</pre>
+   *
+   * @return an {@link UploadJob}
```
BenWhitehead, line 56: We should call out that this method will return quickly, and that awaiting results should be done on the UploadJob itself. Conversely, when we add the javadocs to UploadJob we should call out that getting results will block the invoking thread until the job is finished.

Same for DownloadJob

BenWhitehead (changes requested):
1. I'd like to see the `@see` cross links on the builder setters/config getters.
2. Adding a note to uploadFiles/downloadBlobs to note they do not block the invoking thread waiting for all the work to complete.
3. A few nits/alternative pharasings - if you feel these don't clarify anything feel free to ignore.

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/ParallelDownloadConfig.java
```diff
@@ -123,30 +136,57 @@ private Builder() {
...
+     */
     @BetaApi
     public Builder setBucketName(String bucketName) {
       this.bucketName = bucketName;
       return this;
     }
 
+    /**
+     * Sets the BlobSourceOptions that will be applied to each download request. Note these options
+     * will be applied to every single download request.
```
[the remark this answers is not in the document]
sydney-munro: The goal is to emphasize that every request gets these options. Since each was used in the first sentence it would be directly repeating the same thing. I can also remove the second sentence, it was more for emphasis than necessity.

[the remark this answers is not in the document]
BenWhitehead: Ah, okay. Either leaving as is, or removing emphasis sentence make sense to me. Up to you.

BenWhitehead (approved):

PR #2105, opened 2023-07-06
feat: Initial Preview of Transfer Manager

sydney-munro:
# Google Cloud Storage Transfer Manager Preview

Transfer Manager is currently in Preview, to denote this we have added the `@BetaApi` annotation to all related Transfer Manager APIs. Using Transfer Manager APIs at this time comes with the possibility of breaking changes in the future. The `@BetaApi` annotation will be removed to denote we are out of preview and the API is stable and will not break outside of a major release.

## Notable features

1. Parallel Uploads for a list of Files.
2. Parallel Downloads for a list of Blobs.
3. Parallel Downloads for large chunkable blobs. (i.e. Divide and Conquer Downloads)

## Not yet Implemented

1. Parallel Uploads for large chunkable files. (i.e. Divide and Conquer Uploads)


google-cloud-storage/src/main/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Google LLC
+ * Copyright 2023 Google LLC
```
BenWhitehead, line 2: We can move this file back into the test directory, it's no longer used by transfermanager prod code.

BenWhitehead (changes requested):

google-cloud-storage/src/test/java/com/google/cloud/storage/PackagePrivateMethodWorkarounds.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Google LLC
+ * Copyright 2023 Google LLC
```
BenWhitehead, line 2: Can we also put the copyright back to 2021 since we're undoing the move?

BenWhitehead (approved):

PR #2138, opened 2023-07-24
chore: introduce Throughput and ThroughputSink

BenWhitehead:
_Pre-work_

In some upcoming work we want to be able to keep a running window of throughput performance in order to provide improved throughput. This PR introduces a new utility class to model and compute throughput, and the concept of a ThroughputSink which values can be appended to.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
@@ -38,12 +38,12 @@ final class ThroughputMovingWindow {
   }
 
   void add(Instant now, Throughput value) {
-    houseKeeping(now);
+    removeExpiredEntries(now);
     values.add(new Entry(now, value));
   }
 
   Throughput avg(Instant now) {
-    houseKeeping(now);
+    removeExpiredEntries(now);
     return values.stream()
         .map(Entry::getValue)
         .reduce(
@@ -51,7 +51,7 @@ final class ThroughputMovingWindow {
             (tp1, tp2) -> Throughput.of(tp1.getNumBytes() + tp2.getNumBytes(), window));
   }
 
-  private void houseKeeping(Instant now) {
+  private void removeExpiredEntries(Instant now) {
     Instant newMin = now.minus(window);
     values.removeIf(e -> lteq(e.getAt(), newMin));
   }
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
```diff
@@ -0,0 +1,98 @@
...
+  Throughput avg(Instant now) {
+    houseKeeping(now);
+    return values.stream()
+        .map(Entry::getValue)
+        .reduce(
+            Throughput.zero(),
+            (tp1, tp2) -> Throughput.of(tp1.getNumBytes() + tp2.getNumBytes(), window));
+  }
+
+  private void houseKeeping(Instant now) {
```
sydney-munro, line 54: nit: houseKeeping is a little confusing, I had to read through the code a few times to figure out its purpose. maybe shiftWindow or something...

google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
```diff
@@ -0,0 +1,98 @@
...
+    this.window = window;
+    this.values = new PriorityQueue<>(Entry.COMP);
+  }
+
+  void add(Instant now, Throughput value) {
+    houseKeeping(now);
+    values.add(new Entry(now, value));
+  }
+
+  Throughput avg(Instant now) {
```
sydney-munro, line 45: nit: I'm sure we could expand if we needed it but it seems odd to me that we need to add a value and update the window in order to query the average of a window.

google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputSink.java
```diff
@@ -0,0 +1,251 @@
...
+      return delegate.isOpen();
+    }
+
+    @Override
+    public void close() throws IOException {
+      delegate.close();
+    }
+  }
+
+  final class TeeThroughputSink implements ThroughputSink {
```
sydney-munro, line 211: what does 'Tee' mean?

google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
BenWhitehead: We can come up with a different name. The idea of this method is to trim the entries we're tracking to those that fit within the time window relative to `now`.

The advantage of this approach is we don't need a background thread to do any cleanup or bounding, each operation done with the Moving Window implicitly does that.

google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputSink.java
BenWhitehead: Tee is common convention for taking an input and passing it to two different outputs.

The wikipedia article does a pretty good job giving an overview of its use in command line environments https://en.wikipedia.org/wiki/Tee_(command)

google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
BenWhitehead: Calling housekeeping will trim any expired values before we perform the aggregation. Similarly when adding a new value, trim any expired values before adding our new one.

sydney-munro: understood, so basically moving means it will only ever reflect now-window and not any other windowed period.

google-cloud-storage/src/main/java/com/google/cloud/storage/ThroughputMovingWindow.java
sydney-munro: makes sense, it would be beneficial to either add a comment explaining this or renaming of the method (or both).

BenWhitehead: renamed the method to `removeExpiredEntries`

sydney-munro (approved):
LGTM

PR #2139, opened 2023-07-24
feat: implement BufferToDiskThenUpload BlobWriteSessionConfig

BenWhitehead:
There are scenarios in which disk space is more plentiful than memory space. This new BlobWriteSessionConfig allows augmenting an instance of storage to prefer buffering to disk rather than keeping things in memory.

Once the file on disk is closed, the entire file will then be uploaded to GCS.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/RecoveryFile.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/RecoveryFile.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableSet;
+import java.io.IOException;
+import java.nio.channels.SeekableByteChannel;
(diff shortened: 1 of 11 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/RecoveryFile.java
```diff
@@ -0,0 +1,100 @@
...
+    this.throughputSink = throughputSink;
+    this.onCloseCallback = onCloseCallback;
+  }
+
+  public Path getPath() {
+    return path;
+  }
+
+  public Path touch() throws IOException {
+    return Files.createFile(path);
```
sydney-munro, line 55: shouldn't we be checking if the file exists before attempting to create it?

google-cloud-storage/src/main/java/com/google/cloud/storage/RecoveryFile.java
```diff
@@ -0,0 +1,100 @@
...
+        .add("throughputSink", throughputSink)
+        .add("onCloseCallback", onCloseCallback)
+        .toString();
+  }
+
+  Unsafe unsafe() {
+    return new Unsafe();
+  }
+
+  final class Unsafe {
```
sydney-munro, line 85: I assume this has to do with my previous comment on touch but I don't fully understand the purpose of this? it just wraps the exception?

google-cloud-storage/src/main/java/com/google/cloud/storage/RecoveryFile.java
BenWhitehead: That is done atomically by this helper method[1].

If we were to try and do it ourselves, we could not provide the same atomicity.

[1] https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createFile-java.nio.file.Path-java.nio.file.attribute.FileAttribute...-

google-cloud-storage/src/main/java/com/google/cloud/storage/RecoveryFile.java
BenWhitehead: This is a bit of a utility decorator to make using any Unsafe methods easier from Streams.

Because all of these IO methods declare `throws IOException` attempting to use them in Streams results in needing to handle the exception in every use.

sydney-munro (approved):

PR #2146, opened 2023-07-28
fix: optimize resumable uploads to allow sending bytes during finalization

BenWhitehead:
Add new methods to UnbufferedWritableByteChannel to allow writing and closing in a single call
* writeAndClose(ByteBuffer)
* writeAndClose(ByteBuffer[])
* writeAndClose(ByteBuffer[], int, int)

Update grpc and json implementation to leverage new methods and to write and finalize in the same call.

DefaultBufferedWritableByteChannel will use the new methods as appropriate.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultStorageRetryStrategy.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultStorageRetryStrategy.java
@@ -145,7 +145,7 @@ final class DefaultStorageRetryStrategy implements StorageRetryStrategy {
     public RetryResult beforeEval(Exception exception) {
       if (exception instanceof IllegalArgumentException) {
         IllegalArgumentException illegalArgumentException = (IllegalArgumentException) exception;
-        if (illegalArgumentException.getMessage().equals("no JSON input found")) {
+        if ("no JSON input found".equals(illegalArgumentException.getMessage())) {
           return RetryResult.RETRY;
         }
       }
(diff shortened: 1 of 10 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultStorageRetryStrategy.java
```diff
@@ -145,7 +145,7 @@ private static final class EmptyJsonParsingExceptionInterceptor implements BaseI
     public RetryResult beforeEval(Exception exception) {
       if (exception instanceof IllegalArgumentException) {
         IllegalArgumentException illegalArgumentException = (IllegalArgumentException) exception;
-        if (illegalArgumentException.getMessage().equals("no JSON input found")) {
+        if ("no JSON input found".equals(illegalArgumentException.getMessage())) {
```
BenWhitehead, line 148: Move the constant to the left side to avoid NullPointerException if there is ever an IllegalArgumentException which doesn't have a message.

google-cloud-storage/src/main/java/com/google/cloud/storage/UnbufferedWritableByteChannelSession.java
```diff
@@ -28,12 +28,26 @@ interface UnbufferedWritableByteChannelSession<ResultT>
   interface UnbufferedWritableByteChannel extends WritableByteChannel, GatheringByteChannel {
     @Override
     default int write(ByteBuffer src) throws IOException {
-      return Math.toIntExact(write(new ByteBuffer[] {src}));
+      return Math.toIntExact(write(new ByteBuffer[] {src}, 0, 1));
```
frankyn, line 31: Does `0, 1` represent there's only 1 buffer right now?

BenWhitehead: Correct. These values represent the index to start from in the provided array and how many indices to consume.

This is reducing a stack entry where before it would do `write(ByteBuffer) -> write(ByteBuffer[]) -> write(ByteBuffer[], int, int)`. While logically equivalent (and the JVM can probably inline the method to reduce impact anyway) doing this makes sure we don't add another stack frame.

frankyn (approved):

PR #2168, opened 2023-08-16
fix: update User-Agent handling for resumable uploads

BenWhitehead:
The apiary library manually injects ApplicationName into each request it constructs. For resumable uploads (start and PUTs) we are using the apiary clients http client directly. Update our registered interceptor to add the user-agent if it is null.

Fixes b/294437023

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -123,14 +123,15 @@ public class HttpStorageRpc implements StorageRpc {
     this.options = options;
 
     // Open Census initialization
+    String applicationName = options.getApplicationName();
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -183,6 +191,16 @@ public void intercept(HttpRequest request) throws IOException {
         }
         headers.set("x-goog-api-client", newValue);
         headers.set("x-goog-gcs-idempotency-token", invocationId);
+
+        String userAgent = headers.getUserAgent();
+        if ("benasdf".equals(userAgent)) {
```
sydney-munro, line 196: is this debugging left in?

BenWhitehead: yeah, cleaned up

sydney-munro (approved):
LGTM

PR #2182, opened 2023-08-25
fix: a resumable session without a Range header should be interpreted as 0 length

BenWhitehead:
According to https://cloud.google.com/storage/docs/performing-resumable-uploads#status-check a 308 response that does not contain a Range header should interpret as GCS having received no data.

Include x-goog-gcs-idempotency-token in Json Resumable upload debug context

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
@@ -65,8 +65,7 @@ enum JsonResumableSessionFailureScenario {
       BaseServiceException.UNKNOWN_CODE,
       "dataLoss",
       "Client side data loss detected. Bytes acked is more than client sent."),
-  SCENARIO_9(503, "backendNotConnected", "Ack less than bytes sent"),
-  QUERY_SCENARIO_1(503, "", "Missing Range header in response");
+  SCENARIO_9(503, "backendNotConnected", "Ack less than bytes sent");
 
   private static final String PREFIX_I = "\t|< ";
   private static final String PREFIX_O = "\t|> ";
@@ -79,6 +78,7 @@ enum JsonResumableSessionFailureScenario {
           .or(matches("Content-Type"))
           .or(matches("Range"))
           .or(startsWith("X-Goog-Stored-"))
+          .or(matches("X-Goog-GCS-Idempotency-Token"))
           .or(matches("X-GUploader-UploadID"));
 
   private static final Predicate<Map.Entry<String, ?>> includeHeader =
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionQueryTask.java
```diff
@@ -92,13 +92,14 @@ final class JsonResumableSessionQueryTask
         }
       } else if (JsonResumableSessionFailureScenario.isContinue(code)) {
         String range1 = response.getHeaders().getRange();
+        //
```
frankyn, line 95: Unneeded comment

google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionQueryTask.java
```diff
@@ -92,13 +92,14 @@ final class JsonResumableSessionQueryTask
...
         if (range1 != null) {
           ByteRangeSpec range = ByteRangeSpec.parse(range1);
           long endOffset = range.endOffset();
           return ResumableOperationResult.incremental(endOffset);
         } else {
-          throw JsonResumableSessionFailureScenario.QUERY_SCENARIO_1.toStorageException(
-              uploadId, response);
+          // https://cloud.google.com/storage/docs/performing-resumable-uploads#status-check
```
frankyn, line 101: Could you add a comment about this doc link stating the case?

google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionFailureScenario.java
```diff
@@ -79,6 +78,7 @@ enum JsonResumableSessionFailureScenario {
           .or(matches("Content-Type"))
           .or(matches("Range"))
           .or(startsWith("X-Goog-Stored-"))
+          .or(matches("X-Goog-GCS-Idempotency-Token"))
```
frankyn, line 81: What if the code collapsed to `startsWith("X-Goog-")`?

BenWhitehead: I wanted to reduce the scope to prevent an accidental inclusion of a header that might contain a sensitive value.

If it were relaxed to `x-goog-` that could pull in object metadata fields which could have sensitive values.

Since these error messages go into peoples logs, we should only include what we know is safe.

google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSessionQueryTask.java
BenWhitehead: done

frankyn (approved):

PR #2185, opened 2023-08-30
chore(benchmarking): Fix formatting for custom metric ingestion

sydney-munro:


storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/CloudMonitoringResult.java
```diff
@@ -147,6 +147,23 @@ public int hashCode() {
         throughput);
   }
 
+  public String generateCustomMetric() {
+    return String.format(
+        "throughput{library=%s,api=%s,op=%s,object_size=%d,chunksize=%d,workers=%d,crc32c_enabled=%b,md5_enabled=%b,bucket_name=%s,status=%s,app_buffer_size=%d}%.1f",
```
sydney-munro, line 152: I apologize, this is kinda ugly.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/CloudMonitoringResult.java
```diff
@@ -147,6 +147,23 @@ public int hashCode() {
         throughput);
   }
 
+  public String generateCustomMetric() {
```
BenWhitehead, line 150: I'd probably name this something like `formatAsCustomMetric` so it's clear this is a "taking this thing and turning it into another thing (which just happens to be a string)"

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/CloudMonitoringResult.java
BenWhitehead: This looks okay to me as long as all the strings going in here are guaranteed to be non-null.

Do you have a plan in mind for how to handle fields which aren't collected for a specific benchmark?

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/CloudMonitoringResult.java
```diff
@@ -147,6 +147,23 @@ public int hashCode() {
...
+        api,
+        op,
+        object_size,
+        chunksize,
+        workers,
+        crc32c_enabled,
+        md5_enabled,
+        bucket_name,
+        status,
+        app_buffer_size,
```
BenWhitehead, line 163: Now that we're actually rendering the output string for the metrics ingestion, can we change this field names to camel case rather than snake case?

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/CloudMonitoringResult.java
sydney-munro: I did think about this. we could have some overloaded versions of this where we take in the metric name and value and have an appropriate format string for those types.

ie ..
`public String generateCustomMetric(String metric, <some other type> <some other val>)`

sydney-munro: I think when we are using those values I am fine with including them and having them come out as their defaults for irrelevant workloads.

BenWhitehead: Sounds okay to me. Might want to sync with the other languages and follow their lead so we aren't expanding the cardinality on the metrics slowing down queries.

BenWhitehead (approved):

PR #2208, opened 2023-09-14
chore(benchmarking): Putting temp files in try with resource for cleanup

sydney-munro:


storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/W1R3.java
```diff
@@ -58,37 +58,39 @@ final class W1R3 implements Callable<String> {
   @Override
   public String call() throws Exception {
     // Create the file to be uploaded and fill it with data
-    TmpFile file = DataGenerator.base64Characters().tempFile(tempDirectory, objectSize);
-    BlobInfo blob = BlobInfo.newBuilder(bucketName, file.toString()).build();
+    try(TmpFile file = DataGenerator.base64Characters().tempFile(tempDirectory, objectSize)) {
```
sydney-munro, line 61: The only actual changes here are the additions of the `try` here

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/W1R3.java
```diff
@@ -58,37 +58,39 @@ final class W1R3 implements Callable<String> {
...
+      Duration elapsedTimeUpload = Duration.between(startTime, endTime);
       printWriter.println(
           generateCloudMonitoringResult(
+              "WRITE",
+              StorageSharedBenchmarkingUtils.calculateThroughput(
+                  created.getSize().longValue(), elapsedTimeUpload),
+              created)
+              .formatAsCustomMetric());
+      for (int i = 0; i <= StorageSharedBenchmarkingUtils.DEFAULT_NUMBER_OF_READS; i++) {
+        try (TmpFile dest = TmpFile.of(tempDirectory, "prefix", "bin")) {
```
sydney-munro, line 78: and here

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/W1R3.java
```diff
@@ -58,37 +58,39 @@ final class W1R3 implements Callable<String> {
   @Override
   public String call() throws Exception {
     // Create the file to be uploaded and fill it with data
-    TmpFile file = DataGenerator.base64Characters().tempFile(tempDirectory, objectSize);
-    BlobInfo blob = BlobInfo.newBuilder(bucketName, file.toString()).build();
+    try (TmpFile file = DataGenerator.base64Characters().tempFile(tempDirectory, objectSize)) {
```
BenWhitehead, line 61: you don't want to use the base64 character set here for generate. It creates highly compressible and dedupeable data which Gcs sometimes will optimize for leading to skewed data.

You can either use the existing `rand` or port the `devUrandom` implementation from the existing benchmarks into this repos version.

it's fine to do this in a follow up pr.

BenWhitehead (approved):
Primary change looks okay, but there is a follow up for the data generation that will need to be addressed.

PR #2222, opened 2023-09-22
fix: consider Storage#delete(BlobId) idempotent when id has generation

BenWhitehead:
Update logic to consider an object delete request to be idempotent if the objects generation is specified in the BlobId.


danielduhh (approved):
Looks solid. Curious if the grpc testbench work would have caught this?

PR #2237, opened 2023-09-28
chore(benchmarking): Add a profile to build the SSB cli

sydney-munro:
Created a profile for packaging the CLI without including it in our releases.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -208,5 +208,11 @@
         <module>samples</module>
       </modules>
     </profile>
+    <profile>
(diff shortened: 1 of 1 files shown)


BenWhitehead (approved):
Might want to add an additional github action to validate the project keeps compiling when this profile is enabled.

you can probably copy https://github.com/googleapis/java-storage/blob/main/.github/workflows/samples.yaml as a starting point, and modify that to run `mvn -Pinclude-storage-shared-benchmarking clean verify`

PR #2242, opened 2023-10-02
docs: reorder columns in table of BlobWriteSessionConfigs.java

BenWhitehead:
The considerations column being last is cut off in the ultimate rendered javadocs. Move it left before retries and apis used.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteSessionConfigs.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteSessionConfigs.java
@@ -41,10 +41,10 @@ import java.util.Collection;
  *     <th>Strategy</th>
  *     <th>Factory Method(s)</th>
  *     <th>Description</th>
- *     <th>Retry Support</th>
- *     <th>Transports Supported</th>
- *     <th>Cloud Storage API used</th>
+ *     <th>Transport(s) Supported</th>
  *     <th>Considerations</th>
+ *     <th>Retry Support</th>
+ *     <th>Cloud Storage API used</th>
  *   </tr>
  *   <tr>
  *     <td>Default (Chunk based upload)</td>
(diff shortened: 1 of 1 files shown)


sydney-munro (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -534,36 +534,42 @@ public boolean delete(String bucket, String blob, BlobSourceOption... options) {
...
+    }
   }
 
   @Override
   public boolean delete(BlobId blob) {
     return delete(blob, new BlobSourceOption[0]);
   }
 
+  @Override
+  public Void internalObjectDelete(BlobId id, Opts<ObjectSourceOpt> opts) {
```
frankyn, line 552: why is this public instead of private? 

Not sure if this should be this in PR given it's mainly docs updates.

google-cloud-storage/src/test/java/com/google/cloud/storage/ParallelCompositeUploadWritableByteChannelTest.java
```diff
@@ -839,9 +839,13 @@ public BlobInfo compose(ComposeRequest composeRequest) {
     }
 
     @Override
-    public boolean delete(BlobId id) {
+    public Void internalObjectDelete(BlobId id, Opts<ObjectSourceOpt> opts) {
```
frankyn, line 842: why is `v` capitalized?

frankyn (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
BenWhitehead: It is a docs only update, because this was branch off another branch the commit had a distinct sha in this PR after merging the earlier PR to main.

I've rebased, and there is only javadoc update again.

google-cloud-storage/src/test/java/com/google/cloud/storage/ParallelCompositeUploadWritableByteChannelTest.java
BenWhitehead: Same thing as the other comment.

frankyn (approved):

PR #2246, opened 2023-10-04
test: ensure Storage instances are closed in ITStorageLifecycleTest

BenWhitehead:
Looking at CI logs I noticed grpc was logging a leaked channel pool. The storage instance wasn't being closed in these tests.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/ITStorageLifecycleTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/ITStorageLifecycleTest.java
@@ -66,12 +66,12 @@ public final class ITStorageLifecycleTest {
     service1.close();
 
     // expect a new instance to be returned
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/ITStorageLifecycleTest.java
```diff
@@ -66,12 +66,12 @@ public void grpc() throws Exception {
     service1.close();
```
frankyn, line 66: When is service2 closed?

BenWhitehead: `service1` and `service2` are the same instance, so when `service1` is closed so it `service2`

frankyn (approved):

PR #2277, opened 2023-10-24
feat: add object retention feature

JesseLovelace:
Adds the object retention feature and tests


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -168,6 +172,116 @@ public final boolean equals(Object o) {
     }
   }
 
+  /**
+   * Defines a blob's Retention policy. Can only be used on objects in a retention-enabled bucket.
+   */
+  public static class Retention implements Serializable {
```
BenWhitehead, line 178:
```suggestion
  public static final class Retention implements Serializable {
```

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -168,6 +172,116 @@ public final boolean equals(Object o) {
...
+
+    public static final class Mode extends StringEnumValue {
```
BenWhitehead, line 180: nit: move nested class after nested builder.

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobInfo.java
```diff
@@ -168,6 +172,116 @@ public final boolean equals(Object o) {
...
+    public String toString() {
+      return MoreObjects.toStringHelper(this)
+          .add("mode", mode)
+          .add("retainUntilTime", retainUntilTime)
+          .toString();
+    }
+
+    public static Builder newBuilder() {
+      return new Builder();
+    }
```
BenWhitehead, line 253: I can't remember, do we usually do `toBuilder()` for the nested model classes or only the top level ones?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -427,6 +432,90 @@ public Autoclass build() {
     }
   }
 
+  public static final class ObjectRetention implements Serializable {
+    public static final class Mode extends StringEnumValue {
```
BenWhitehead, line 436: nit: move this class below the nested builder

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -1536,6 +1625,8 @@ public Builder setRetentionPeriodDuration(Duration retentionPeriod) {
 
     public abstract Builder setCustomPlacementConfig(CustomPlacementConfig customPlacementConfig);
 
+    abstract Builder setObjectRetention(ObjectRetention objectRetention);
```
BenWhitehead, line 1628: This is intentionally package-private?

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -158,7 +158,10 @@ enum BucketField implements FieldSelector, NamedField {
     @TransportCompatibility({Transport.HTTP, Transport.GRPC})
     CUSTOM_PLACEMENT_CONFIG("customPlacementConfig", "custom_placement_config"),
     @TransportCompatibility({Transport.HTTP, Transport.GRPC})
-    AUTOCLASS("autoclass");
+    AUTOCLASS("autoclass"),
+
```
BenWhitehead, line 162: nit: remove blank line to keep consistency with other groupings.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -256,7 +259,10 @@ enum BlobField implements FieldSelector, NamedField {
     @TransportCompatibility({Transport.HTTP, Transport.GRPC})
     TIME_STORAGE_CLASS_UPDATED("timeStorageClassUpdated", "update_storage_class_time"),
     @TransportCompatibility({Transport.HTTP, Transport.GRPC})
-    CUSTOMER_ENCRYPTION("customerEncryption", "customer_encryption");
+    CUSTOMER_ENCRYPTION("customerEncryption", "customer_encryption"),
+
```
BenWhitehead, line 263: nit: remove blank line to keep consistency with other groupings.

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -1399,6 +1407,14 @@ public Mapper<UpdateBucketRequest.Builder> updateBucket() {
     }
   }
 
+  static final class EnableObjectRetention extends RpcOptVal<Boolean> implements BucketTargetOpt {
+    private static final long serialVersionUID = -2581147719605551578L;
+
+    private EnableObjectRetention(boolean val) {
+      super(StorageRpc.Option.ENABLE_OBJECT_RETENTION, val);
+    }
```
BenWhitehead, line 1415: since this doesn't support grpc yet, can you define an erroring mapper for the updateBucket() method similar to what was done with match glob in https://github.com/googleapis/java-storage/pull/1965/files#diff-4162f80a70285d5bd410d76a86cf59b5fa513dc4ad8e093115926eddda94f68bR1085-R1089

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -1623,6 +1639,16 @@ public String toString() {
...
   }
 
+  static final class OverrideUnlockedRetention extends RpcOptVal<Boolean>
+      implements ObjectTargetOpt {
+
+    private static final long serialVersionUID = -7764590745622588287L;
+
+    private OverrideUnlockedRetention(boolean val) {
+      super(StorageRpc.Option.OVERRIDE_UNLOCKED_RETENTION, val);
+    }
```
BenWhitehead, line 1649: same here, erroring mapper for updateObject() method similar to that of matchGlob

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBucketTest.java
```diff
@@ -417,6 +419,70 @@ public void testCreateBucketWithAutoclass() {
...
+  @CrossRun.Exclude(transports = Transport.GRPC)
+  public void testObjectRetention() {
+    String bucketName = generator.randomBucketName();
+
+    // Create a bucket with object retention enabled
+    storage.create(
+        BucketInfo.newBuilder(bucketName).build(), BucketTargetOption.enableObjectRetention(true));
+
+    try {
+      Bucket remoteBucket = storage.get(bucketName);
```
BenWhitehead, line 432: If interested, you could use the TemporaryBucket helper to create and automatically cleanup the bucket. Example of this https://github.com/googleapis/java-storage/blob/cf944aeb9f8be34c7e860a57515971b626267cb6/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBucketTest.java#L123-L124

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryConversions.java
```diff
@@ -238,6 +246,11 @@ private StorageObject blobInfoEncode(BlobInfo from) {
         from.getRetentionExpirationTimeOffsetDateTime(),
         dateTimeCodec::encode,
         to::setRetentionExpirationTime);
+    if (from.getRetention() == null) {
+      to.setRetention(Data.nullOf(StorageObject.Retention.class));
```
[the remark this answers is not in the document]
JesseLovelace: It doesn't work, due to the modified fields not being passed along to the final patch call. I talked to ben and we agreed to just do it this way, there's precedent for it in the logging encoding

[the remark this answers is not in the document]
frankyn: Thanks for clarifying, is this value mutable once set?

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
BenWhitehead: Sounds good, thanks.

google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryConversions.java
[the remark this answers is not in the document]
JesseLovelace: Only certain cases work (i.e. going from Unlocked to Locked, reducing the retainUntilTime on an unlocked policy, and setting to null), and you have to pass in `overrideUnlockedRetention` for any of those cases to work, but yes

JesseLovelace:
The current failure is just because the test project isn't allowlisted yet, it should be soon

BenWhitehead (approved):

PR #2297, opened 2023-11-07
chore: Add in TM entry for x-goog-api-client

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
@@ -60,6 +60,7 @@ import java.io.IOException;
 import java.io.Serializable;
 import java.net.URI;
 import java.time.Clock;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


BenWhitehead (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -207,13 +208,23 @@ private Tuple<StorageSettings, Opts<UserProject>> resolveSettingsAndOpts() throw
...
+                ste -> ste.getClassName().startsWith("com.google.cloud.storage.transfermanager"));
+
     HeaderProvider internalHeaderProvider =
         StorageSettings.defaultApiClientHeaderProviderBuilder()
             .setClientLibToken(ServiceOptions.getGoogApiClientLibName(), getLibraryVersion())
             .build();
+    if (isTm) {
+      internalHeaderProvider =
+          XGoogApiClientHeaderProvider.of(
+              internalHeaderProvider, ImmutableList.of("gccl-gcs-cmd/tm"));
```
danielduhh, line 223: There's 4 different TM header options:
```
tm.upload_many
tm.download_many
tm.upload_sharded
tm.download_sharded
```
Are you planning to support this? cc: @andrewsg

BenWhitehead: The lifecycle of this header in java is only as granular as the client instance. We do not have the means of doing per invocation selection. This limitation was mentioned back during the design phase, and is part of why there is `tm.` to allow prefix course selection while allowing fine grained selection for those languages where it is possible.

PR #2299, opened 2023-11-09
test: add support to ITRetryConformanceTest suite to allow running gRPC transport tests

BenWhitehead:
* Current test case resolver still only resolves HTTP cases
* RetryTestFixture is updated to turn a `501 Not Implemented` into an AssumptionViolationException signaling the test is skipped/ignored
* RetryTestFixture is updated to call close on Storage instances -- needed so we don't leak channel builders with grpc
* RetryTestFixture nonTestStorage will always be HTTP, only testStorage will be built with the selected transport
* TestRetryConformance generated names now include either `_h` or `_g` depending on transport -- this removes collisions of names between the two transports
* TestRetryConformance will now generate test names of the following format: `TestRetryConformance/{lower_transport}-{scenario_id}-[{instructions}]-{method_name}-{mapping_id}`
* TestBench.RetryTestResource has a new property `transport`

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
@@ -30,6 +30,7 @@ import com.google.cloud.conformance.storage.v1.RetryTest;
 import com.google.cloud.conformance.storage.v1.RetryTests;
 import com.google.cloud.storage.CIUtils;
 import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.TransportCompatibility.Transport;
 import com.google.cloud.storage.conformance.retry.Functions.CtxFunction;
 import com.google.cloud.storage.conformance.retry.ITRetryConformanceTest.RetryConformanceParameterProvider;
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/conformance/retry/ITRetryConformanceTest.java
```diff
@@ -276,69 +283,76 @@ private List<RetryTestCase> generateTestCases(
...
-                      testCase.getPreconditionProvided(),
-                      false);
-              if (testAllowFilter.test(key, testRetryConformance)) {
-                testCases.add(
-                    new RetryTestCase(testRetryConformance, RpcMethodMapping.notImplemented(key)));
-              }
-            } else {
-              for (RpcMethodMapping mapping : mappings) {
+      Transport[] values = Transport.values();
+      for (Transport transport : values) {
```
BenWhitehead, line 287: The change here is the introduction of the transport outer loop.

The existing loops had their indentation increased with the new outer loop. Github is doing a not so great job of illustrating that here.

frankyn (approved):

cojenco (approved):
🎉🎉

PR #2342, opened 2023-12-28
fix: Add an exception to zero byte uploads on CreateFrom

JesseLovelace:
[This PR](https://github.com/googleapis/java-storage/commit/4c2f44e28a1ff19ffb2a02e3cefc062a1dd98fdc) introduced a bug in using createFrom to upload a zero byte object. I can't figure out where exactly the bug is, but something is causing the content range to be 0-1 instead of 0-0 in this case. It ends up getting rejected because the range is bigger than the content size, which is still zero.

We can fix the bug later when we identify the root cause, but for now I'm just reverting to the old way of doing it when we detect that we're uploading a zero byte file.

I also added a test for this case

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
@@ -230,6 +230,10 @@ final class StorageImpl extends BaseService<StorageOptions> implements Storage {
     if (Files.isDirectory(path)) {
       throw new StorageException(0, path + " is a directory");
     }
+    long size = Files.size(path);
+    if (size == 0L) {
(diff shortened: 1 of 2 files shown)


BenWhitehead:
Fix looks well to me. In fact skipping creating a resumable session at all is a good call when the object is empty.

My guess is the error is actually a Content-Range header value of `0--1/0` (zero to -1 of zero). Which is calculated based on the 0 being passed in for size here https://github.com/googleapis/java-storage/blob/b2fc2ae184f76acdcae0d97ac7fe7d0b99449705/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java#L255-L256

Thanks for the fix @JesseLovelace

frankyn (approved):

PR #2343, opened 2024-01-02
feat: Add Bidi write feature

JesseLovelace:
This adds a new BlobWriteSessionConfig which uses the Bidi streaming feature. It's largely the same as our normal resumable upload flow, with two key differences:

-Instead of closing and re-opening a stream on each flush, we keep a stream open for the life of the upload
-On each flush, it sends a signal to GCS with the "flush" and "state_lookup" flags (which are exclusive to bidi writes) set to true

I still need to run a formal profiling test, but basic profiling in IntelliJ is consistent with the 10-15% increase we're expecting


google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
```diff
@@ -121,6 +146,27 @@ private static WriteObjectRequest possiblyPairDownRequest(
...
+    if (!firstMessageOfStream) {
+      b.clearUploadId();
+    }
+
+    if (message.getWriteOffset() > 0) {
+      b.clearWriteObjectSpec();
+    }
+
+    if (message.getWriteOffset() > 0 && !message.getFinishWrite()) {
+      b.clearObjectChecksums();
```
frankyn, line 165: Does this remove checksums for every message after initial / ignoring last? I would expect checksums to be supplied for intermediate messages as well.

google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
```diff
@@ -121,6 +146,27 @@ private static WriteObjectRequest possiblyPairDownRequest(
     return b.build();
   }
 
+  private static BidiWriteObjectRequest possiblyPairDownBidiRequest(
```
frankyn, line 149: Please add unit tests for this helper.

google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
```diff
@@ -199,6 +264,82 @@ public void close(@Nullable WriteObjectRequest req) {
...
+            boolean first = true;
+            for (BidiWriteObjectRequest message : segments) {
+              message = possiblyPairDownBidiRequest(message, first);
+
+              stream.onNext(message);
+              first = false;
+            }
+            BidiWriteObjectRequest message =
+                BidiWriteObjectRequest.newBuilder().setFlush(true).setStateLookup(true).build();
+            stream.onNext(message);
```
frankyn, line 313: IIUC the main difference between bidi and non-bidi is that stream.onComplete() is not called until the Observer observes onCompleted?

JesseLovelace: Correct, current flow calls onComplete on each flush. Bidi will only call onComplete once, at the end of the upload

google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
frankyn: My original intention was to introduce `@VisibleForTesting` for the specific comment thread around removing checksum in intermediate messages which is a misunderstanding on my part after talking with Jesse.

image

States that checksum can only be provided in first or last message sent to API so this comment and the one below are obsolete.

google-cloud-storage/src/main/java/com/google/cloud/storage/WriteFlushStrategy.java
frankyn: Ignore my understanding was incomplete.

frankyn (approved):
Remaining comments are nits; PR LGTM, thanks for the offline discussion @JesseLovelace

PR #2345, opened 2024-01-04
feat: Add ability to create a PCU Prefix at the object level

sydney-munro:
This feature enabled the user to create a prefix that will be unique per parallel composite upload. If the user specifies a prefix the prefix will be appended to the ultimate object name which will be appended to the part name to create a staging area for objects per PCU.

If the user does not specify a prefix then the parts will simply be created at the level of the ultimate object name.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ParallelCompositeUploadBlobWriteSessionConfig.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ParallelCompositeUploadBlobWriteSessionConfig.java
@@ -521,6 +521,11 @@ public final class ParallelCompositeUploadBlobWriteSessionConfig extends BlobWri
       return new WithPrefix(rand, prefixPattern);
     }
 
+    @BetaApi
+    public static PartNamingStrategy objectLevelPrefix() {
+      return objectLevelPrefix("");
+    }
+
     /**
      * Strategy in which an explicit stable prefix with object name included is present on each part
      * and intermediary compose object.
@@ -583,7 +588,9 @@ public final class ParallelCompositeUploadBlobWriteSessionConfig extends BlobWri
 
       private WithObjectLevelPrefix(SecureRandom rand, String prefix) {
         super(rand);
-        this.prefix = prefix;
+        // If no prefix is specified we will create the part files under the same directory as the
+        // ultimate object.
+        this.prefix = prefix.isEmpty() ? prefix :  prefix + "/";
       }
 
       @Override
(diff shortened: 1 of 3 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/ParallelCompositeUploadBlobWriteSessionConfig.java
```diff
@@ -534,8 +566,40 @@ private WithPrefix(SecureRandom rand, String prefix) {
       }
 
       @Override
-      protected String fmtFields(String randomKey, String nameDigest, String partRange) {
+      protected String fmtFields(String randomKey, String ultimateObjectName, String partRange) {
```
frankyn, line 569: ~Shouldn't this implementation of `fmtFields()` be left alone?~

ah before fmtFields() would receive a digest instead of the ultimateObject name.

google-cloud-storage/src/main/java/com/google/cloud/storage/ParallelCompositeUploadBlobWriteSessionConfig.java
```diff
@@ -523,6 +520,41 @@ public static PartNamingStrategy prefix(String prefixPattern) {
...
+     * </code></pre>
+     *
+     * <p>{@code {objectInfoDigest}} will be fixed for an individual {@link BlobWriteSession}.
+     *
+     * <p><b><i>NOTE:</i></b>The way in which both {@code randomKeyDigest} and {@code
+     * objectInfoDigest} are generated is undefined and subject to change at any time.
+     *
+     * <p>Care must be taken when choosing to specify a stable prefix as this can create hotspots in
+     * the keyspace for object names. See <a
+     * href="https://cloud.google.com/storage/docs/request-rate#naming-convention">Object Naming
```
frankyn, line 545: I don't think this objectLevelPrefix is needed as a prefix to object name. Reading through https://cloud.google.com/storage/docs/request-rate#randomness_after_a_common_prefix_is_effective_under_the_prefix; the default empty string impl would be enough along with the randomization that it uses.

frankyn (commented):
Looks good, just single clarifying point.

google-cloud-storage/src/main/java/com/google/cloud/storage/ParallelCompositeUploadBlobWriteSessionConfig.java
```diff
@@ -523,6 +520,35 @@ public static PartNamingStrategy prefix(String prefixPattern) {
...
     }
 
+    /**
+     * Strategy in which the end object name is the prefix included and is present on each part and
+     * intermediary compose object.
+     *
+     * <p>General format is
+     *
+     * <pre><code>
+     *   {objectName}-parts/{randomKeyDigest};{objectInfoDigest};{partIndex}.part
```
frankyn, line 530: I really think this should be updated to not use a directory structure from possible unknowns.

PR #2353, opened 2024-01-09
test: add IT for HNS bucket

frankyn:
Added a short IT for HNS Folder creation to test out new storage control client.

### DIFF
--- a/google-cloud-storage-control/src/test/java/com/google/storage/control/v2/ITFoldersTest.java
+++ b/google-cloud-storage-control/src/test/java/com/google/storage/control/v2/ITFoldersTest.java
@@ -24,6 +24,7 @@ import com.google.api.services.storage.model.Bucket;
 import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.storage.StorageOptions;
 import java.util.UUID;
+import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


google-cloud-storage-control/src/test/java/com/google/storage/control/v2/ITFoldersTest.java
```diff
@@ -0,0 +1,65 @@
...
+      Storage.Buckets.Insert insert = storage.buckets().insert(option.getProjectId(), bucket);
+      insert.execute();
+      StorageControlClient storageControlClient = StorageControlClient.create();
+      Folder folder =
+          storageControlClient.createFolder(
+              CreateFolderRequest.newBuilder()
+                  .setParent(BucketName.format("_", bucketName))
+                  .setFolderId("foldername/")
+                  .build());
+    } finally {
```
sydney-munro, line 61: are we just making sure we are not throwing an exception? no checks?

frankyn: yea, that was my intention but i'll add an assert. If for example a folder creation is attempted without parent folder existing `createFolder` will throw an exception.

sydney-munro (approved):

PR #2359, opened 2024-01-16
chore(benchmarks): Add in optional warmup to benchmarks

sydney-munro:


BrennaEpp (commented):
Some initial comments, though I know this is still in draft so I know you may be aware of these already.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/W1R3.java
```diff
@@ -67,33 +70,35 @@ public String call() throws Exception {
...
         }
       }
       StorageSharedBenchmarkingUtils.cleanupObject(storage, created);
     }
     return "OK";
   }
 
+  private void printResult(String op, Blob created, Duration duration) {
+    if (isWarmup) {
+      printWriter.println(op + " Warming...");
```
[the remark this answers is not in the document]
sydney-munro: good to know! did not realize this would impact SSB.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
```diff
@@ -83,6 +80,9 @@ public final class StorageSharedBenchmarkingCli implements Runnable {
       description = "Specify the path where the temporary directory should be located")
   String tempDirLocation;
 
+  @Option(names = "-warmup", description = "Warmup in seconds", defaultValue = "0")
```
tritone, line 83: Maybe be a little more specific about what this means.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
```diff
@@ -108,35 +108,50 @@ private void runWorkload1() {
...
     Range objectSizeRange = Range.of(objectSize);
     for (int i = 0; i < samples; i++) {
       int objectSize = getRandomInt(objectSizeRange.min, objectSizeRange.max);
       PrintWriter pw = new PrintWriter(System.out, true);
-      workloadRuns.add(
-          convert(
+      long startTime = System.currentTimeMillis();
+      long endTime = startTime + (warmup * 1000);
+      // Run Warmup
+      while (System.currentTimeMillis() < endTime) {
```
tritone, line 144: It seems perhaps a little confusing to have this as just a few lines within W1R3 rather than as a separate method. Not a java expert here but maybe it's better to factor this out for ease of readability and/or sharing with other workloads?

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/W1R3.java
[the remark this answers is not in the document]
tritone: Yeah, better to keep any extra output from the warmup out of stdout.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
```diff
@@ -108,35 +108,50 @@ private void runWorkload1() {
...
+        convert(
+                executorService.submit(
+                    new W1R3(storageClient, workers, api, pw, objectSize, tempDir, bucket, true)))
```
tritone, line 147: In Go we did 4 threads/workers for the warmup regardless of the number of threads in the main workload. Could this be achieved for Java as well? If I'm reading this correctly looks like for a single-threaded W1R3 the warmup would also be single threaded, which is not what we want.

tritone (commented):
Couple comments, overall looks good to me. Thanks for your work getting this going.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
```diff
@@ -80,16 +80,28 @@ public final class StorageSharedBenchmarkingCli implements Runnable {
       description = "Specify the path where the temporary directory should be located")
   String tempDirLocation;
 
-  @Option(names = "-warmup", description = "Warmup in seconds", defaultValue = "0")
+  @Option(
+      names = "-warmup",
+      description = "Number of seconds a W1R3 warmup will run on all available processors",
```
tritone, line 85: Let's keep the number of processors consistent between this and Go. We could either
1. Hard code 4 for both
2. Make it an option to specify the number
3. Change Go to use numCPU as well/

@BrennaEpp what do you prefer?

tritone (commented):
Overall LGTM aside from the open question about number of workers. You might also want to get a review from @frankyn on Java.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
BrennaEpp: So Go does use numCPU for warmups now, this would be consistent with that. Option 2 would be fine as well as we could make it consistent. I think numCPU is fine for warmups.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
```diff
@@ -108,35 +120,68 @@ private void runWorkload1() {
...
+      ListeningExecutorService executorService =
+          MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(workers));
+      for (int i = 0; i < samples; i++) {
+        if (warmup > 0) {
+          runWarmup(storageClient);
+        }
+        runW1R3(storageClient, false, executorService);
+      }
+    } catch (Exception e) {
+      System.err.println("Failed to run workload 4: " + e.getMessage());
```
frankyn, line 152: IIUC, you'll want to output this error as a `FAIL` status instead which means making status an option in: https://github.com/googleapis/java-storage/blob/4c7ab68396892766f11f0ae22e10887cb6878e95/storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/W1R3.java#L111

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
```diff
@@ -108,35 +120,68 @@ private void runWorkload1() {
     StorageOptions retryStorageOptions =
         StorageOptions.newBuilder().setProjectId(project).setRetrySettings(retrySettings).build();
     Storage storageClient = retryStorageOptions.getService();
-    runW1R3(storageClient);
+    try {
+      ListeningExecutorService executorService =
```
frankyn, line 124: IIUC, this code block was factored out of runW1R3; and to only sample once but created a copy for workload4 as well; Can you factor in sampling as an option to W1R3() and only print a sample if not a warm up instead?

frankyn (changes requested):
@sydney-munro, from our offline conversation, this PR is not yet ready, so some of my comment may not be relevant after the change is verified to work and is cleaned up.

I'll re-review once ready, but overall LGTM.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
tritone: Ahh okay, yeah in that case we are fine.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
sydney-munro: can you rephrase this ask?

frankyn: refactor runW1R3() to accept "isSample" (bool) and "samples" (int) to reduce duplication

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
sydney-munro: I added cloudmonitoring results for failures in the callable but left this in just for anything else (mainly so the IDE is happy)

frankyn (approved):

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
```diff
@@ -108,35 +120,66 @@ private void runWorkload1() {
...
-        tempDirLocation != null
-            ? Paths.get(tempDirLocation)
-            : Paths.get(System.getProperty("java.io.tmpdir"));
+  private void runW1R3(Storage storageClient) throws ExecutionException, InterruptedException {
     ListeningExecutorService executorService =
         MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(workers));
-    List<ApiFuture<String>> workloadRuns = new ArrayList<>();
-    Range objectSizeRange = Range.of(objectSize);
     for (int i = 0; i < samples; i++) {
+      runWarmup(storageClient);
```
BrennaEpp, line 146: This is interesting - so if you request 5 samples, it will run the warmup for every sample?

For SSB I guess it doesn't matter, because we only request one sample (unless it were server mode, which would require other changes too and we're not sure is needed yet for Java). But in general, I would expect the application only need to warm up once in the beginning, unless there are some specifics of the Java implementation I'm not aware of?

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
```diff
@@ -108,35 +120,66 @@ private void runWorkload1() {
...
+      Range objectSizeRange = Range.of(objectSize);
       int objectSize = getRandomInt(objectSizeRange.min, objectSizeRange.max);
-      PrintWriter pw = new PrintWriter(System.out, true);
-      workloadRuns.add(
-          convert(
+      convert(
               executorService.submit(
-                  new W1R3(storageClient, workers, api, pw, objectSize, tempDir, bucket))));
+                  new W1R3(
+                      storageClient, workers, api, printWriter, objectSize, tempDir, bucket, true)))
```
BrennaEpp, line 180: Trying to understand the Java here - `workers` in this line is passed in solely for output, right?

BrennaEpp (commented):
Thanks Sydney, a couple questions.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
frankyn: I missed this; good catch. This line needs to be moved out of the loop;

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
sydney-munro: correct.

storage-shared-benchmarking/src/main/java/com/google/cloud/storage/benchmarking/StorageSharedBenchmarkingCli.java
BrennaEpp: Hm, it just occurred to me that on a single threaded benchmark, the connections for the later data points may be in fact less warmed than the first ones, since the warmups are done with multiple threads. But, that applies in Go, not sure for Java. The effect may also be negligible within a certain time frame. 

For consistency and because we aren't sure the effect of the warmups yet in Java, I'd say move this line out of the loop. It's easier to test with the line outside of the loop, since it's not multiplying the warmup period by the number of the sample.

For SSB workload 1, yes, every data point needs to be warmed but that's because every data point runs on a fresh client. In this case the client is used across samples so it's already warmed up. Outside of the single (or perhaps low) threaded benchmarks, spending more time warming is not necessary.

PR #2362, opened 2024-01-19
feat: Adds TPC support

JesseLovelace:
This PR adds an intercept when `UniverseDomain` is set in `StorageOptions`, which overrides the host in StorageOptions. It also changes signed URLs to respect custom hosts, including a specified universe domain. 

There are no tests for this because Kokoro is not able to access any TPC universes yet. I tested all these changes locally, and they're working against the staging universe.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
@@ -323,6 +323,11 @@ public class HttpStorageOptions extends StorageOptions {
     public ServiceRpc create(StorageOptions options) {
       if (options instanceof HttpStorageOptions) {
         HttpStorageOptions httpStorageOptions = (HttpStorageOptions) options;
+        // todo: In the future, this step will be done automatically, and the getResolvedApiaryHost helper method will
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
```diff
@@ -323,6 +323,16 @@ public HttpStorageRpcFactory() {}
     public ServiceRpc create(StorageOptions options) {
       if (options instanceof HttpStorageOptions) {
         HttpStorageOptions httpStorageOptions = (HttpStorageOptions) options;
+        // todo: In the future, this step will be done automatically, and the getResolvedApiaryHost
```
frankyn, line 326: Is there a tracking bug to know when this will be addressed?

frankyn (commented):
LGTM, just comment on how to track when the code blocks and be remove / updated.

google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
JesseLovelace: Added!

frankyn (approved):

PR #2385, opened 2024-01-29
chore: testing google-http-client 1.44.1 [no need to review]

suztomo:
google-http-client 1.44.1 stopped declaring commons-logging. Would that work?

(But google-api-client still declares commons-logging via httpclient. So maybe this experiment does not show anything)


suztomo:
It worked.

PR #2403, opened 2024-02-08
feat: add soft delete feature

JesseLovelace:
Implements the soft delete feature as requested in go/gcs-soft-delete-client-request

Also adds the `get(String blob, Long generation, BlobGetOption...options)` method signature to Bucket.java. It's already present in Storage.java, I'm adding it now because it's necessary to get a soft deleted object.


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -540,7 +541,9 @@ private Storage.Objects.Get getCall(StorageObject object, Map<Option, ?> options
         .setIfGenerationMatch(Option.IF_GENERATION_MATCH.getLong(options))
         .setIfGenerationNotMatch(Option.IF_GENERATION_NOT_MATCH.getLong(options))
         .setFields(Option.FIELDS.getString(options))
-        .setUserProject(Option.USER_PROJECT.getString(options));
+        .setUserProject(Option.USER_PROJECT.getString(options))
+        .setSoftDeleted(Option.SOFT_DELETED.getBoolean(options));
+    // .setGeneration(Option.GENERATION.getLong(options));
```
frankyn, line 546: Might not be intentional

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -562,6 +565,38 @@ public StorageObject get(StorageObject object, Map<Option, ?> options) {
...
+          storage
+              .objects()
+              .restore(object.getBucket(), object.getName(), object.getGeneration(), object);
+      return restore
+          .setProjection(DEFAULT_PROJECTION)
+          .setIfMetagenerationMatch(Option.IF_METAGENERATION_MATCH.getLong(options))
+          .setIfMetagenerationNotMatch(Option.IF_METAGENERATION_NOT_MATCH.getLong(options))
+          .setIfGenerationMatch(Option.IF_GENERATION_MATCH.getLong(options))
+          .setIfGenerationNotMatch(Option.IF_GENERATION_NOT_MATCH.getLong(options))
+          .setCopySourceAcl(Option.COPY_SOURCE_ACL.getBoolean(options))
```
frankyn, line 583: This is an interesting addition to the API; surprised it needs to be added if you're restoring an object.

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
```diff
@@ -1083,6 +1089,28 @@ public Blob get(String blob, BlobGetOption... options) {
...
+   * Blob blob = bucket.get(blobName, generation);
+   * }</pre>
+   *
+   * @param blob name of the requested blob
+   * @param generation the generation to get
+   * @param options blob search options
+   * @throws StorageException upon failure
+   */
+  @TransportCompatibility({Transport.HTTP, Transport.GRPC})
+  public Blob get(String blob, Long generation, BlobGetOption... options) {
```
frankyn, line 1110: IIUC this is needed because you can't specify generation on Bucket.get() that exists today; Would recommend using BlobId instead of adding Long generation overload.

Do you think this is hard requirement to support this feature since there's [`storage.get(BlobId..)`](https://github.com/googleapis/java-storage/blob/main/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java#L3033)?

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -375,6 +378,13 @@ private Bucket bucketInfoEncode(BucketInfo from) {
     ifNonNull(from.getAcl(), toImmutableListOf(bucketAclCodec::encode), to::addAllAcl);
     ifNonNull(from.getIamConfiguration(), iamConfigurationCodec::encode, to::setIamConfig);
     ifNonNull(from.getAutoclass(), autoclassCodec::encode, to::setAutoclass);
+    ifNonNull(from.getSoftDeletePolicy(), softDeletePolicyCodec::encode, to::setSoftDeletePolicy);
+    if (from.getModifiedFields().contains(SOFT_DELETE_POLICY)
+        && from.getSoftDeletePolicy() == null) {
+      System.out.println("this is happening");
```
frankyn, line 384: debug code

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBlobReadMaskTest.java
```diff
@@ -201,7 +201,9 @@ public ImmutableList<?> parameters() {
               new Args<>(BlobField.UPDATED, LazyAssertion.equal()),
               new Args<>(
                   BlobField.RETENTION,
-                  LazyAssertion.skip("TODO: jesse fill in buganizer bug here")));
+                  LazyAssertion.skip("TODO: jesse fill in buganizer bug here")),
```
frankyn, line 204: I understand this code already exists; just wondering why is Retention field mask test skipped?

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -402,6 +403,33 @@ public Blob get(BlobId blob) {
...
+
+  private Blob internalObjectRestore(BlobId blobId, Opts<ObjectSourceOpt> opts) {
+    Opts<ObjectSourceOpt> finalOpts = opts.prepend(defaultOpts).prepend(ALL_BLOB_FIELDS);
+    GrpcCallContext grpcCallContext =
+        finalOpts.grpcMetadataMapper().apply(GrpcCallContext.createDefault());
+    RestoreObjectRequest.Builder builder =
+        RestoreObjectRequest.newBuilder()
+            .setBucket(bucketNameCodec.encode(blobId.getBucket()))
+            .setObject(blobId.getName());
+    ifNonNull(blobId.getGeneration(), builder::setGeneration);
```
frankyn, line 420: Is the API failure helpful when generation isn't supplied?

frankyn (commented):
Have a couple of questions. Thanks Jesse

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITBlobReadMaskTest.java
JesseLovelace: This is related to object retention being JSON only. This will be removed once it's in the grpc api as well

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
JesseLovelace: Yeah, it says "you must specify a generation"

google-cloud-storage/src/main/java/com/google/cloud/storage/Bucket.java
JesseLovelace: I think it would feel pretty weird that soft deleted would become the only state an object can be in where you can get it from `Storage` but can't get it from the `Bucket` that the object lives in.

I also think using `BlobId` here would feel weird, because you're already calling from a Bucket, so it's weird to have to supply the same Bucket name into the BlobId. There are no other methods in `Bucket.java` that have a `BlobId` in the signature

frankyn: Bleh, missed that bucket name is also part of BlobId. thanks.
We haven't had generation in Bucket.get() at all which is required for version enabled buckets, that's why i asked this question as well.

For the sake of discussion, WDYT about having generation under BlobGetOption's? It's a query paramter just like IfGenerationMatch etc.

frankyn (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -16,11 +16,8 @@
 
 package com.google.cloud.storage;
 
-import static com.google.cloud.storage.Utils.bucketNameCodec;
-import static com.google.cloud.storage.Utils.ifNonNull;
-import static com.google.cloud.storage.Utils.lift;
-import static com.google.cloud.storage.Utils.projectNameCodec;
-import static com.google.cloud.storage.Utils.topicNameCodec;
+import static com.google.cloud.storage.Storage.BucketField.SOFT_DELETE_POLICY;
+import static com.google.cloud.storage.Utils.*;
```
frankyn, line 20: Revert `com.google.cloud.storage.Utils.*;`

google-cloud-storage/src/main/java/com/google/cloud/storage/HttpRetryAlgorithmManager.java
```diff
@@ -213,6 +213,11 @@ public ResultRetryAlgorithm<?> getForObjectsGet(
     return retryStrategy.getIdempotentHandler();
   }
 
+  public ResultRetryAlgorithm<?> getForObjectsRestore(
+      StorageObject pb, Map<StorageRpc.Option, ?> optionsMap) {
+    return retryStrategy.getIdempotentHandler();
```
frankyn, line 218: Missed this on first pass; ObjectRestore is idempotent because it can only succeed once which is similar to Bucket; is that your understanding as well?

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpcSpans.java
```diff
@@ -30,6 +30,8 @@ class HttpStorageRpcSpans {
   static final String SPAN_NAME_LIST_OBJECTS = getTraceSpanName("list(String,Map)");
   static final String SPAN_NAME_GET_BUCKET = getTraceSpanName("get(Bucket,Map)");
   static final String SPAN_NAME_GET_OBJECT = getTraceSpanName("get(StorageObject,Map)");
+
```
frankyn, line 33: remove extra line

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/StorageRpc.java
```diff
@@ -73,7 +73,11 @@ enum Option {
     DETECT_CONTENT_TYPE("detectContentType"),
     ENABLE_OBJECT_RETENTION("enableObjectRetention"),
     RETURN_RAW_INPUT_STREAM("returnRawInputStream"),
-    OVERRIDE_UNLOCKED_RETENTION("overrideUnlockedRetention");
+    OVERRIDE_UNLOCKED_RETENTION("overrideUnlockedRetention"),
+    SOFT_DELETED("softDeleted"),
+
```
frankyn, line 78: remove extra line

frankyn (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/HttpRetryAlgorithmManager.java
JesseLovelace: Yes, restore has the same idempotency as Copy and similar operations

PR #2413, opened 2024-02-14
chore: use sdk-platform-java-config to consolidate build configs

mpeddada1:
Notable Changes:
1) Use `gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform*` docker images for Kokoro GraalVM tests  instead of `gcr.io/cloud-devrel-kokoro-resources/graalvm*`.
2) Use `com.google.cloud:sdk-platform-java-config` as the parent which inherits configs from `java-shared-config` and hosts the `google-cloud-shared-dependencies` version under the `google-cloud-shared-dependencies.version` property.  This artifact is versioned to be the same as google-cloud-shared-dependencies.
3) Adjust renovate-bot settings to update docker images when a new version of `sdk-platform-java-config` is on Maven Central.

Example renovate-bot update PR in google-cloud-java: https://github.com/googleapis/google-cloud-java/pull/10290


burkedavison (approved):

sydney-munro (commented):
no real issues with the changes. A couple of questions which were also pinged offline. copying here for future reference.

will this continue to trigger all the normal workflows? such as release please etc

also the tickets we are cut during the release train all track shared deps. Will they be cut and unblocked using the same processes with this new dependency

sydney-munro:
> no real issues with the changes. A couple of questions which were also pinged offline. copying here for future reference.
> 
> will this continue to trigger all the normal workflows? such as release please etc
>  
> also the tickets we are cut during the release train all track shared deps. Will they be cut and unblocked using the same processes with this new dependency
@mpeddada responded that release workflows will continue but tickets will need to be updated to block and unblock.

PR #2417, opened 2024-02-15
docs: add samples for object retention

JesseLovelace:
Adds object retention samples and tests

### DIFF
--- a/samples/snippets/src/main/java/com/example/storage/bucket/CreateBucketWithObjectRetention.java
+++ b/samples/snippets/src/main/java/com/example/storage/bucket/CreateBucketWithObjectRetention.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.storage.bucket;
+
+// [START storage_create_bucket_with_object_retention]
+
+import com.google.cloud.storage.Bucket;
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+
+public class CreateBucketWithObjectRetention {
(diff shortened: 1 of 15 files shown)


samples/snippets/src/main/java/com/example/storage/object/GetObjectMetadata.java
```diff
@@ -68,6 +68,7 @@ public static void getObjectMetadata(String projectId, String bucketName, String
     System.out.println("StorageClass: " + blob.getStorageClass());
     System.out.println("TimeCreated: " + new Date(blob.getCreateTime()));
     System.out.println("Last Metadata Update: " + new Date(blob.getUpdateTime()));
+    System.out.println("Object Retention Policy: " + blob.getRetention());
```
frankyn, line 71: GetBucketMetadata and this sample are missing test changes; please add them.

samples/snippets/src/main/java/com/example/storage/bucket/CreateBucketWithObjectRetention.java
```diff
@@ -0,0 +1,48 @@
...
+
+// [START storage_create_bucket_with_object_retention]
+
+import com.google.cloud.storage.Bucket;
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+
+public class CreateBucketWithObjectRetention {
+  public static void createBucketWithObjectRetention(String projectId, String bucketName) {
```
frankyn, line 27: Is this sample required in docs? Wondering if users could be pointed to `BucketTargetOption.enableObjectRetention(true) instead of writing this sample in 8 languages.

samples/snippets/src/main/java/com/example/storage/object/SetObjectRetentionPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+                    .setRetainUntilTime(now().plusDays(10))
+                    .build())
+            .build()
+            .update();
+
+    System.out.println("Retention policy for object " + objectName + " was set to:");
+    System.out.println(updated.getRetention().toString());
+
+    // To modify an existing policy on an Unlocked object, pass in the override parameter
+    blob.toBuilder()
```
frankyn, line 64: This looks like two samples in one; were you planning on splitting it or keeping as-is?

JesseLovelace: Planning to keep as is. I think it's important that a user sees both cases, so keeping them in one sample makes sure that it can't be missed.

samples/snippets/src/main/java/com/example/storage/bucket/CreateBucketWithObjectRetention.java
frankyn: sg, thanks for clarifying.

PR #2426, opened 2024-02-27
chore(benchmarking): Remove retries and switch to publishing api_latency instead of throughput

sydney-munro:
In order to better compare performance of gRPC and JSON I have removed retries to keep retried requests from impacting latency metrics. We can monitor failed metrics and perhaps in the future publish metrics around if a request would have been retried or not.

I also believe publishing latency is more useful than throughput. If we want throughput the math is simpler to conduct in the dashboards than the other way around.


frankyn (approved):
One question; otherwise LGTM.

PR #2445, opened 2024-03-12
feat: add hierarchical namespace and folders features

JesseLovelace:
Adds support for hierarchical namespace and folders. Copied over from an internal workspace


BenWhitehead (commented):
Superficially LGTM, waiting on @frankyn for final approval as I believe he's more familiar with this

frankyn (approved):

PR #2455, opened 2024-03-19
build: include junit-vintage-engine in test dependencies

BenWhitehead:
Include junit-vintage-engine in test dependencies for google-cloud-storage to ensure junit4 and junit5 tests are both run during `mvn test`.

Related b/329865405


pom.xml
```diff
@@ -89,30 +89,11 @@
...
-        <artifactId>easymock</artifactId>
-        <version>4.3</version>
-        <scope>test</scope>
-      </dependency>
       <dependency>
         <groupId>org.mockito</groupId>
         <artifactId>mockito-core</artifactId>
         <version>4.11.0</version>
         <scope>test</scope>
-        <exclusions>
```
BenWhitehead, line 103: let objenesis come in for mockito, since we don't depend on it directly and are no longer needing to disambiguate it between easymock.

sydney-munro (approved):

PR #2469, opened 2024-03-25
chore: Transfer Manager Parallel Composite Uploads Prework

sydney-munro:
Performing a chunked upload vs a chunked download are very different operations and should be able to be configured separately. This renames the previous allowDivideAndConquer to be allowDivideAndConquerDownload and introduces a new field specific to ParallelCompositeUploads.


BenWhitehead (approved):

BenWhitehead:
Changed the title from `feat` to `chore` since this is pre-work. When TM is leveraging the PCU implementation is when we'd want the `feat` in the title.

PR #2472, opened 2024-03-27
feat: port DefaultBlobWriteSessionConfig to work with HttpStorageOptions

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultBlobWriteSessionConfig.java
```diff
@@ -146,6 +150,39 @@ public WritableByteChannelSession<?, BlobInfo> writeSession(
...
             WRITE_OBJECT_RESPONSE_BLOB_INFO_DECODER);
+      } else if (s instanceof StorageImpl) {
+        StorageImpl json = (StorageImpl) s;
+
+        return new DecoratedWritableByteChannelSession<>(
+            new LazySession<>(
+                new LazyWriteChannel<>(
+                    () -> {
+                      final Map<StorageRpc.Option, ?> optionsMap = opts.getRpcOptions();
+                      BlobInfo.Builder builder = info.toBuilder().setMd5(null).setCrc32c(null);
```
sydney-munro, line 161: why do we need to set Md5 and CRC32C to null here?

sydney-munro (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/DefaultBlobWriteSessionConfig.java
BenWhitehead: This is replicated to follow the pattern of all the other write methods, for example https://github.com/googleapis/java-storage/blob/e1fb8577e6b7a36981fb78937fe8c5dd606a4c2a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java#L696

The reasoning why this is done, is due to the fact if the checksum is present in the object GCS will enforce it, and we only want to enforce it if the option `crc32cMatch()` or `md5Match()` are provided. At which point the files will be set back to non-null in the following line with `opts.blobInfoMapper().apply(builder).build()`.

The usage pattern this enables is the read-modify-write of an object, i.e.
```
// assume storage.createFrom(info, Path.get("/file/on/disk/1")); created the current generation
Blob gen1 = storage.get(BlobId.of("bucket", "object"));
Blob gen2 = storage.createFrom(gen1, Paths.get("/file/on/disk/2"));
```

PR #2481, opened 2024-04-01
fix: fix BidiBlobWriteSessionConfig to respect preconditions

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -598,6 +598,14 @@ public Mapper<WriteObjectRequest.Builder> writeObject() {
       };
     }
 
+    @Override
+    public Mapper<BidiWriteObjectRequest.Builder> bidiWriteObject() {
+      return b -> {
+        b.getObjectChecksumsBuilder().setCrc32C(val);
```
frankyn, line 604: How did this get past tests? (unless it was part of the JUnit test issue you're working on)

frankyn (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
BenWhitehead: The tests that were added for bidi specifically didn't tests any preconditions. I'm adding some tests that will run for every BlobWriteSessionConfig to verify they all handle semantics similarly.

frankyn: It's subtle i would likely make the same mistake implementing options; thanks ben!

PR #2493, opened 2024-04-08
fix: update ApiaryUnbufferedWritableByteChannel to be graceful of non-quantum aligned write calls

BenWhitehead:
Despite GCS only allowing incremental updates to a resumable session at 256KiB byte boundaries, we have observed an extremely rare case (<0.000001%) of an incremental write being non-quantum aligned. The change in this PR makes ApiaryUnbufferedWritableByteChannel graceful to this possibility, and will only set the finalization header when close is invoked.

If a write is not quantum aligned, it will either: 1) not be consumed at all, at which point write can be called again with the still enqueued bytes 2) partially consumed, with the position of the provided ByteBuffers updated to reflect how much of their bytes were consumed, matching up with the number of bytes actually consumed returned from `write()`

Add new integration test to intentionally perform non-quantum aligned `write()` calls to ApiaryUnbufferedWritableByteChannel.

Separate change to gRPC affected code path to come in a later PR.

b/330550326

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryUnbufferedWritableByteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryUnbufferedWritableByteChannel.java
@@ -96,30 +96,32 @@ final class ApiaryUnbufferedWritableByteChannel implements UnbufferedWritableByt
     }
     RewindableContent content = RewindableContent.of(Utils.subArray(srcs, offset, length));
     long available = content.getLength();
+    // as long as request has at least 256KiB GCS will accept bytes in 256KiB increments,
+    // however if a request is smaller than 256KiB it MUST be the finalization request.
+    if (!finalize && available < ByteSizeConstants._256KiB) {
+      return 0;
+    }
     long newFinalByteOffset = cumulativeByteCount + available;
     final HttpContentRange header;
     ByteRangeSpec rangeSpec = ByteRangeSpec.explicit(cumulativeByteCount, newFinalByteOffset);
-    boolean quantumAligned = available % ByteSizeConstants._256KiB == 0;
-    if (quantumAligned && finalize) {
+    if (finalize) {
       header = HttpContentRange.of(rangeSpec, newFinalByteOffset);
       finished = true;
-    } else if (quantumAligned) {
+    } else {
       header = HttpContentRange.of(rangeSpec);
-    } else { // not quantum aligned, have to finalize
-      header = HttpContentRange.of(rangeSpec, newFinalByteOffset);
-      finished = true;
     }
     try {
       ResumableOperationResult<@Nullable StorageObject> operationResult =
           session.put(content, header);
       long persistedSize = operationResult.getPersistedSize();
       committedBytesCallback.accept(persistedSize);
+      long written = persistedSize - cumulativeByteCount;
       this.cumulativeByteCount = persistedSize;
       if (finished) {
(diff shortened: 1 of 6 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java
```diff
@@ -110,51 +110,6 @@ public void emptyObjectHappyPath() throws Exception {
...
-   * <h4>S.9</h4>
-   *
-   * Partial successful append to session
-   *
-   * <p>The client has sent N bytes, the server confirmed N bytes as committed. The client sends K
-   * bytes starting at offset N. The server responds with only N + L with 0 &lt;= L &lt; K bytes as
-   * committed.
-   */
-  @Test
-  public void scenario9() throws Exception {
```
BenWhitehead, line 125: This test would verify that GCS ack'ing fewer bytes than were sent was correctly categorized as a failure scenario, however as we are now gracefully handling the partial write case. We no longer need this test for failure categorization.

google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryUnbufferedWritableByteChannel.java
```diff
@@ -96,30 +96,32 @@ private long internalWrite(ByteBuffer[] srcs, int offset, int length, boolean fi
     }
     RewindableContent content = RewindableContent.of(Utils.subArray(srcs, offset, length));
     long available = content.getLength();
+    // as long as request has at least 256KiB GCS will accept bytes in 256KiB increments,
+    // however if a request is smaller than 256KiB it MUST be the finalization request.
+    if (!finalize && available < ByteSizeConstants._256KiB) {
```
frankyn, line 101: does this allow for 256KiB incremental writes?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITUnbufferedResumableUploadTest.java
```diff
@@ -0,0 +1,117 @@
...
+    long size = objectSize + additional;
+    ByteBuffer b = DataGenerator.base64Characters().genByteBuffer(size);
+
+    UnbufferedWritableByteChannel open = session.open();
+    int written = open.write(b);
+    assertThat(written).isEqualTo(objectSize);
+    assertThat(b.remaining()).isEqualTo(additional);
+
+    int writtenAndClose = open.writeAndClose(b);
+    assertThat(writtenAndClose).isEqualTo(additional);
```
frankyn, line 111: Can you add a test to validate the rewind step works as expected when unaligned PUTs are partially received to align with 256kib?

google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryUnbufferedWritableByteChannel.java
frankyn: answered through offline discussion

google-cloud-storage/src/test/java/com/google/cloud/storage/ITUnbufferedResumableUploadTest.java
BenWhitehead: This is validated by the assertion above `assertThat(b.remaining()).isEqualTo(additional);` the only way that assertion will succeed is by the rewind having taken place.

PR #2494, opened 2024-04-10
feat: Transfer Manager ParallelCompositeUploads

sydney-munro:
This implements ParallelCompositeUploads(PCU) in Transfer Manager.

ParallelCompositeUpload in Transfer Manager is an opt-in feature, customers will have to update their config to allow Transfer Manager to perform PCU operations on their behalf. Transfer Manager will only do this if the following conditions are met: 

- The customer has set allowParallelCompositeUpload to true
- The file to be uploaded is 4 * the buffer per worker size
- The customer has specified enough threads that ParallelCompositeUpload can be of benefit while still cuing other work.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
@@ -263,8 +263,9 @@ final class TransferManagerImpl implements TransferManager {
    * all the threads waiting for PCU uploads to complete, the PCU work doesn't have any threads
    * available to itself.
    *
-   * This class represents a single worker that will be submitted to the executor service and will
-   * poll a queue to process a single PCU at a time, leaving any other threads free for PCU work.
+   * <p>This class represents a single worker that will be submitted to the executor service and
+   * will poll a queue to process a single PCU at a time, leaving any other threads free for PCU
+   * work.
    */
   private final class PcuPoller implements Runnable {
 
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
```diff
@@ -67,6 +81,26 @@ final class TransferManagerImpl implements TransferManager {
...
+              .setBlobWriteSessionConfig(pcuConfig)
+              .build();
+    } else if (storageOptions instanceof HttpStorageOptions
+        && transferManagerConfig.isAllowParallelCompositeUpload()) {
+      storageOptions =
+          ((HttpStorageOptions) storageOptions)
+              .toBuilder()
+              .setBlobWriteSessionConfig(pcuConfig)
+              .build();
+    }
```
BenWhitehead, line 102: If you rebase against `main` this isn't necessary any longer, `setBlobWriteSessionConfig` has been added to `StorageOptions`

```suggestion
    // Create the blobWriteSessionConfig for ParallelCompositeUpload
    if (transferManagerConfig.isAllowParallelCompositeUpload()) {
      ParallelCompositeUploadBlobWriteSessionConfig pcuConfig =
          BlobWriteSessionConfigs.parallelCompositeUpload()
              .withExecutorSupplier(ExecutorSupplier.useExecutor(executor));
      storageOptions =
          storageOptions
              .toBuilder()
              .setBlobWriteSessionConfig(pcuConfig)
              .build();
    }
```

BenWhitehead (changes requested):
One cleanup item related to options.

Can you also purge out the owlbot commit, it's adding noise to the PR unrelated to your change.

BenWhitehead (approved):

PR #2499, opened 2024-04-17
fix: update grpc handling of IAM Policy etag to account for base64 encoding

BenWhitehead:
Etags returned by the JSON api are base64 encoded, and IAM Policy is currently modeled around this for its public api. Update GrpcConversions to base64 decode/encode appropriately.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
@@ -60,11 +60,13 @@ import com.google.storage.v2.ObjectChecksums;
 import com.google.storage.v2.Owner;
 import com.google.type.Date;
 import com.google.type.Expr;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.OffsetDateTime;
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -1058,7 +1065,7 @@ private Policy policyDecode(com.google.iam.v1.Policy from) {
     Policy.Builder to = Policy.newBuilder();
     ByteString etag = from.getEtag();
     if (!etag.isEmpty()) {
-      to.setEtag(etag.toStringUtf8());
+      to.setEtag(byteStringB64StringCodec.encode(etag));
```
frankyn, line 1068: this is very subtle how hasn't this come up before?

BenWhitehead: I think setIamPolicy so far has been exclusively used in json or grpc, and not yet mixed. With the conformance tests we did get mixing, because the policy is resolved in setup via json the modified via grpc.

frankyn: subtle; thanks!

frankyn (approved):

PR #2512, opened 2024-04-24
fix: update GzipReadableByteChannel to be tolerant of one byte reads

BenWhitehead:
If the ReadChannel chunkSize is 0 there will be no library buffering performed during read calls. If gzip decompression support is enabled on the read channel, there is a possibility someone could read fewer than the 4 bytes we initially read.

This change updates the byte tracking to ensure single byte reads are supported.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GzipReadableByteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GzipReadableByteChannel.java
@@ -36,6 +36,7 @@ final class GzipReadableByteChannel implements UnbufferedReadableByteChannel {
 
   private boolean retEOF = false;
   private ScatteringByteChannel delegate;
+  private ByteBuffer leftovers;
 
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GzipReadableByteChannel.java
```diff
@@ -51,8 +52,8 @@ public long read(ByteBuffer[] dsts, int offset, int length) throws IOException {
     // if our delegate is null, that means this is the first read attempt
     if (delegate == null) {
       // try to determine if the underlying data coming out of `source` is gzip
-      byte[] first4 = new byte[4]; // 4 bytes = 32-bits
-      final ByteBuffer wrap = ByteBuffer.wrap(first4);
+      byte[] firstByte = new byte[1];
+      ByteBuffer wrap = ByteBuffer.wrap(firstByte);
```
frankyn, line 56: comments below should be updated.

frankyn (approved):

PR #2537, opened 2024-05-08
fix: update GapicUnbufferedChunkedResumableWritableByteChannel to be tolerant of non-quantum aligned writes

BenWhitehead:
Update GapicUnbufferedChunkedResumableWritableByteChannel to only accept bytes at the 256KiB boundary when `write(ByteBuffer[], int, int)` is called.

Calls to `writeAndClose(ByteBuffer[], int, int)` will consume all bytes.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
@@ -18,7 +18,10 @@ package com.google.cloud.storage;
 
 import com.google.cloud.storage.Crc32cValue.Crc32cLengthKnown;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.math.IntMath;
 import com.google.protobuf.ByteString;
+import java.math.RoundingMode;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
```diff
@@ -79,32 +89,92 @@ ChunkSegment[] segmentBuffers(ByteBuffer[] bbs) {
...
+    return segmentBuffers(bbs, offset, length, true);
+  }
+
+  ChunkSegment[] segmentBuffers(
+      ByteBuffer[] bbs, int offset, int length, boolean allowUnalignedBlocks) {
+    // turn this into a single branch, rather than multiple that would need to be checked each
+    // element of the iteration
+    if (allowUnalignedBlocks) {
+      return segmentWithUnaligned(bbs, offset, length);
+    } else {
```
sydney-munro, line 101: will we ever reach this?

BenWhitehead: This will actually be the normal case. For a resumable upload, this will stop consuming bytes unless there is a block worth of bytes remaining.

sydney-munro (approved):

Issue #2308, opened 2023-11-17
Improve Virtual Thread compatibility


BenWhitehead:
With the release of Java 21 [Virtual Threads](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html) are now a first class feature.

To better support the use of virtual threads in applications we will need to evaluate and update certain assumptions and aspects the library is built upon today.

[Virtual Threads: An Adoption Guide](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-B163BC51-039B-43BD-87ED-27BE384B509D) outlines several considerations we should keep in mind when doing this work.

Items to investigate:
1. [ ] `synchronized` method and blocks
2. [ ] Existing locking strategies
3. [ ] ExecutorService usage
4. [ ] Context propagation with other libraries (OTel, OpenCensus)
5. [ ] Parallel Stream usage
6. [ ] other `java.util.concurrent` usages

cpoissonnier:
We are impacted by thread pinning due to this lib. Any improvements would be greatly appreciated 😊

PR #2553, opened 2024-05-17
fix: reduce Java 21 Virtual Thread Pinning in IO operations

BenWhitehead:
Setup a small workload using virtual threads and turned on pinning detection (`-Djdk.tracePinnedThreads=full`), and run various methods for uploading and downloading object data.

I then reduced the pinning sights to a minimized list (`grep 'monitor' run.log | grep 'com.google.cloud.storage' | sort -u`)
```
    com.google.cloud.storage.BaseStorageReadChannel.read(BaseStorageReadChannel.java:105) <== monitors:1
    com.google.cloud.storage.BlobWriteSessions$WritableByteChannelSessionAdapter.open(BlobWriteSessions.java:46) <== monitors:1
    com.google.cloud.storage.LazyWriteChannel.getSession(LazyWriteChannel.java:59) <== monitors:1
    com.google.cloud.storage.ParallelCompositeUploadWritableByteChannel.close(ParallelCompositeUploadWritableByteChannel.java:268) <== monitors:1
    com.google.cloud.storage.ParallelCompositeUploadWritableByteChannel.write(ParallelCompositeUploadWritableByteChannel.java:178) <== monitors:1
    com.google.cloud.storage.StorageByteChannels$SynchronizedBufferedWritableByteChannel.close(StorageByteChannels.java:119) <== monitors:1
    com.google.cloud.storage.StorageByteChannels$SynchronizedBufferedWritableByteChannel.write(StorageByteChannels.java:109) <== monitors:1
    com.google.cloud.storage.StorageByteChannels$SynchronizedUnbufferedReadableByteChannel.read(StorageByteChannels.java:139) <== monitors:1
```

To avoid pinning during IO operations, uses of synchronization have been replaced with ReenterantLock. There were a few synchronization sites that were also guarded by an outer operations synchronization, these too have been updated to use ReenterantLock.

We stick with `synchronized` for init related operations. In the case of creating a WriteChannel, the ability to open depends on the resumable session being created successfully, which requires a network request. Our objects in this area are not written in a top to bottom async friendly way to remove the synchronization at this time. Given these operations are performed at most once per WriteChannel impact is much less drastic compared to `write(ByteBuffer)` and `close()`.

After this change, I ran the same workload and we are left with only the two following pins detected
```
    com.google.cloud.storage.BlobWriteSessions$WritableByteChannelSessionAdapter.open(BlobWriteSessions.java:46) <== monitors:1
    com.google.cloud.storage.LazyWriteChannel.getSession(LazyWriteChannel.java:59) <== monitors:1
```

Some investigation into if/how this type of pinning detection can be introduced to integration suite still needs to happen. Superficially, I'm not sure if it's practical to get JUnit4 to run everything in virtual threads, we might need to depend on some other CI external validation.

Part of work needed for #2308

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/AsyncAppendingQueue.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/AsyncAppendingQueue.java
@@ -90,7 +90,7 @@ final class AsyncAppendingQueue<@NonNull T> implements AutoCloseable {
     lock = new ReentrantLock();
   }
 
-  synchronized AsyncAppendingQueue<T> append(ApiFuture<T> value) throws ShortCircuitException {
+  AsyncAppendingQueue<T> append(ApiFuture<T> value) throws ShortCircuitException {
     lock.lock();
     try {
       checkState(state.isOpen(), "already closed");
@@ -124,7 +124,7 @@ final class AsyncAppendingQueue<@NonNull T> implements AutoCloseable {
   }
 
   @Override
-  public synchronized void close() {
+  public void close() {
     lock.lock();
     try {
       if (!state.isOpen()) {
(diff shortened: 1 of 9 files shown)


JesseLovelace (approved):
LGTM, thanks!

PR #2556, opened 2024-05-21
test: add new tests around Autoclass modification

BenWhitehead:


google-cloud-storage/pom.xml
```diff
@@ -18,6 +18,8 @@
     <site.installationModule>google-cloud-storage</site.installationModule>
     <pubsub-proto.version>1.111.3</pubsub-proto.version>
     <junit-platform.version>5.10.2</junit-platform.version>
+    <!-- Set a single test instead of the full suite -->
+    <test>com.google.cloud.storage.it.ITAutoclassTest</test>
```
sydney-munro, line 22: what is the impact of this?

BenWhitehead: This will constrain the Integration suite to only run that single class rather than all integration tests. I don't need the PR open any more, so I'll close it.

PR #2566, opened 2024-05-28
build: enable hermetic library generation

diegomarquezp:
This is part of the effort to enable multi-version support in hand-written libraries. More context and details in [this doc](https://docs.google.com/document/d/1wrpyBtphdenM3BNelcnpBKGADYrGJUo686HXvSA0h-0/edit?pli=1&tab=t.0#heading=h.lbr9t8ls3oip).

This PR adds two workflows and their scripts:
 * `update_generation_config` which updates `generation_config.yaml` (added in this PR) to contain the latest `googleapis_committish` (source of truth of library generation) as well as other dependencies such as `gapic_generator_version`. This job runs nightly and creates a PR that modifies the config yaml.
 * `hermetic_library_generation` which reads `generation_config.yaml` and pushes a new commit to any PR whose latest commit affects `generation_config.yaml` (e.g. the nightly PR created by `update_generation_config`). It essentially translates changes in the config yaml to changes in the source code (of the generated part). This workflow will be a required check.

As an example of how they work in practice:
 1. Once per night, `update_generation_config` creates a PR that updates `generation_config.yaml` with the latest dependencies and googleapis committish
 2. `hermetic_library_generation` is triggered by this PR and generates the corresponding code changes. A new commit with the code changes is added to the PR. This step may occur for any PR affecting `generation_config.yaml`.

### Follow ups
 - [x] Add `CLOUD_JAVA_BOT_TOKEN` to repository for the hermetic generation workflow to work. We confirmed it works in a [fork](https://github.com/diegomarquezp/java-storage/pull/10) ([job link](https://github.com/diegomarquezp/java-storage/pull/10)).


generation_config.yaml
```diff
@@ -0,0 +1,25 @@
+gapic_generator_version: 2.42.0
+googleapis_commitish: b2c018cb067ab01cb5db2170b73a82caef173ae4
+libraries_bom_version: 26.38.0
```
BenWhitehead, line 3: Will `libraries_bom_version` and `gapic_generator_version` be updated by renovate bot, or the update_generation_config action?

.github/workflows/update_generation_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+  schedule:
+  - cron: '0 2 * * *'
+  workflow_dispatch:
+
+jobs:
+  update-generation-config:
+    runs-on: ubuntu-22.04
+    env:
+      # the branch into which the pull request is merged
+      base_branch: main
```
BenWhitehead, line 27: When we make branches for lts releases will this need to be manually changed?

generation_config.yaml
diegomarquezp: They will be updated by `update_generation_config`

https://github.com/googleapis/java-storage/blob/68ee05a4c47acd2709412cab1b461d89d8dc0014/.github/scripts/update_generation_config.sh#L97-L103

BenWhitehead: Could that cause any issues if the version in the library or samples updates via renovate-bot first?

.github/workflows/update_generation_config.yaml
blakeli0: It depends on the type of LTS. 

- If it is an LTS that does not need regular updates to googleapis_commitish or generator version, basically the LTS branches we current have. Then we would not set up any automation for it, and only change the generator version when there is a critical bug fix.
- If it is an LTS that requires regular updates, for example, if we want to introduce a Java 21 branch, then yes we will need to update this config on the Java 21 branch.

generation_config.yaml
blakeli0: `libraries_bom_version` is only used to generate the [version in README](https://github.com/googleapis/java-storage/blame/main/README.md#L18-L26) by the hermetic build scripts, it should not cause any conflicts with the [libraries bom version](https://github.com/googleapis/java-storage/blob/bbcb4fbaf9167af0c24d317629fc3bc55fb52927/samples/snippets/pom.xml#L34) in samples.

There is no `gapic_generator_version` in pom, I guess you mean [sdk-patform-java-config](https://github.com/googleapis/java-storage/blame/main/pom.xml#L15-L17), which includes gax changes? If yes, this is similar to what we have right now with owlbot PRs. 
The owlbot PR changes from the generator update may require gax changes. Similarly, the Cloud Java Bot PR may require us to merge the `sdk-patform-java-config` update first. But there should not be any issues the other way around, we should be free to merge `sdk-patform-java-config` update anytime, without waiting on Cloud Java Bot PRs created from the `gapic_generator_version` update.

blakeli0 (approved):

generation_config.yaml
BenWhitehead: Thanks for clarifying

BenWhitehead (approved):
I'm okay with the changes as outlined here. How do we get the status check to actually turn green?

diegomarquezp:
> I'm okay with the changes as outlined here. How do we get the status check to actually turn green?

@BenWhitehead We confirmed it works in a https://github.com/diegomarquezp/java-storage/pull/10 (https://github.com/diegomarquezp/java-storage/pull/10).

It requires `CLOUD_JAVA_BOT_TOKEN` to be added, which is a follow up item.

PR #2514, opened 2024-04-24
chore: remove WriteFlusher instead having dedicated WritableByteChannels

BenWhitehead:
WriteFlusher has grown unwieldy to be able to properly handle the kinds of uploads that are performed while being able to perform appropriate fine grained response validation.

This is the first change in a series to remove WriteFlusher and create purpose dedicated UnbufferedWritableByteChannels for the different kinds of uploads we perform. By having dedicated channels more fine grained response validation and request packing can be performed.

I left multiple commits in this PR for the time being. They are generally:
1. Introduce GapicUnbufferedDirectWritableByteChannel used for direct/single-shot uploads
2. Introduce GapicUnbufferedChunkedResumableWritableByteChannel used for chunked resumable uploads
3. Introduce GapicUnbufferedFinalizeOnCloseResumableWritableByteChannel used for for resumable uploads from files on disk
4. Refactor GapicBidiUnbufferedWritableByteChannel
5. Cleanup and removal of the majority of WriteFlushStrategy.java and its corresponding tests now that the functionality is distributed in each of the dedicated channel implementations

PR #2570, opened 2024-05-31
fix: update grpc bidi resumable uploads to validate ack'd object size

BenWhitehead:
Follow up to #2527 and #2514

The tests added in this PR account for ~940 of the added lines, however their scenarios are the same as the other upload scenario tests just using the different proto types. The remaining ~300 lines are the fix and small refactor to get the tests passing. I've left individual commits in the PR to make it easier to drill down if needed.

* Part 1, add tests and simulation server
  * Tests are a port of those added in #2527 using BidiWriteObjectRequest and BidiWriteObjectResponse
  * simulation server is a port of com.google.cloud.storage.ITGapicUnbufferedWritableByteChannelTest.DirectWriteService using BidiWriteObjectRequest and BidiWriteObjectResponse

* Cleanup unused inner-class

* Move request trimming inline rather than at the end

* Update GapicBidiUnbufferedWritableByteChannel to correctly handle failure scenarios {1, 2, 3, 4, 4_1, 4_2, 5, 7}
    
  Refactor ResumableSessionFailureScenario to accept Message and List rather than WriteObjectResponse and List<WriteObjectRequest>.
    
  The shape of BidiWriteObjectRequest/WriteObjectRequest and BidiWriteObjectResponse/WriteObjectResponse are largely the same (bidi has two extra fields) so rather than overloading toStorageException yet again, this time there is a single method for grpc messages that internally can branch when formatting the request. In this case, we're quite safe taking this relaxed typing because it is an internal API that is only called from a grpc context where protos will be used.

* Update GapicBidiUnbufferedWritableByteChannel to correctly handle partial consumption of content
    
  Tests passing now.
    
  Separate tracking of client detected errors and stream errors.
    
  When await is invoked, if a client detected error is present AND a stream error is present, the client detected error will take precedence. If a stream error happens and the client detected error has not yet been observed, the stream error will be added as a suppressed exception to the client detected error.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ResumableSessionFailureScenario.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ResumableSessionFailureScenario.java
@@ -26,11 +26,11 @@ import com.google.api.gax.rpc.ApiException;
 import com.google.cloud.BaseServiceException;
 import com.google.cloud.storage.StorageException.IOExceptionCallable;
 import com.google.common.io.CharStreams;
-import com.google.protobuf.MessageOrBuilder;
+import com.google.protobuf.Message;
+import com.google.storage.v2.BidiWriteObjectRequest;
 import com.google.storage.v2.ChecksummedData;
 import com.google.storage.v2.ObjectChecksums;
 import com.google.storage.v2.WriteObjectRequest;
-import com.google.storage.v2.WriteObjectResponse;
 import io.grpc.StatusRuntimeException;
 import java.io.IOException;
 import java.io.InputStreamReader;
@@ -81,10 +81,6 @@ enum ResumableSessionFailureScenario {
   private static final String PREFIX_I = "\t|< ";
   private static final String PREFIX_O = "\t|> ";
   private static final String PREFIX_X = "\t|  ";
-  // define some constants for tab widths that are more compressed that the literals
-  private static final String T1 = "\t";
-  private static final String T2 = "\t\t";
-  private static final String T3 = "\t\t\t";
 
   private static final Predicate<String> includedHeaders =
       matches("Content-Length")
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/ResumableSessionFailureScenario.java
```diff
@@ -359,16 +331,122 @@ private static String headerValueToString(Object o) {
...
+    }
+    if (req.hasObjectChecksums()) {
+      ObjectChecksums objectChecksums = req.getObjectChecksums();
+      sb.append("\n").append(prefix).append(t2).append("object_checksums: ").append("{");
+      fmt(objectChecksums, prefix, t3, sb);
+      sb.append("\n").append(prefix).append(t2).append("}");
+    }
+  }
+
+  enum Indentation {
```
JesseLovelace, line 429: 2 small things:

1) Should this maybe go in Utils.java? I know it's not used anywhere else yet, but I could see it happening
2) I think it doesn't matter in this case, but would there be a benefit to using StringEnumValue instead?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITGapicBidiUnbufferedWritableByteChannelTest.java
```diff
@@ -0,0 +1,938 @@
...
+      new ChunkSegmenter(Hasher.noop(), ByteStringStrategy.copy(), _256KiB, _256KiB);
+
+  /**
+   *
+   *
+   * <h4>S.1</h4>
+   *
+   * Attempting to append to a session which has already been finalized should raise an error
+   *
+   * <table>
```
JesseLovelace, line 68: Does javadoc from the integration tests actually get published as markdown anywhere? (It's probably still helpful even if not so you can see it in intellij)

JesseLovelace (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/ResumableSessionFailureScenario.java
BenWhitehead: 1. I think it's easy enough to move it to Utils if/when the time comes that something else might want to use this.
2. StringEnumValue is beneficial when there is an external input that needs to be translated and handled. In this case, we're defining the errors and never externalizing the enum (simply using it for convenience) or reversing a string value into an enum value.

PR #2588, opened 2024-06-14
fix: add a workaround to make sure grpc clients' hosts always match their universe domain

JesseLovelace:
Workaround in place until https://github.com/googleapis/sdk-platform-java/issues/2911 is fixed

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
@@ -600,7 +600,12 @@ public final class GrpcStorageOptions extends StorageOptions
     @BetaApi
     @Override
     public GrpcStorageOptions build() {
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -600,7 +600,12 @@ public GrpcStorageOptions.Builder setBlobWriteSessionConfig(
     @BetaApi
     @Override
     public GrpcStorageOptions build() {
-      return new GrpcStorageOptions(this, defaults());
+      GrpcStorageOptions options = new GrpcStorageOptions(this, defaults());
+      if (options.getUniverseDomain() != null) {
```
BenWhitehead, line 604: Can we add a todo-link to the issue to be fixed in sdk-platform-java? Merge away afterward.

BenWhitehead (approved):

PR #2595, opened 2024-06-24
docs: Update DeleteObject Sample to be clearer on object versioning behavior

sydney-munro:


samples/snippets/src/main/java/com/example/storage/object/DeleteObject.java
```diff
@@ -38,16 +38,13 @@ public static void deleteObject(String projectId, String bucketName, String obje
       System.out.println("The object " + objectName + " wasn't found in " + bucketName);
       return;
     }
+    // Permanently deletes a blob, even if object versioning is enabled.
+    // If object versioning is enabled you can use
+    // storage.delete(BlobId.of(bucketName, objectName));
+    // to mark an object as non-current
+    storage.delete(blob.getBlobId());
```
BenWhitehead, line 45: What do you think about something like this instead?
```suggestion
    BlobId idWithGeneration = blob.getBlobId();
    // Deletes the blob specified by its id. When the generation is present and non-null it will be
    // specified in the request.
    // If versioning is enabled on the bucket and the generation is present, only that version of
    // the object will be deleted.
    // If instead you want to delete the current version, the generation should be dropped by
    // performing the following before calling delete.
    // BlobId idWithoutGeneration =
    //    BlobId.of(idWithGeneration.getBucket(), idWithGeneration.getName());
    storage.delete(idWithGeneration);

```

sydney-munro: happy to work with that. 👍

BenWhitehead (approved):
LGTM, getting another person to read over it would be good

samples/snippets/src/main/java/com/example/storage/object/DeleteObject.java
```diff
@@ -38,16 +39,18 @@ public static void deleteObject(String projectId, String bucketName, String obje
...
-    System.out.println("Object " + objectName + " was deleted from " + bucketName);
+    BlobId idWithGeneration = blob.getBlobId();
+    // Deletes the blob specified by its id. When the generation is present and non-null it will be
+    // specified in the request.
+    // If versioning is enabled on the bucket and the generation is present, only that version of
+    // the object will be deleted.
+    // If instead you want to delete the current version, the generation should be dropped by
+    // performing the following before calling delete.
+    // BlobId idWithoutGeneration =
+    //    BlobId.of(idWithGeneration.getBucket(), idWithGeneration.getName());
```
JesseLovelace, line 50: Should we specify that in this case, `storage.delete(idWithoutGeneration)` would need to be called instead of `storage.delete(idWithGeneration)`?

samples/snippets/src/main/java/com/example/storage/object/DeleteObject.java
```diff
@@ -38,16 +39,18 @@ public static void deleteObject(String projectId, String bucketName, String obje
...
-    // conditions and data corruptions. The request to upload returns a 412 error if
-    // the object's generation number does not match your precondition.
-    Storage.BlobSourceOption precondition =
-        Storage.BlobSourceOption.generationMatch(blob.getGeneration());
-
-    storage.delete(bucketName, objectName, precondition);
-
-    System.out.println("Object " + objectName + " was deleted from " + bucketName);
+    BlobId idWithGeneration = blob.getBlobId();
+    // Deletes the blob specified by its id. When the generation is present and non-null it will be
```
JesseLovelace, line 43: I think "when the generation is present" is a little vague, maybe we should specify as "When the generation is present in the BlobId object"?

sydney-munro: rephrased the comments. let me know if that is clearer.

samples/snippets/src/main/java/com/example/storage/object/DeleteObject.java
```diff
@@ -38,16 +39,19 @@ public static void deleteObject(String projectId, String bucketName, String obje
...
-    // Optional: set a generation-match precondition to avoid potential race
-    // conditions and data corruptions. The request to upload returns a 412 error if
-    // the object's generation number does not match your precondition.
-    Storage.BlobSourceOption precondition =
-        Storage.BlobSourceOption.generationMatch(blob.getGeneration());
-
-    storage.delete(bucketName, objectName, precondition);
-
-    System.out.println("Object " + objectName + " was deleted from " + bucketName);
+    BlobId idWithGeneration = blob.getBlobId();
```
JesseLovelace, line 42: maybe add a comment after this line like "// The BlobId object has the Generation specified"

JesseLovelace (approved):

samples/snippets/src/main/java/com/example/storage/object/DeleteObject.java
sydney-munro: i think the variable name is clear on this, no?

PR #2613, opened 2024-07-01
test: Ignore failing DirectPath tests

sydney-munro:

### DIFF
--- a/samples/snippets/src/test/java/com/example/storage/QuickstartSampleIT.java
+++ b/samples/snippets/src/test/java/com/example/storage/QuickstartSampleIT.java
@@ -26,6 +26,7 @@ import com.google.storage.control.v2.StorageLayoutName;
 import java.util.UUID;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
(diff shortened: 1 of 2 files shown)


samples/snippets/src/test/java/com/example/storage/QuickstartSampleIT.java
```diff
@@ -69,6 +70,7 @@ public void testQuickstartGrpc() throws Exception {
     assertThat(got).contains(String.format("Bucket %s created.", bucketName));
   }
 
+  @Ignore("https://github.com/googleapis/java-storage/issues/2612")
```
BenWhitehead, line 73: Can we add a comment into the linked issue of when we can turn it back on? The issues doesn't include any note to explain why they would need to be turned back on.

BenWhitehead (approved):

samples/snippets/src/test/java/com/example/storage/QuickstartSampleIT.java
sydney-munro: yeah can do

PR #2618, opened 2024-07-02
chore: surround string with special characters with quotes in generation_config

diegomarquezp:


diegomarquezp:
```
  File "/usr/local/lib/python3.11/site-packages/yaml/scanner.py", line 577, in fetch_value
    raise ScannerError(None, None,
yaml.scanner.ScannerError: mapping values are not allowed here
  in "/workspace/baseline_generation_config.yaml", line 9, column 157
```

This error is expected since it has issues parsing the **baseline** config yaml, which is the one in `main` with the problematic string without quotes

JoeWang1127 (approved):

PR #2625, opened 2024-07-02
chore: update hermetic_library_generation.yaml to only run on PRs that modify generation_config.yaml

BenWhitehead:


sydney-munro (approved):

diegomarquezp (changes requested):
The underlying script already does this check. We need the workflow to run when the _last commit_ has affected `generation_config.yaml`. Otherwise we would have a loop (generation_config -> source code + generation_config -> source code + generation_config)

https://github.com/googleapis/java-storage/blob/a038b69e318d572e2dbf99aec5b080c2d936aa10/.github/scripts/hermetic_library_generation.sh#L72-L76

We may expand this check to allow other files being modified, such as the OwlBot yaml, or even some more sophisticated logic.

cc: @JoeWang1127

BenWhitehead:
I acknowledge that the script detects this, but github will also fail if the action is present on a branch that does not have a token. My preference would be that the workflow for the generation script not even attempt to access the token until a "write step", but the current script is all or nothing. Filtering on input change seems like an okay compromise. 

image

RE:
> run when the last commit has affected generation_config.yaml

Could this be a check in the script `$COMMITISH == HEAD` if it matches proceed, otherwise exit as not applicable.

sydney-munro (approved):

diegomarquezp:
@BenWhitehead, we have disabled the check on PRs that come from fork (i.e. repos that don't have the token).

https://github.com/googleapis/java-storage/blob/2f800d09d128abdb3fb6b990ce4d2ed24cdd45cf/.github/workflows/hermetic_library_generation.yaml#L23

For all other PRs, the token should be accessible since it's a [repo-level secret](https://github.com/googleapis/java-storage/settings/secrets/actions).

Regarding the screenshot you shared, would you mind sharing the link to it to see which PR it comes from? I suspect it's either a workflow run from before we set the token or from a fork. 

cc: @blakeli0 @JoeWang1127

diegomarquezp (approved):

BenWhitehead:
Spoke offline to get consensus.

Issue #2612, opened 2024-07-01
Turn DirectPath Sample test back on


sydney-munro:

sydney-munro:
turn test back on when internal bug b/348378825 is resolved

PR #2645, opened 2024-07-24
test: Turn DirectPath quickstart test back on and fix abandoned storage instance in GRPC test

sydney-munro:
Closing instance of storage that gets abandoned and then eventually cleaned up which results in a long error message that pollutes the logs

Fixes issue #2612

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
@@ -258,5 +258,6 @@ public final class ITGrpcTest {
         () ->
             assertThat(
                 storage.getOptions().getHost().equals("https://storage.my-universe-domain.com")));
+    storage.close();
   }
 }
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITGrpcTest.java
```diff
@@ -258,5 +258,6 @@ public void testGrpcUniverseDomainMatchesHost() throws Exception {
         () ->
             assertThat(
                 storage.getOptions().getHost().equals("https://storage.my-universe-domain.com")));
+    storage.close();
```
BenWhitehead, line 261: I think this test can be changed to stop at the options piece. Its assertion the host from the options in the storage instance.

```
  @Test
  public void testGrpcUniverseDomainMatchesHost() throws Exception {
    StorageOptions options =
        StorageOptions.grpc().setUniverseDomain("my-universe-domain.com").build();
    assertAll(
        () -> assertThat(options.getUniverseDomain().equals("my-universe-domain.com")),
        () -> assertThat(options.getHost().equals("https://storage.my-universe-domain.com")));
  }
```

Also, if one of the assertions failed, we wouldn't make it to this close. If we were keeping the storage instance around, we should wrap it in a try-with-resource to auto close it.

BenWhitehead (approved):

PR #2679, opened 2024-08-20
fix: Update TransferManagerUtils#createBlobName to only use the file name instead of the full file path (#2573)

### LINKED ISSUE #2573: Support StripPrefix on upload for TransferManager
Currently transfer manager supports stripping a customer supplied prefix on an object on object download, the same functionality is not supported on upload and the name of the object maintains the full File name as the object name.

chalmagr:
- [x] Code coverage does not decrease (if any source code was changed)

Fixes #2573 

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerUtils.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerUtils.java
@@ -28,9 +28,9 @@ final class TransferManagerUtils {
 
   static String createBlobName(ParallelUploadConfig config, Path file) {
     if (config.getPrefix().isEmpty()) {
-      return file.toString();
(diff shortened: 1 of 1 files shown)


sydney-munro:
Thanks for your contribution, unfortunately this would introduce a breaking change into an already GA feature in our client. We cannot accept this PR.

chalmagr:
> Thanks for your contribution, unfortunately this would introduce a breaking change into an already GA feature in our client. We cannot accept this PR.

Hi Sydney,

I don't think that this GA feature is of any use the way it works today. Whoever may be using this, they would need to expect that the destination files will be located mimicking the source directory structure (see my comment on [#2573](https://github.com/googleapis/java-storage/issues/2573#issuecomment-2299334702))

We could add some flag to enable this behavior, but I believe that this feature is rendered unusable because of this issue

PR #2696, opened 2024-09-09
fix: close pending zero-copy responses when Storage#close is called

BenWhitehead:
Update gRPC based Storage instances to close out in progress reads when zero-copy is used and Storage#close() is called.


sydney-munro (approved):

BenWhitehead:
integration test failures being investigated in https://github.com/googleapis/java-storage/pull/2697

PR #2708, opened 2024-09-13
fix: update grpc based ReadObject rpcs to remove race condition between cancellation and message handling

BenWhitehead:
Update GapicUnbufferedReadableByteChannel to manage the grpc stream itself rather than using the stream iterator provided by gax. This allows us to ensure the cancellation is observed and our draining performs before returning from close().

As a side effect of not using the gax stream iterator, we now must handle stream restarts ourselves. GrpcStorageOptions.ReadObjectResumptionStrategy has been removed entirely, while RetryingDependencies and ResultRetryAlgorithm are now plumbed all the way down to the GapicUnbufferedReadableByteChannel.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
@@ -150,10 +150,6 @@ final class GapicUnbufferedReadableByteChannel
                   "Mismatch Generation between subsequent reads. Expected %d but received %d",
                   metadata.getGeneration(), respMetadata.getGeneration()));
         }
-
-        if (!result.isDone()) {
-          result.set(metadata);
-        }
       }
       ChecksummedData checksummedData = resp.getChecksummedData();
       ByteString content = checksummedData.getContent();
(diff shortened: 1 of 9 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GapicUnbufferedReadableByteChannel.java
```diff
@@ -153,98 +179,108 @@ public void close() throws IOException {
...
     close();
     StorageException cause =
         new StorageException(HttpStatusCodes.STATUS_CODE_PRECONDITION_FAILED, message);
     throw new IOException(message, cause);
   }
 
-  private final class LazyServerStreamIterator implements Iterator<ReadObjectResponse>, Closeable {
-    private ServerStream<ReadObjectResponse> serverStream;
-    private Iterator<ReadObjectResponse> responseIterator;
+  private final class ReadObjectObserver extends StateCheckingResponseObserver<ReadObjectResponse> {
```
BenWhitehead, line 243: Unfortunately, git thinks this class is a modification of the previous LazyServerStreamIterator except it's a whole new class.

Rather than comparing this class to what was there before, evaluate this class as brand new.

sydney-munro (approved):

PR #2735, opened 2024-09-25
chore: use composite action to generate library

JoeWang1127:
In this PR:
- Use composite action defined in sdk-platform-java to generate library, to reduce code duplication.
- Change renovate to update the action version.


.github/workflows/hermetic_library_generation.yaml
```diff
@@ -16,35 +16,30 @@
 name: Hermetic library generation upon generation config change through pull requests
 on:
   pull_request:
-    paths:
-      - 'generation_config.yaml'
-
```
BenWhitehead, line 21: Please keep these lines

BenWhitehead (changes requested):

.github/workflows/hermetic_library_generation.yaml
JoeWang1127: Done.

BenWhitehead (approved):

PR #2754, opened 2024-10-07
docs: add samples for soft delete (objects)

JesseLovelace:
Samples for working with soft deleted objects, intended for https://cloud.google.com/storage/docs/use-soft-delete and https://cloud.google.com/storage/docs/use-soft-deleted-objects

### DIFF
--- a/samples/snippets/src/main/java/com/example/storage/bucket/DisableSoftDelete.java
+++ b/samples/snippets/src/main/java/com/example/storage/bucket/DisableSoftDelete.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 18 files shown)


samples/snippets/src/main/java/com/example/storage/bucket/DisableSoftDelete.java
```diff
@@ -0,0 +1,44 @@
...
+public class DisableSoftDelete {
+    public static void disableSoftDelete(String projectId, String bucketName) {
+        // The ID of your GCP project
+        // String projectId = "your-project-id";
+
+        // The ID of your GCS bucket
+        // String bucketName = "your-unique-bucket-name";
+
+        Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+        storage.get(bucketName).toBuilder().setSoftDeletePolicy(
```
sydney-munro, line 35: nit: could we separate this into two calls to make it more readable? the get call and then the update.

samples/snippets/src/main/java/com/example/storage/bucket/SetSoftDeletePolicy.java
```diff
@@ -0,0 +1,43 @@
...
+public class SetSoftDeletePolicy {
+    public static void setSoftDeletePolicy(String projectId, String bucketName) {
+        // The ID of your GCP project
+        // String projectId = "your-project-id";
+
+        // The ID of your GCS bucket
+        // String bucketName = "your-unique-bucket-name";
+
+        Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+        storage.get(bucketName).toBuilder().setSoftDeletePolicy(
```
sydney-munro, line 35: same nit, I think readability would be improved if this were two separate calls.

sydney-munro (approved):
approved with 2 small comments.

sydney-munro:
> approved with 2 small comments.

LGTM

PR #2779, opened 2024-10-15
chore: Use libraries_bom_version and library_version from parent READ…

blakeli0:


blakeli0:
Seems like `.readme-partial` does not have access to metadata info from the parent README template, closing this PR.

PR #2783, opened 2024-10-18
feat: Add OpenTelemetry Traces to GRPC

sydney-munro:
To keep this more closely aligned with the introduction in HTTP the scope here is small.

To follow will be to add the attributes for recording exceptions and testing that flow.

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -162,6 +162,22 @@ branchProtectionRules:
       - 'Kokoro - Test: Java GraalVM Native Image'
       - 'Kokoro - Test: Java 17 GraalVM Native Image'
       - javadoc
+  - pattern: otel-v1-branch
+    isAdminEnforced: true
+    requiredApprovingReviewCount: 1
+    requiresCodeOwnerReviews: true
+    requiresStrictStatusChecks: false
+    requiredStatusCheckContexts:
+    - dependencies (17)
+    - lint
+    - clirr
+    - units (8)
+    - units (11)
+    - 'Kokoro - Test: Integration'
+    - cla/google
+    - 'Kokoro - Test: Java GraalVM Native Image'
+    - 'Kokoro - Test: Java 17 GraalVM Native Image'
+    - javadoc
 permissionRules:
   - team: yoshi-admins
     permission: admin
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
@@ -642,7 +642,7 @@ public final class GrpcStorageOptions extends StorageOptions
     /**
      * Enable OpenTelemetry Tracing and provide an instance for the client to use.
      *
-     * @param openTelemetrySdk
+     * @param openTelemetrySdk User defined instance of OpenTelemetry SDK to be used by the library
      */
     public GrpcStorageOptions.Builder setOpenTelemetrySdk(OpenTelemetrySdk openTelemetrySdk) {
       requireNonNull(openTelemetrySdk, "openTelemetry must be non null");
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
@@ -119,6 +119,7 @@ public final class GrpcStorageOptions extends StorageOptions
   private final boolean grpcClientMetricsManuallyEnabled;
   private final GrpcInterceptorProvider grpcInterceptorProvider;
   private final BlobWriteSessionConfig blobWriteSessionConfig;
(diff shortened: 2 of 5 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -634,9 +639,15 @@ public GrpcStorageOptions.Builder setBlobWriteSessionConfig(
       return this;
     }
 
-    @Override
-    public StorageOptions.Builder setOpenTelemetrySdk(@NonNull OpenTelemetrySdk openTelemetrySdk) {
-      return null;
+    /**
+     * Enable OpenTelemetry Tracing and provide an instance for the client to use.
+     *
+     * @param openTelemetrySdk
```
JesseLovelace, line 645: I think we need a description of the param here, or just delete the @param line

JesseLovelace (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
sydney-munro: done. added a description.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITOpenTelemetryTest.java
```diff
@@ -58,29 +48,80 @@ public void checkInstrumentation() {
...
+    String grpcBucket = randomBucketName();
+    // NoOp for HTTP
+    StorageOptions storageOptionsHttp = StorageOptions.http().build();
+    Storage storageHttp = storageOptionsHttp.getService();
+    storageHttp.create(BucketInfo.of(httpBucket));
+
+    // NoOp for Grpc
+    StorageOptions storageOptionsGrpc = StorageOptions.grpc().build();
+    Storage storageGrpc = storageOptionsGrpc.getService();
+    storageGrpc.create(BucketInfo.of(grpcBucket));
```
frankyn, line 107: What is no-op testing and checking?

.github/sync-repo-settings.yaml
```diff
@@ -162,6 +162,22 @@ branchProtectionRules:
...
+    requiredApprovingReviewCount: 1
+    requiresCodeOwnerReviews: true
+    requiresStrictStatusChecks: false
+    requiredStatusCheckContexts:
+    - dependencies (17)
+    - lint
+    - clirr
+    - units (8)
+    - units (11)
+    - 'Kokoro - Test: Integration'
```
frankyn, line 176: Does Kokoro need to be updated internally to watch this branch as well?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITOpenTelemetryTest.java
```diff
@@ -58,29 +48,80 @@ public void checkInstrumentation() {
...
+                    .build())
+            .build();
+    StorageOptions storageOptions =
+        StorageOptions.grpc().setOpenTelemetrySdk(openTelemetrySdk).build();
     Storage storage = storageOptions.getService();
-    storage.create(BucketInfo.of(generator.randomBucketName()));
+    String bucket = randomBucketName();
+    storage.create(BucketInfo.of(bucket));
+    TestExporter testExported = (TestExporter) exporter;
+    SpanData spanData = testExported.getExportedSpans().get(0);
```
frankyn, line 83: Do you plan on testing which spans are collected per operation in addition to Span attributes? I think this would be more useful for Upload and Download ops.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITOpenTelemetryTest.java
sydney-munro: just checking to make sure that it still runs without issue and does nothing. basically just that it doesnt error out.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITOpenTelemetryTest.java
sydney-munro: I don't know if i understand this question. what do you mean by which spans are collected

.github/sync-repo-settings.yaml
sydney-munro: ill double check!

google-cloud-storage/src/test/java/com/google/cloud/storage/ITOpenTelemetryTest.java
sydney-munro: ah I see, I haven't expanded this for that case yet but yeah we would want to test each span is generated. Maybe we could check for the size to make sure if we expect 3 writeObjectRequests we would have 3 spans

frankyn: It's a good start to assert span count; and expand from there.

PR #2792, opened 2024-10-24
wip create(Blob) instrumentation

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
@@ -248,8 +248,23 @@ final class GrpcStorageImpl extends BaseService<StorageOptions>
   public Blob create(
       BlobInfo blobInfo, byte[] content, int offset, int length, BlobTargetOption... options) {
     Opts<ObjectTargetOpt> opts = Opts.unwrap(options).resolveFrom(blobInfo);
-    return internalDirectUpload(blobInfo, opts, ByteBuffer.wrap(content, offset, length))
-        .asBlob(this);
+    // Start the otel span to retain information of the origin of the request
+    OpenTelemetryTraceUtil.Span otelSpan =
+        openTelemetryTraceUtil.startSpan("create(BlobInfo, BlobTargetOption");
+    try (OpenTelemetryTraceUtil.Scope unused = otelSpan.makeCurrent()) {
+      return internalDirectUpload(
+              blobInfo,
+              opts,
+              ByteBuffer.wrap(content, offset, length),
+              openTelemetryTraceUtil.currentContext())
+          .asBlob(this);
+    } catch (Exception e) {
+      otelSpan.recordException(e);
+      otelSpan.setStatus(io.opentelemetry.api.trace.StatusCode.ERROR, e.getClass().getSimpleName());
+      throw StorageException.coalesce(e);
+    } finally {
+      otelSpan.end();
+    }
   }
 
   @Override
@@ -799,9 +814,14 @@ final class GrpcStorageImpl extends BaseService<StorageOptions>
 
   @Override
   public BlobInfo internalDirectUpload(
-      BlobInfo blobInfo, Opts<ObjectTargetOpt> opts, ByteBuffer buf) {
+      BlobInfo blobInfo,
+      Opts<ObjectTargetOpt> opts,
+      ByteBuffer buf,
+      OpenTelemetryTraceUtil.Context ctx) {
     requireNonNull(blobInfo, "blobInfo must be non null");
     requireNonNull(buf, "content must be non null");
+    OpenTelemetryTraceUtil.Span otelSpan =
+        openTelemetryTraceUtil.startSpan("internalDirectUpload(BlobInfo)", ctx);
     Opts<ObjectTargetOpt> optsWithDefaults = opts.prepend(defaultOpts);
     GrpcCallContext grpcCallContext =
         optsWithDefaults.grpcMetadataMapper().apply(GrpcCallContext.createDefault());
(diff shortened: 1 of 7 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -248,8 +248,23 @@ public Blob create(BlobInfo blobInfo, byte[] content, BlobTargetOption... option
...
-        .asBlob(this);
+    // Start the otel span to retain information of the origin of the request
+    OpenTelemetryTraceUtil.Span otelSpan =
+        openTelemetryTraceUtil.startSpan("create(BlobInfo, BlobTargetOption");
+    try (OpenTelemetryTraceUtil.Scope unused = otelSpan.makeCurrent()) {
+      return internalDirectUpload(
+              blobInfo,
+              opts,
+              ByteBuffer.wrap(content, offset, length),
+              openTelemetryTraceUtil.currentContext())
```
frankyn, line 259: What does currentContext() solve in this instrumentation?

frankyn (commented):
still reviewing adding one question; I see it's WIP; are you ready for review?

sydney-munro:
> still reviewing adding one question; I see it's WIP; are you ready for review?

not yet just testing some stuff out as the testbench integration is a pain to sort out locally.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
sydney-munro: it is used to link this span to the child spans in internalDirectUpload

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -799,38 +814,51 @@ public GrpcBlobWriteChannel writer(BlobInfo blobInfo, BlobWriteOption... options
...
   public BlobInfo internalDirectUpload(
-      BlobInfo blobInfo, Opts<ObjectTargetOpt> opts, ByteBuffer buf) {
+      BlobInfo blobInfo,
+      Opts<ObjectTargetOpt> opts,
+      ByteBuffer buf,
+      OpenTelemetryTraceUtil.Context ctx) {
     requireNonNull(blobInfo, "blobInfo must be non null");
     requireNonNull(buf, "content must be non null");
+    OpenTelemetryTraceUtil.Span otelSpan =
+        openTelemetryTraceUtil.startSpan("internalDirectUpload(BlobInfo)", ctx);
```
sydney-munro, line 824: @frankyn we pass the parent span here when we start the child span.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
sydney-munro: since internalDirectUpload is used here and also in PCU I want the parent linked so we know where the original call came from.

frankyn: Gotcha; that makes sense. I don't think the automatic context propagation will work in the case of PCU

I was referring to: https://javadoc.io/doc/io.opentelemetry/opentelemetry-api/1.1.0/io/opentelemetry/api/trace/Tracer.html

> The automatic context propagation is done using Context which is a gRPC independent implementation for in-process Context propagation mechanism which can carry scoped-values across API boundaries and between threads. Users of the library must propagate the Context between different threads.

The last sentence is why you would need to do what you're doing now. Good call!

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -248,8 +248,23 @@ public Blob create(BlobInfo blobInfo, byte[] content, BlobTargetOption... option
   public Blob create(
       BlobInfo blobInfo, byte[] content, int offset, int length, BlobTargetOption... options) {
     Opts<ObjectTargetOpt> opts = Opts.unwrap(options).resolveFrom(blobInfo);
-    return internalDirectUpload(blobInfo, opts, ByteBuffer.wrap(content, offset, length))
-        .asBlob(this);
+    // Start the otel span to retain information of the origin of the request
+    OpenTelemetryTraceUtil.Span otelSpan =
+        openTelemetryTraceUtil.startSpan("create(BlobInfo, BlobTargetOption");
```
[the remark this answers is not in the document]
sydney-munro: helpful! I will remove the parameters.

google-cloud-storage/src/test/java/com/google/cloud/storage/ParallelCompositeUploadWritableByteChannelTest.java
```diff
@@ -356,9 +357,12 @@ public void partsRetainMetadata() throws Exception {
...
               @Override
               public BlobInfo internalDirectUpload(
-                  BlobInfo info, Opts<ObjectTargetOpt> opts, ByteBuffer buf) {
+                  BlobInfo info,
+                  Opts<ObjectTargetOpt> opts,
+                  ByteBuffer buf,
+                  OpenTelemetryTraceUtil.Context ctx) {
                 metadatas.add(info.getMetadata());
-                return super.internalDirectUpload(info, opts, buf);
+                return super.internalDirectUpload(info, opts, buf, null);
```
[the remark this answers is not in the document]
sydney-munro: thats a good idea.

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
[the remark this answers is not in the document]
sydney-munro: how do we want to differentiate create(Bucket) and create(Blob)?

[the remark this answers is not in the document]
frankyn: we could identify either by attribute uri metdata but an alternative would be to supply the full method name with all parameters in Java as part of attribute information for `rpc.method`

PR #2806, opened 2024-11-01
feat: adds bucket restore support

JesseLovelace:
Adds support for bucket restore, bucket generation, hard delete time, soft delete time, listing/getting soft deleted buckets

gRPC not yet available, adding DNM tag until it is


ShreyasSinha:
Looks like a stale PR, no activity for > 1 year. closing.
Feel free to re-open if required.

PR #2808, opened 2024-11-04
feat: Instrument HTTP Reads and Rewrites

sydney-munro:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
@@ -889,7 +889,7 @@ public class HttpStorageRpc implements StorageRpc {
     OpenTelemetryTraceUtil.Span otelSpan = openTelemetryTraceUtil.startSpan("read");
     Span span = startSpan(HttpStorageRpcSpans.SPAN_NAME_READ);
     Scope scope = tracer.withSpan(span);
-    try (OpenTelemetryTraceUtil.Scope unused = otelSpan.makeCurrent()){
+    try (OpenTelemetryTraceUtil.Scope unused = otelSpan.makeCurrent()) {
       checkArgument(position >= 0, "Position should be non-negative, is " + position);
       Get req = createReadRequest(from, options);
       Boolean shouldReturnRawInputStream = Option.RETURN_RAW_INPUT_STREAM.getBoolean(options);
@@ -1183,8 +1183,11 @@ public class HttpStorageRpc implements StorageRpc {
     OpenTelemetryTraceUtil.Span otelSpan = openTelemetryTraceUtil.startSpan("continueRewrite");
     Span span = startSpan(HttpStorageRpcSpans.SPAN_NAME_CONTINUE_REWRITE);
     Scope scope = tracer.withSpan(span);
-    try (OpenTelemetryTraceUtil.Scope unused = otelSpan.makeCurrent()){
-      return rewrite(previousResponse.rewriteRequest, previousResponse.rewriteToken, openTelemetryTraceUtil.currentContext());
+    try (OpenTelemetryTraceUtil.Scope unused = otelSpan.makeCurrent()) {
+      return rewrite(
+          previousResponse.rewriteRequest,
+          previousResponse.rewriteToken,
+          openTelemetryTraceUtil.currentContext());
     } finally {
       otelSpan.end();
       scope.close();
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -848,9 +848,10 @@ private Get createReadRequest(StorageObject from, Map<Option, ?> options) throws
   @Override
   public long read(
       StorageObject from, Map<Option, ?> options, long position, OutputStream outputStream) {
+    OpenTelemetryTraceUtil.Span otelSpan = openTelemetryTraceUtil.startSpan("read");
```
frankyn, line 851: I misunderstood the span name; Does startSpan provide context on class for this method?

Pattern: `$package.$service/$method`
Example using C++: `storage::Client::WriteObject/CreateResumableUpload`

google-cloud-storage/src/test/java/com/google/cloud/storage/ITHttpOpenTelemetryTest.java
```diff
@@ -97,6 +109,62 @@ public void runCreateBlob() {
...
+  public void runRead() throws IOException {
+    BlobInfo blobInfo =
+        BlobInfo.newBuilder(blobId).setContentEncoding("gzip").setContentType("text/plain").build();
+    storage.create(blobInfo, helloWorldGzipBytes);
+    Path helloWorldTxtGz = File.createTempFile(blobId.getName(), ".txt.gz").toPath();
+    storage.downloadTo(
+        blobId, helloWorldTxtGz, Storage.BlobSourceOption.shouldReturnRawInputStream(true));
+    TestExporter testExported = (TestExporter) exporter;
+    List<SpanData> spanData = testExported.getExportedSpans();
+    for (SpanData span : spanData) {
```
frankyn, line 122: `spanData` can be de-duped as a helper that validates attributes

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
sydney-munro: yes https://github.com/googleapis/java-storage/blob/otel-v1-branch/google-cloud-storage/src/main/java/com/google/cloud/storage/otel/OpenTelemetryInstance.java#L156

google-cloud-storage/src/test/java/com/google/cloud/storage/ITHttpOpenTelemetryTest.java
sydney-munro: what do you mean by this?

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
frankyn: Oh, I was thinking it would be 

`storage.spi.v1/read` in this case so it's clear where this span exists. 

@cojenco could you take a look at this?

google-cloud-storage/src/test/java/com/google/cloud/storage/ITHttpOpenTelemetryTest.java
frankyn:
```
List<SpanData> spanData = testExported.getExportedSpans();
    for (SpanData span : spanData) {
      Assert.assertEquals("Storage", getAttributeValue(span, "gcp.client.service"));
      Assert.assertEquals("googleapis/java-storage", getAttributeValue(span, "gcp.client.repo"));
      Assert.assertEquals(
          "com.google.cloud.google-cloud-storage", getAttributeValue(span, "gcp.client.artifact"));
      Assert.assertEquals("http", getAttributeValue(span, "rpc.system"));
    }
    ```

Is duplicated for runRead() and runCopy(); thinking a helper method can be used to reduce duplication of the same asserts.

frankyn (approved):
We can always update span names after.

google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
cojenco: +1 to providing context on the class/module for this method, especially considering there are many call paths in java

The OTel semantic conventions recommends the pattern of `$package.$service/$method` where $method MUST NOT contain slashes. I'm not very familiar with how the classes are structured here, but I agree that it would be a good idea to include some context of spi/v1/HttpStorageRpc in the span name for better debuggability

PR #2809, opened 2024-11-05
fix: set default values for monitored resource

frankyn:
b/377466694

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/OpenTelemetryBootstrappingUtils.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/OpenTelemetryBootstrappingUtils.java
@@ -31,6 +31,7 @@ import io.grpc.ManagedChannelBuilder;
 import io.grpc.opentelemetry.GrpcOpenTelemetry;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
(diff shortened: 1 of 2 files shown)


sydney-munro (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/OpenTelemetryBootstrappingUtils.java
```diff
@@ -196,24 +197,36 @@ static SdkMeterProvider createMeterProvider(
         shouldSuppressExceptions
             ? new PermissionDeniedSingleReportMetricsExporter(cloudMonitoringExporter)
             : cloudMonitoringExporter;
+    AttributesBuilder attributesBuilder =
+        Attributes.builder()
+            .put("gcp.resource_type", "storage.googleapis.com/Client")
+            .put("project_id", projectIdToUse)
+            .put("instance_id", UUID.randomUUID().toString())
+            .put("api", "grpc");
```
sydney-munro, line 205: nit: could we use something to detect grpc vs http even if this isn't set up for http yet? would make this reusable in the future.

frankyn: AFAIU, resource detector is platform dependent on https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md#gcp-metadata

API type would need to be captured client side / added when introduced into the java storage client.

PR #2815, opened 2024-11-12
feat: Instrument ReadAllBytes and CreateFrom in gRPC

sydney-munro:


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageImpl.java
```diff
@@ -372,37 +388,46 @@ public Blob createFrom(BlobInfo blobInfo, InputStream content, BlobWriteOption..
   public Blob createFrom(
       BlobInfo blobInfo, InputStream in, int bufferSize, BlobWriteOption... options)
       throws IOException {
-    requireNonNull(blobInfo, "blobInfo must be non null");
+    OpenTelemetryTraceUtil.Span otelSpan =
+        openTelemetryTraceUtil.startSpan("createFrom", this.getClass().getName());
```
JesseLovelace, line 392: since `createFrom` has a few different signatures, does which signature we're working with need to be captured in the span?

sydney-munro: we are only concerned with the method not the parameters.

JesseLovelace (approved):

PR #2826, opened 2024-11-21
feat: introduce `java.time` methods and variables

diegomarquezp:
This PR introduces `java.time` alternatives to existing `org.threeten.bp.*` methods, as well as switching internal variables (if any) to `java.time`

The main constraint is to keep the changes backwards compatible, so for each existing threeten method "`method1(org.threeten.bp.Duration)`" we will add an alternative with a _Duration_ (or _Timestamp_ when applicable) suffix: "`method1Duration(java.time.Duration)`".

For most cases, the implementation will be held in the `java.time` method and the old threeten method will just delegate the call to it. However, for the case of abstract classes, the implementation will be kept in the threeten method to avoid breaking changes (i.e. users that already overloaded the method in their user code).

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
@@ -16,12 +16,15 @@
 
 package com.google.cloud.storage;
 
+import static com.google.api.gax.util.TimeConversionUtils.toJavaTimeDuration;
+import static com.google.api.gax.util.TimeConversionUtils.toThreetenDuration;
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
 import com.google.api.core.ApiClock;
 import com.google.api.core.BetaApi;
 import com.google.api.core.InternalApi;
+import com.google.api.core.ObsoleteApi;
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.core.FixedCredentialsProvider;
 import com.google.api.gax.core.NoCredentialsProvider;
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -652,9 +665,15 @@ public StorageRetryStrategy getStorageRetryStrategy() {
...
+    @ObsoleteApi("Use getTerminationAwaitDurationDuration() instead")
+    public org.threeten.bp.Duration getTerminationAwaitDuration() {
+      return toThreetenDuration(getTerminationAwaitDurationDuration());
+    }
+
     /** @since 2.14.0 */
-    public Duration getTerminationAwaitDuration() {
-      return Duration.ofMinutes(1);
+    public java.time.Duration getTerminationAwaitDurationDuration() {
+      return java.time.Duration.ofMinutes(1);
```
lqiu96, line 676: I don't think we've seen this case before (or at least I don't recall). But if the method ends with Duration then we come across `DurationDuration` and I think that's even more confusing.

Maybe we want to keep this for consistency with all other SDK methods, but I'm thinking that naming scheme isn't ideal.

Perhaps `getTerminationAwaitJavaTimeDuration()`? Open to suggestions for this.

diegomarquezp: I agree with the point in https://github.com/googleapis/java-datastore/pull/1671#discussion_r1854745739 saying that using `DurationJavaTime` keeps the prefixes consistent

diegomarquezp: to confirm this with @sydney-munro and @JesseLovelace

sydney-munro (approved):

PR #2829, opened 2024-11-27
feat: Instrument Reader and Writer methods

sydney-munro:
This PR also adds back testbench support that was previously removed for local development.


google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/TestBench.java
```diff
@@ -16,6 +16,7 @@
```
sydney-munro, line 16: This is adding back code that was removed for local development.

JesseLovelace (approved):

PR #2861, opened 2025-01-07
docs: Create OpenTelemetry Quickstart Sample

sydney-munro:


samples/snapshot/pom.xml
```diff
@@ -68,6 +68,16 @@
...
+    <dependency>
+      <groupId>com.google.cloud.opentelemetry</groupId>
+      <artifactId>exporter-trace</artifactId>
+      <version>0.31.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.opentelemetry</groupId>
+      <artifactId>propagators-gcp</artifactId>
+      <version>0.33.0-alpha</version>
+    </dependency>
```
BenWhitehead, line 80: I'm not sure we need to declare this, we're not declaring it in ssb, and it doesn't seem to be pulled in transitively.

samples/snippets/src/main/java/com/example/storage/QuickstartOpenTelemetrySample.java
```diff
@@ -0,0 +1,58 @@
...
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+
+// [START storage_enable_otel_tracing]
+public class QuickstartOpenTelemetrySample {
+  public static void main(String... args) throws Exception {
+    SpanExporter exporter = TraceExporter.createWithDefaultConfiguration();
```
BenWhitehead, line 37: nit: `s/exporter/spanExporter/g`

samples/snippets/src/main/java/com/example/storage/QuickstartOpenTelemetrySample.java
```diff
@@ -0,0 +1,58 @@
...
+        W3CTraceContextPropagator.getInstance(),
+        new XCloudTraceContextPropagator(/*oneway=*/true));
+
+    OpenTelemetrySdk openTelemetry =
+        OpenTelemetrySdk.builder()
+            .setPropagators(ContextPropagators.create(propagators))
+            .setTracerProvider(
+                SdkTracerProvider.builder()
+                    // Sample Rate set to always
+                    .setSampler(Sampler.alwaysOn())
```
BenWhitehead, line 48: We probably want a ratio based default documented here rather than `alwaysOn`. traces can get throttled.

```suggestion
                    .setSampler(Sampler.traceIdRatioBased(0.2)) // sample only 20% of trace ids
```

BenWhitehead (changes requested):

samples/snippets/src/main/java/com/example/storage/QuickstartOpenTelemetrySample.java
sydney-munro: I think we should follow suit with python and use always for the sample just to ensure people will actually produce traces no matter how small the request load for the quickstart.

samples/snapshot/pom.xml
sydney-munro: I can take this out but I will have to take out the line where we suggest the XCloudTraceContextPropagator. This isn't required so I am fine with removing it.

BenWhitehead: https://cloud.google.com/trace/docs/setup#force-trace has some detail about `X-Cloud-Trace-Context`. I think it's worth keep in the sample for most customers.

We didn't need it in ssb because ssb is the originator of all traces, and therefore doesn't need to propogate anything.

samples/snippets/src/main/java/com/example/storage/QuickstartOpenTelemetrySample.java
BenWhitehead: In that case, we probably want to swap the `traceIdRatioBased` into a comment next to the `alwaysOn()`. To try and ensure customers are shown that they should probably use sampling in their app.

cojenco: Good idea! I can change the python sample and template to use alwaysOn and comment on sampling rate

BenWhitehead (approved):

cojenco:
Hi @BenWhitehead is there anything pending for this PR? Could you help merge this in, thank you!

PR #2862, opened 2025-01-07
test: integrate otel into our StorageITRunner

BenWhitehead:
Integrate OpenTelemetry tracing into our StorageITRunner.

By default the `OpenTelementry.noop()` instance will be used (avoiding the need for the requisite permissions).

To enable trace exporting for the tests, and the storage instances provide either the system property `STORAGE_IT_OTEL_ENABLE=true` or the environment variable `STORAGE_IT_OTEL_ENABLE=true`. They are evaluated as an `OR`, meaning that if both are specified and one is true and one is false, tracing will be enabled.

An example screenshot showing the registry cleaning up after test runs:
image

### DIFF
--- a/google-cloud-storage/pom.xml
+++ b/google-cloud-storage/pom.xml
@@ -164,6 +164,12 @@
       <groupId>com.google.cloud.opentelemetry</groupId>
       <artifactId>exporter-metrics</artifactId>
     </dependency>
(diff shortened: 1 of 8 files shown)


sydney-munro (approved):
LGTM :)

PR #2876, opened 2025-01-10
test: fix HttpRpcContextTest to not attempt environment credentials resolution

BenWhitehead:


danielduhh (commented):
Whats the backstory?

danielduhh (approved):

BenWhitehead:
Someone on another team was attempting to run our unit tests, and run into failures due to the lack of credentials present in their environment but with access to the metadata service.

This test should be a unit test, but slipped through our detection so far because we haven't run our unit tests in that type of configuration.

PR #2882, opened 2025-01-14
feat: add new Storage#moveBlob method to atomically rename an object

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2778,6 +2780,150 @@ public static Builder newBuilder() {
...
+
+    public List<BlobTargetOption> getTargetOptions() {
+      return targetOptions;
+    }
+
+    public Builder toBuilder() {
+      return new Builder(source, target, sourceOptions, targetOptions);
+    }
+
+    public static Builder newBuilder() {
```
JesseLovelace, line 2827: Do we maybe want to add something like MoveBlobRequest.of(BlobIdSource, BlobId target) for convenience?

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -4882,4 +5028,15 @@ default void close() throws Exception {}
   default BlobWriteSession blobWriteSession(BlobInfo blobInfo, BlobWriteOption... options) {
     return throwGrpcOnly(fmtMethodName("blobWriteSession", BlobInfo.class, BlobWriteOption.class));
   }
+
+  /**
+   * Atomically move an object from one name to another.
+   *
+   * <p>This new method is an atomic equivalent of the previous rewrite + delete, however without
```
JesseLovelace, line 5035: I think "the previous" isn't the right phrasing here, because you can still do it that way right?

Maybe this should say something more like "Using this method is an atomic equivalent of doing a {@code Storage.copy} and then {@code Storage.delete}, except that it doesn't change the metadata fields of the target object"

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
BenWhitehead: We could, but I sort of like having the builder methods that nudge people to at least think about preconditions. A move is only retryable if the target has a precondition set on it.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
BenWhitehead: I'll remove `previous`, but there are some metadata fields that are changed. I know at least generation, create_time change when the move happens.

I'll also add the cross-method links as you suggest.

JesseLovelace (approved):

PR #2941, opened 2025-02-19
feat: add new Options to allow per method header values

BenWhitehead:
Add new "Option"s for those methods which already have option types to allow providing an ImmutableMap<String, String> to be applied as extra headers to all requests sent as part of that operation.

If an operation has multiple sources of input Options (rewrite) the "first" (i.e. source option) will be the one added to the request.

The following resources do not have "Option"s and therefor do not have extra headers support at this time:
* Acl
* DefaultAcl
* ServiceAccount
* Notification

### Overview of the changes

This blind number of lines is large, but the number of differences is actually fairly low.

* Add `Headers` Opt` in UnifiedOpts, this is the primary guts of the new feature and is responsible for holding onto the extra header values and validating the provided headers to ensure no collisions with headers otherwise set by our library/restricted by gcs.
* The ~800 lines of change in Storage.java are the same "add new option `extraHeaders`" to each of the 15 option classes along with accompanying javadocs. All javadocs are the exact same, and were copy-pasted for each method.
* Create new utility method `Utils#headerNameToLowerCase` to centralize the logic for transforming header names across all locations
* Add new `StorageRpc.Options#EXTRA_HEADERS` to support extra headers for JSON calls
  * update JSON request building code in `HttpStorageRpc` (433 lines), `ApiaryUnbufferedReadableByteChannel` (13 lines) and `JsonResumable*` (~45 lines + ~50 lines for existing tests) classes to take `EXTRA_HEADERS` into consideration
* test: add new suite of tests `ITExtraHeadersOptionTest` to validate headers are properly added to each request of various operations
* test: introduce new interface `AssertRequestHeaders` which both `GrpcRequestAuditing` and `RequestAuditing` now implement -- this interface makes it easy for each transport to share common methods for asserting request headers

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
@@ -62,7 +62,10 @@ import java.net.URLConnection;
 import java.security.Key;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Locale;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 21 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -1920,6 +1935,184 @@ public Mapper<MoveObjectRequest.Builder> moveObject() {
...
+    @SuppressWarnings("unchecked")
+    @Override
+    public Mapper<ImmutableMap.Builder<StorageRpc.Option, Object>> mapper() {
+      return optionBuilder -> {
+        if (val.isEmpty()) {
+          return optionBuilder;
+        }
+        // not ideal, but ImmutableMap.Builder doesn't have any read methods so we can detect
+        // collision
+        // before build time.
```
ddelgrosso1, line 2053: nit: can probably condense this into 2 lines.

ddelgrosso1 (approved):
I'm not a java expert but LGTM

google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
BenWhitehead: Ah, the formatter wrapped the first line. I'll clean this up.

PR #2954, opened 2025-02-25
fix: categorize a WatchdogTimeoutException as retriable for grpc ReadObject

BenWhitehead:
By categorizing as retriable the client will be allowed to retry the request rather than failing out.

[1] [`WatchdogTimeoutException`](https://github.com/googleapis/sdk-platform-java/blob/v2.53.0/gax-java/gax/src/main/java/com/google/api/gax/rpc/WatchdogTimeoutException.java#L45)


cojenco (approved):
🐶

PR #2974, opened 2025-03-05
fix: update usages of String.format to explicitly pass Locale.US

### LINKED ISSUE #2972: Storage: Content-Range header is formatted with default locale
#### Environment details
OS type and version: `Mac OS X/15.3.1/aarch64`
3. Java version: `Oracle Corporation/OpenJDK 64-Bit Server VM/23/23+37-2369`
4.  version(s): `2.48.2`

#### Steps to reproduce

  1. Run Java with locale that uses non-english number characters (e.g. `ps`)
  2. Execute a resumable upload

#### Any additional information below

The range header is generated formatting the numbers using the default locale. So for `ps` we get

`bytes ۰-۹۲۱۷۳۷۵/۹۲۱۷۳۷۶` when it should be `bytes 0-9217376/9217376`

See examples of the issue
 - https://github.com/googleapis/java-storage/blob/68e1cf21b5b88157a6cfaca2963f3ac84818c78c/google-cloud-storage/src/main/java/com/google/cloud/storage/HttpContentRange.java#L94
 - https://github.com/googleapis/java-storage/blob/68e1cf21b5b88157a6cfaca2963f3ac84818c78c/google-cloud-storage/src/main/java/com/google/cloud/storage/HttpContentRange.java#L148

It seems it could be resolved by using `%s` instead of `%d` perhaps?

BenWhitehead:
Usage of `Storage#format(String, Object...)` will use the default locale for it's formatting. This can lead to unexpected formatting if a right-to-left language such as Arabic is the default locale.

Update all usages to use `String.format(Locale.US, pattern, Object...)` so we ensure things like headers or error messages are formatted according to the US conventions which match `en_US` and ascii byte conventions.

Incidentally, things like right-to-left formatting seem to only apply to java11+ not java8.

Fixes #2972

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryUnbufferedReadableByteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ApiaryUnbufferedReadableByteChannel.java
@@ -52,6 +52,7 @@ import java.nio.channels.ClosedChannelException;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.channels.ScatteringByteChannel;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.function.Function;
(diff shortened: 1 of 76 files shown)


ddelgrosso1 (approved):
Yowza! This touches a lot of files but looks pretty boilerplate

nicktindall:
> Incidentally, things like right-to-left formatting seem to only apply to java11+ not java8.

Perhaps related to https://openjdk.org/jeps/252

PR #3014, opened 2025-03-28
deps: remove explicit version declarations for packages that are in shared-dependencies 

JesseLovelace:
gcp-resources still declares a version for now, but that will also be added to shared-deps and removed later

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,33 +82,6 @@
       * io.opentelemetry.contrib:opentelemetry-gcp-resources
       -->
       <dependency>
-        <groupId>io.opentelemetry.semconv</groupId>
-        <artifactId>opentelemetry-semconv</artifactId>
-        <version>1.27.0-alpha</version>
-      </dependency>
-      <dependency>
-        <groupId>com.google.cloud.opentelemetry</groupId>
-        <artifactId>detector-resources</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -81,33 +81,6 @@
       * io.opentelemetry.instrumentation:opentelemetry-resources
```
BenWhitehead, line 81: Can we remove this big comment too, it's no longer needed now that we're removing the semconv declaration.

pom.xml
```diff
@@ -81,33 +81,6 @@
...
-        <exclusions>
-          <exclusion>
-            <groupId>io.opentelemetry.semconv</groupId>
-            <artifactId>opentelemetry-semconv</artifactId>
-          </exclusion>
-        </exclusions>
-      </dependency>
       <dependency>
         <groupId>io.opentelemetry.contrib</groupId>
         <artifactId>opentelemetry-gcp-resources</artifactId>
```
BenWhitehead, line 86: Can we remove the exclusion from this as well? Now that the semceonv is managed by the bom it's version will take precedence.

BenWhitehead (approved):
A couple additional cleanup items, but feel free to merge after those are taken care of.

PR #3026, opened 2025-04-09
chore: make ReadProjectionConfigs#asFuture{Bytes,ByteString} public

BenWhitehead:
Additional javadocs, test added to ensure getting a future after a session has closed results in an exception.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/AsyncSessionClosedException.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/AsyncSessionClosedException.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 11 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/AsyncSessionClosedException.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.google.api.core.BetaApi;
+
+/**
+ * Root exception for async tasks which fail due to a session being closed.
+ *
+ * @see BlobReadSession
+ */
+@BetaApi
+public final class AsyncSessionClosedException extends RuntimeException {
```
bajajneha27, line 27: What's the difference between this new class vs `AsynchronousCloseException` ?

bajajneha27 (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/AsyncSessionClosedException.java
BenWhitehead: `java.nio.channels.AsynchronousCloseException` is intended to be used for channels specifically. In our usage here, we're using it on our session object not on a channel, so it feel more appropriate to use our own exception.

PR #3035, opened 2025-04-14
chore: fix a ConcurrentModificationException during BlobReadSession#close()

BenWhitehead:
A ConcurrentModificationException could happen while cleaning up an individual read at the same time the session had multiple child streams.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ObjectReadSessionImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ObjectReadSessionImpl.java
@@ -32,12 +32,13 @@ import java.io.InterruptedIOException;
 import java.util.ArrayList;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map.Entry;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.locks.ReentrantLock;
-import org.checkerframework.checker.lock.qual.GuardedBy;
+import java.util.function.BiFunction;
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/ObjectReadSessionImpl.java
```diff
@@ -81,54 +77,42 @@ public Object getResource() {
...
-          throw new IllegalStateException(
-              String.format(
-                  Locale.US,
-                  "Broken java enum %s value=%s",
-                  ProjectionType.class.getName(),
-                  config.getType().name()));
-      }
-    } finally {
-      lock.unlock();
+    checkState(open, "Session already closed");
```
BenWhitehead, line 80: locking has been moved into the new ConcurrentIdentityMap, and the try-finally has been removed changing the indentation.

google-cloud-storage/src/main/java/com/google/cloud/storage/ObjectReadSessionImpl.java
```diff
@@ -81,54 +77,42 @@ public Object getResource() {
...
-        Entry<ObjectReadSessionStream, ObjectReadSessionState> next = it.next();
-        ObjectReadSessionStream subStream = next.getKey();
-        it.remove();
-        closing.add(subStream.closeAsync());
+      if (!open) {
+        return;
       }
+      open = false;
+      List<ApiFuture<Void>> closing =
+          children.drainEntries((subStream, subStreamState) -> subStream.closeAsync());
```
BenWhitehead, line 108: use the new method, instead of manually traversing the iterator

bajajneha27 (approved):

PR #3097, opened 2025-05-07
fix: update grpc client side metrics detection to be graceful when not running on gcp

BenWhitehead:
Refactor OpenTelemetryBootstrappingUtils to return a channel configurator function rather than mutating a passed in value.

Add ChannelConfigurator interface to make testing easier and to simplify composition with any existing channel configurator function present in the InstantiatingGrpcChannelProvider.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/OpenTelemetryBootstrappingUtils.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/OpenTelemetryBootstrappingUtils.java
@@ -17,7 +17,6 @@
 package com.google.cloud.storage;
 
 import com.google.api.core.ApiFunction;
-import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
 import com.google.api.gax.rpc.PermissionDeniedException;
 import com.google.api.gax.rpc.UnavailableException;
 import com.google.cloud.opentelemetry.metric.GoogleCloudMetricExporter;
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/OpenTelemetryBootstrappingUtils.java
```diff
@@ -88,15 +89,41 @@ final class OpenTelemetryBootstrappingUtils {
...
       boolean shouldSuppressExceptions) {
+    GCPResourceProvider resourceProvider = new GCPResourceProvider();
+    Attributes detectedAttributes = resourceProvider.getAttributes();
+
+    @Nullable String detectedProjectId =
+        detectedAttributes.get(AttributeKey.stringKey("cloud.account.id"));
+    if (projectId == null && detectedProjectId == null) {
+      log.warning(
+          "Unable to determine the Project ID in order to report metrics. No gRPC client metrics will be reported.");
+      return channelConfigurator != null ? channelConfigurator : ChannelConfigurator.identity();
```
cojenco, line 107: qq: is this the case of not running on GCP?

cojenco (approved):
Thanks Ben!

google-cloud-storage/src/main/java/com/google/cloud/storage/OpenTelemetryBootstrappingUtils.java
BenWhitehead: In a round about manner yes. The metrics reporting relies on application default credentials for initialization, and if the project can't be determined either as a passed in value from the Storage client, or resolved from the GCP resolved otel attributes there isn't any way to bootstrap.

PR #3119, opened 2025-05-27
feat: expose BucketInfo.getProject as a BigInteger

### LINKED ISSUE #3023: Add getProjectNumber method to Bucket.java
**Is your feature request related to a problem? Please describe.**

The Java SDK for GCS does not have a method for fetching the project number even though the value is returned from the describeBucket API. This helper exists on the C# SDK as per: https://cloud.google.com/storage/docs/getting-bucket-metadata#storage-get-bucket-metadata-csharp.

Please also see also JSON API reference: https://cloud.google.com/storage/docs/json_api/v1/buckets#resource

**Describe the solution you'd like**
Add the method for returning the projectNumber. I intend to open a PR shortly adding the required method.

**Describe alternatives you've considered**
n/a

BenWhitehead:
Expose publicly the project field which was previously internal only. This field represents the project number of the project the bucket belongs to.

This field is still OUTPUT_ONLY, and as such does hot have a setter exposed.

Fixes #3023


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITOptionRegressionTest.java
```diff
@@ -328,7 +328,7 @@ public void storage_BucketGetOption_fields_BucketField() {
             "website",
             "softDeletePolicy",
             "hierarchicalNamespace",
-            "website");
+            "projectNumber");
```
BenWhitehead, line 331: `website` was defined twice, changing the second one to our new expected value.

bajajneha27 (approved):

PR #3166, opened 2025-06-24
fix: fix Journaling BlobWriteSessionConfig to properly handle multiple consecutive retries

BenWhitehead:
* Clear copy buffer used to load bytes from the recovery file so, we don't have dangling bytes from a previous retry
* Do not add to the cumulative crc32c when retrying. The bytes have already been added in the first attempt.


cojenco (approved):
Thanks for walking me through the changes!

wjohn915 (approved):

PR #3167, opened 2025-06-24
chore: cleanup Crc32cValue and associated Hasher operations

BenWhitehead:
* Add `Crc32cValue#zero()`
  * `Crc32cValue.zero().concat(v) == v`
* Add temporary WriteCtx constructor that allows specifying the initial crc32c value
  * The desired default will be `Crc32cValue.zero()`, but not all tests or uploads are ready to handle this. Stick with the existing null as default, and provide the overload to selectively opt into things incrementally.
* Update Hasher.nullSafeConcat to account for the use of `zero`
  * if Hasher.noop() always return null
  * if Hasher.enabled() only return null if left side is null
* Add nullability annotation to Crc32cValue.nullSafeConcat

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/WriteCtx.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/WriteCtx.java
@@ -37,7 +37,7 @@ final class WriteCtx<RequestFactoryT extends WriteObjectRequestBuilderFactory> {
     this.requestFactory = requestFactory;
     this.totalSentBytes = new AtomicLong(0);
     this.confirmedBytes = new AtomicLong(0);
-    this.cumulativeCrc32c = new AtomicReference<>(Crc32cValue.zero());
+    this.cumulativeCrc32c = new AtomicReference<>(null);
   }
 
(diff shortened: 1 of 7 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/WriteCtx.java
```diff
@@ -31,13 +31,22 @@ final class WriteCtx<RequestFactoryT extends WriteObjectRequestBuilderFactory> {
...
+  private final AtomicReference<@Nullable Crc32cLengthKnown> cumulativeCrc32c;
 
   WriteCtx(RequestFactoryT requestFactory) {
+    this(requestFactory, null);
+  }
+
+  /**
+   * TODO: Remove initialValue and replace with Crc32cValue.zero() once all uploads have been
+   * updated to do e2e checksumming by default.
+   */
```
cojenco, line 43: Discussed in GVC; this is later updated in https://github.com/googleapis/java-storage/pull/3176/

BenWhitehead:

PR #3170, opened 2025-06-25
deps: manage hermetic_library_generation action version via Hermetic Build only

diegomarquezp:
The fix https://github.com/googleapis/sdk-platform-java/pull/3800 was meant for GraalVM image versions.

The regression in #3169, however, affects the hermetic_library_generation GH action instead, which is not in [the sdk-platform-java template](https://github.com/googleapis/sdk-platform-java/blob/844af0f51b016dd03870d36a80bf04f7521dc6e8/hermetic_build/library_generation/owlbot/templates/java_library/renovate.json), but java-storage does not consume the renovate.json template [as per its owlbot.py](https://github.com/googleapis/java-storage/blob/b956eb9b03831405f05691d50ae2de6bf422f73e/owlbot.py#L43).

This PR removes the action yaml from renovate.json, so it's only managed by Hermetic Build


BenWhitehead:
Merge after https://github.com/googleapis/java-storage/pull/3169 merges

PR #3173, opened 2025-06-26
chore: add Hasher#defaultHasher()

BenWhitehead:
By default, it will be the same as Hasher.enabled(). If running on java < 9, a log message will be logged making it apparent there isn't a fast implementation of crc32c.

A system property `com.google.cloud.storage.Hasher.default` can be set to "disabled" to instead make the default Hasher.noop().


BenWhitehead:

PR #3174, opened 2025-06-26
chore: add UnifiedOpts.Opts#getHasher() to return a hasher relative to the set of Opts defined for an operation.

BenWhitehead:
Add `UnifiedOpts.HasherSelector` (also add `UnifiedOptions.BucketObjectHmacKeyAllOpt` which extends all `{Bucket,Object,HmacKey}{Target,List,Source}Opt` and provides an anchor for those opts which apply to all 9 permutations [UserProject & Headers])

Add Opts#getHasher() that will return Hasher.defaultInstance() or a hasher defined based on the last occurring HasherProvider present in the instance of Opts.

Both Crc32cMatch and Md5Match are configured to return Hasher.noop(). If an explicit checksum value is provided by the customer, use it and disable our automatic crc32c calculation.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
@@ -281,6 +281,10 @@ final class UnifiedOpts {
     O asSource();
   }
 
+  interface HasherSelector extends BucketObjectHmacKeyAllOpt {
+    Hasher getHasher();
+  }
+
   /**
    * This class extends off {@link ObjectSourceOpt} and {@link ObjectTargetOpt} in order to satisfy
    * some the shimming constraints of the subclasses of {@link OptionShim}.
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/UnifiedOpts.java
```diff
@@ -589,7 +593,11 @@ static Headers extraHeaders(ImmutableMap<String, String> extraHeaders) {
     return new Headers(extraHeaders);
   }
 
-  static final class Crc32cMatch implements ObjectTargetOpt {
+  static DefaultHasherSelector defaultHasherSelector() {
+    return DefaultHasherSelector.INSTANCE;
+  }
+
+  static final class Crc32cMatch implements ObjectTargetOpt, HasherSelector {
```
cojenco, line 600: Very cool! Thanks for walking through the UnifiedOpts interfaces

cojenco (approved):
LGTM, had a GVC review session with Ben

BenWhitehead:

PR #3182, opened 2025-07-01
fix: give user provided checksum precondition priority for Storage#create methods that accept byte[]

BenWhitehead:
* `Storage#create(BlobInfo, BlobTargetOption...)`
* `Storage#create(BlobInfo, byte[], BlobTargetOption...)`
* `Storage#create(BlobInfo, byte[], int, int, BlobTargetOption...)`

If no user provided checksum is specified, crc32c will still be calculated.

Applies to both `StorageOptions.http()` and `StorageOptions.grpc()` instances.

test: remove obsolete mockito tests that don't test anything unique

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
@@ -1717,14 +1717,15 @@ final class StorageImpl extends BaseService<StorageOptions> implements Storage, 
 
   @Override
   public BlobInfo internalDirectUpload(BlobInfo info, Opts<ObjectTargetOpt> opts, ByteBuffer buf) {
-    BlobInfo.Builder builder = info.toBuilder();
-    @Nullable Crc32cLengthKnown hash = opts.getHasher().hash(buf);
+    BlobInfo.Builder builder =
+        opts.blobInfoMapper().apply(info.toBuilder().clearMd5().clearCrc32c());
+    @Nullable Crc32cLengthKnown hash = opts.getHasher().hash(buf.duplicate());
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java
```diff
@@ -1767,17 +1717,15 @@ public BlobInfo internalCreateFrom(Path path, BlobInfo info, Opts<ObjectTargetOp
...
-
     BlobInfo.Builder builder =
-        info.toBuilder()
-            .setMd5(
-                BaseEncoding.base64().encode(Hashing.md5().hashBytes(buf.duplicate()).asBytes()))
-            .setCrc32c(
-                BaseEncoding.base64()
-                    .encode(Ints.toByteArray(Hashing.crc32c().hashBytes(buf.duplicate()).asInt())));
+        opts.blobInfoMapper().apply(info.toBuilder().clearMd5().clearCrc32c());
+    @Nullable Crc32cLengthKnown hash = opts.getHasher().hash(buf.duplicate());
```
BrandonY, line 1722: Does this create a copy of the data? Can you get a hash without doing that?

BenWhitehead: `ByteBuffer#duplicate` creates a new buffer but shares the underlying bytes https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#duplicate--

This allows us to not have to manually rewind the buffer after it's been consumed by hashing.

BrandonY (approved):

BenWhitehead:

PR #3187, opened 2025-07-08
fix: move crc32c computation before writing to disk for BufferToDiskThenUpload BlobWriteSession config

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/BufferToDiskThenUpload.java
```diff
@@ -211,41 +215,78 @@ public ApiFuture<BlobInfo> getResult() {
...
-                gcs,
-                size,
-                () -> {
-                  BlobInfo blob = storage.internalCreateFrom(path, info, opts);
-                  result.set(blob);
-                });
-          } catch (StorageException | IOException e) {
-            result.setException(e);
-            throw e;
+          lock.lock();
```
BrandonY, line 257: maybe a dumb question, but why does this need a lock now if it didn't need one before?

BenWhitehead: Previously, we didn't have any state we need to keep in sync with what was being written to disk in `write`. Now that we're hashing before writing to disk I want to make sure that our hashing is made threadsafe.

BenWhitehead:

PR #3204, opened 2025-07-23
feat: add new preview Bucket encryption policy configuration

BenWhitehead:
* GoogleManagedEncryptionEnforcementConfig
* CustomerManagedEncryptionEnforcementConfig
* CustomerSuppliedEncryptionEnforcementConfig

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
@@ -2091,7 +2091,7 @@ public class BucketInfo implements Serializable {
   /**
    * Google Managed Encryption (GMEK) enforcement config of a bucket.
    *
-   * @since 2.54.0 This new api is in preview and is subject to breaking changes.
+   * @since 2.55.0 This new api is in preview and is subject to breaking changes.
    */
   @BetaApi
   public static final class GoogleManagedEncryptionEnforcementConfig {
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java
```diff
@@ -164,6 +172,93 @@ final class GrpcConversions {
...
+  private final Codec<EncryptionEnforcementRestrictionMode, String>
+      encryptionEnforcementRestrictionModeCodec =
+          Codec.of(
+              EncryptionEnforcementRestrictionMode::toString,
+              EncryptionEnforcementRestrictionMode::valueOf);
+  private final Codec<
+          GoogleManagedEncryptionEnforcementConfig,
+          Encryption.GoogleManagedEncryptionEnforcementConfig>
+      googleManagedEncryptionEnforcementConfigCodec =
+          Codec.of(
```
BrandonY, line 184: This stuff seems very boilerplate. I wonder if there's a good way to pare it down or make more in common with the other methods. Probably nothing to do about it just now, so just an observation.

BenWhitehead: It is. Unfortunately, the types we are touching are grpc generated and apiary generated classes. Which have the same shape, but none of the shared common class hierarchy. We'd either need what we have here, or some facade that was able to sit over both and make this code easier, but that facade would need to do some translation/mapping itself. So, while not ideal by an abstraction perspective, the conversions classes are at least mostly self contained and low maintenance once the conversions are added.

BrandonY (approved):

PR #3217, opened 2025-08-05
fix: make FlushPolicy${Min,Max}FlushSizeFlushPolicy constructors private

BenWhitehead:
They still retain their public factory methods, just the constructors are now private -- they should have been to begin with.


bajajneha27 (approved):

BenWhitehead:

PR #3222, opened 2025-08-05
chore: add BidiUploadState

BenWhitehead:
This will be used by an upcoming PR that will integrate it with a BidiWriteObject stream.

At a high level this class hierarchy models the state and lifecycle events which happen when performing an upload -- in this change append specific, but many generalize to all uploads.


google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
```diff
@@ -0,0 +1,1099 @@
...
+import org.slf4j.MarkerFactory;
+
+@SuppressWarnings("LoggingSimilarMessage")
+abstract class BidiUploadState {
+  private static final Logger LOGGER = LoggerFactory.getLogger(BidiUploadState.class);
+  private static final Marker TRACE_ENTER = MarkerFactory.getMarker("enter");
+  private static final Marker TRACE_EXIT = MarkerFactory.getMarker("exit");
+
+  static final OneofDescriptor FIRST_MESSAGE_DESCRIPTOR =
+      BidiWriteObjectRequest.getDescriptor().getOneofs().stream()
```
BrandonY, line 68: This can't be the right way to get at the "first_message" field, can it?

google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
```diff
@@ -0,0 +1,1099 @@
...
+   * first then second.
+   *
+   * @throws IllegalArgumentException if both first and second have checksummedData
+   */
+  static BidiWriteObjectRequest concatenate(
+      BidiWriteObjectRequest first, BidiWriteObjectRequest second) {
+    checkArgument(
+        !(first.hasChecksummedData() && second.hasChecksummedData()),
+        "attempting to merge two requests that both specify checksummed_data");
+    BidiWriteObjectRequest.Builder b = first.toBuilder().mergeFrom(second);
```
BrandonY, line 198: Does this throw away the data from one of the two writes?

google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
BenWhitehead: It's the easiest way to probe whether it is set or not. protoc doesn't generate a `.hasFirstMessage()` method like it does for optional fields. But the base message class has `com.google.protobuf.GeneratedMessageV3#hasOneof(com.google.protobuf.Descriptors.OneofDescriptor)` method. Using this to probe presence felt cleaner than `m.hasUploadId() || m.hasWriteObjectSpec() || m.hasAppendObjectSpec()`.

google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
BenWhitehead: The `checkArgument` on line 195-197 guarantees that at most one of the messages contains data. By the time we make it to `mergeFrom`, we know that only 0 or 1 of the messages contains any actual data. If there were a bug in the caller of this method, the exception from `checkArgument` would prevent it from silently making a bad mutation.

BenWhitehead:

PR #3223, opened 2025-08-05
chore: add BidiUploadStreamingStream

BenWhitehead:
This class is a higher level abstraction over bidi stream, retries and state tracking. Taking care of most concerns once data is in a ChunkSegment.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
@@ -74,6 +74,17 @@ abstract class BidiUploadState {
   // seal this class to extension
   private BidiUploadState() {}
 
+  @VisibleForTesting
+  BidiUploadState(String testName) {
+    // some runtime enforcement that this constructor is only called from a test
+    // if we had java9+ we could seal this all the way without this hack
+    StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
+    boolean isJunitTest =
+        Arrays.stream(stackTrace).anyMatch(ste -> ste.getClassName().startsWith("org.junit"));
+
+    checkState(isJunitTest, "not a junit test", testName);
+  }
+
   protected final StorageException err(
       UploadFailureScenario scenario, BidiWriteObjectResponse response) {
     BidiWriteObjectRequest t = peekLast();
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadStreamingStream.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadStreamingStream.java
@@ -1,0 +1,593 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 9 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
```diff
@@ -74,6 +74,17 @@ abstract class BidiUploadState {
   // seal this class to extension
   private BidiUploadState() {}
 
+  @VisibleForTesting
+  BidiUploadState(String testName) {
```
BrandonY, line 78: Seems odd to need a visible-for-testing constructor of an abstract class, but presumably it'll make sense further down.

google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadStreamingStream.java
```diff
@@ -0,0 +1,593 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Consumer;
+import org.checkerframework.checker.nullness.qual.EnsuresNonNull;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.nullness.qual.RequiresNonNull;
+
+final class BidiUploadStreamingStream {
```
BrandonY, line 50: "Streaming" in this case presumably refers to the user streaming unknown data into a Writer? Maybe worth a comment, at least.

google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadStreamingStream.java
```diff
@@ -0,0 +1,593 @@
...
+    this.redirectCounter = new AtomicInteger();
+  }
+
+  public ApiFuture<BidiWriteObjectResponse> getResultFuture() {
+    return state.getResultFuture();
+  }
+
+  public boolean append(ChunkSegmenter.@NonNull ChunkSegment data) {
+    lock.lock();
+    try {
```
BrandonY, line 97: I was going to suggest using a try-with-resource pattern, but it seems that locks aren't AutoClosable. Odd.

BenWhitehead: yeah, unfortunately they aren't. we could make a construct that is but that would require an allocation in order to acquire the lock and probably not worth it here.

google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
BenWhitehead: This is here to facilitate validating some of the behavioral interactions from the BidiStream to the State where getting into a specific state would otherwise be difficult with the actual concrete implementation.

For example, in this test https://github.com/googleapis/java-storage/blob/365bce5e6782581c8e064518848ea30016ec25bb/google-cloud-storage/src/test/java/com/google/cloud/storage/BidiUploadTest.java#L1947-L1951 we are verifying that the redirect is routed to the state by the response observer, whereas doing that same validation with the concrete class would require more involved setup and validation.

BenWhitehead:

PR #3224, opened 2025-08-05
chore: add BidiAppendableUnbufferedWritableByteChannel

BenWhitehead:
This will have tests added in a followup commit

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiAppendableUnbufferedWritableByteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiAppendableUnbufferedWritableByteChannel.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BidiAppendableUnbufferedWritableByteChannel.java
```diff
@@ -0,0 +1,193 @@
...
+    this.writeThrewError = false;
+  }
+
+  @Override
+  public long write(ByteBuffer[] srcs, int srcsOffset, int srcsLength) throws IOException {
+    return internalWrite(srcs, srcsOffset, srcsLength);
+  }
+
+  @Override
+  public long writeAndClose(ByteBuffer[] srcs, int offset, int length) throws IOException {
```
BrandonY, line 63: Why isn't this just

  long written = write(srcs, offset, length);
  close();
  return written;

BenWhitehead: The `internalWrite` and `write` methods are both non-blocking and may not consume all of the bytes present in srcs. However, close does need to block. So we loop here to only call close once all of the bytes have been successfully enqueued. Then close can do it's normal blocking waiting for the necessary ack from gcs.

BenWhitehead:

PR #3225, opened 2025-08-05
chore: make MinFlushBufferedWritableByteChannel capable of being non-blocking

BenWhitehead:
Default is still blocking, but non-blocking will be default in a later commit


BrandonY (approved):

BenWhitehead:

PR #3226, opened 2025-08-05
chore: new bidi appendable channel bootstrappable

BenWhitehead:
* Add FlushPolicy.MinFlushSizeFlushPolicy.maxPendingBytes
* Add new contructor to ChannelSession to allow the resultFuture to be passed in.
* Create temporary BlobAppendableUploadImpl.AppendableUnbufferedWritableByteChannel to allow both old and new implementations to be passed to BlobAppendableUploadImpl -- this interface will be removed in cleanup
* Add new implementation using streaming buffer management as a sibling to the existing implementation. The existing implementation will be removed entirely in cleanup.
* Refactored tests to follow

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobAppendableUploadConfig.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobAppendableUploadConfig.java
@@ -19,21 +19,28 @@ package com.google.cloud.storage;
 import static com.google.cloud.storage.ByteSizeConstants._256KiB;
 import static java.util.Objects.requireNonNull;
 
+import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
 import com.google.api.core.BetaApi;
 import com.google.api.core.InternalApi;
+import com.google.api.core.SettableApiFuture;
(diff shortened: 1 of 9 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobAppendableUploadConfig.java
```diff
@@ -217,55 +239,127 @@ public enum CloseAction {
   }
 
   BlobAppendableUpload create(GrpcStorageImpl storage, BlobInfo info, Opts<ObjectTargetOpt> opts) {
-    boolean takeOver = info.getGeneration() != null;
-    BidiWriteObjectRequest req =
-        takeOver
-            ? storage.getBidiWriteObjectRequestForTakeover(info, opts)
-            : storage.getBidiWriteObjectRequest(info, opts);
+    if (newImpl) {
+      // TODO: make configurable
```
BrandonY, line 243: Are there style rules for TODOs? Do they need bug numbers or names or something? Actually, is there a style guide in general I should be checking?

BenWhitehead:

PR #3228, opened 2025-08-05
test: refactor existing appendable upload tests to work with new implementation

BenWhitehead:
* Add incremental checksums to all messages
* Add final object_checksums for all finalization
* Remove tests that were testing the same thing -- especially when the setup was fairly involved
* Remove some Integration tests that are now covered by unit tests of the bidi stream / bidi upload state (redirect limits, etc)
* Refactor lots of common messages and test run flows to reduce duplication and make debugging easier
* Ignore some tests that previously would partially split data rather than messages -- the current implementation only evicts when a whole message is consumed
* In general, signal `onComplete()` after `onNext` in response to a finalize message
* Update ITAppendableUploadTest to run for multiple permutations of BlobAppendableUploadConfigs

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/ITAppendableUploadTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/ITAppendableUploadTest.java
@@ -15,29 +15,39 @@
  */
 package com.google.cloud.storage;
 
-import static com.google.cloud.storage.ByteSizeConstants._2MiB;
+import static com.google.cloud.storage.ByteSizeConstants._1MiB;
 import static com.google.cloud.storage.TestUtils.assertAll;
+import static com.google.cloud.storage.TestUtils.xxd;
 import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assume.assumeFalse;
+import static org.junit.Assume.assumeTrue;
 
+import com.google.api.core.ApiFuture;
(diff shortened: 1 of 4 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/ITAppendableUploadTest.java
```diff
@@ -58,148 +69,230 @@ public final class ITAppendableUploadTest {
...
+
   @Test
-  public void testAppendableBlobUpload()
+  public void appendableUpload_emptyObject()
       throws IOException, ExecutionException, InterruptedException, TimeoutException {
-    BlobAppendableUploadConfig uploadConfig =
-        BlobAppendableUploadConfig.of()
-            .withFlushPolicy(FlushPolicy.maxFlushSize(2000))
-            .withCloseAction(CloseAction.FINALIZE_WHEN_CLOSING);
+    assumeTrue(
```
BrandonY, line 79: Nit: This could be better as three assertTrue() statements

BenWhitehead: This is an `assumeTrue` not `assertTrue`, which will dynamically skip running this tests if the statement is not true.

I think most of these assumes can go away once we're able to run our CI in prod against a live rapid bucket rather than the partial implementation of testbench used currently.

BenWhitehead:

PR #3229, opened 2025-08-05
chore: remove obsolete internal hasher option

BenWhitehead:
All appendable uploads will crc32c checksum their segments always.


BenWhitehead:

PR #3235, opened 2025-08-06
docs: update BlobAppendableUpload.AppendableUploadWriteableByteChannel to indicate it is non-blocking

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/BlobAppendableUpload.java
```diff
@@ -90,14 +91,38 @@ public interface BlobAppendableUpload extends BlobWriteSession {
...
+     * <b>This method is non-blocking</b>
+     *
+     * <p>Consume as many bytes as can fit in the underlying outbound queue. The size of the
+     * outbound queue is determined from {@link BlobAppendableUploadConfig#getFlushPolicy()}{@code
+     * .}{@link FlushPolicy#getMaxPendingBytes() getMaxPendingBytes()}. If the outbound queue is
+     * full, and can not fit more bytes, this method will return 0.
+     *
+     * <p>This method may be invoked at any time. If another thread has already initiated a write
+     * operation upon this channel, however, then an invocation of this method will block until the
+     * first operation is complete.
```
BrandonY, line 112: Seems contradictory with the first claim? Unless we don't consider that blocking because the first call is simply doing a memory copy?

BrandonY (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/BlobAppendableUpload.java
BenWhitehead: I'll remove that. We can make the lock acquisition conditional too and return 0 if we can't acquire the lock.

BenWhitehead:

PR #3241, opened 2025-08-11
fix: fix otel span lifetime for BlobAppendableUpload

BenWhitehead:
Previously, the BlobAppendableUpload span would when when open completed, now it will end when the opened channel completes.


BenWhitehead:

PR #3248, opened 2025-08-19
chore: make DefaultBufferedWritableByteChannel capable of being non-blocking

BenWhitehead:
non-blocking mode will only be used by appendable uploads, all existing usage will continue to use the existing blocking semantics.


BenWhitehead:
Same logical work for our default buffering which will only flush when the buffer is full as was done in https://github.com/googleapis/java-storage/pull/3225 for our min flush threshold.

ddelgrosso1 (approved):

ddelgrosso1 (approved):

BenWhitehead:

PR #3252, opened 2025-08-20
chore: update BidiUploadStreamingStream to avoid enqueuing multiple equivalent flushes

BenWhitehead:
If the downstream queue doesn't have enough space to accept the message, we can trigger a flush to ensure some progress is made.

However, if the attempted write is not consumed multiple times we want to avoid enqueueing the standalone flush multiple times.

Flag the rewindable content as dirty so the rewind will work when only partial consumption happens.


bajajneha27 (approved):

BenWhitehead:

PR #3255, opened 2025-08-20
fix: update otel integration to properly activate span context for lazy RPCs such as reads & writes

BenWhitehead:
* Fixes for WriteChannel
* Fixes for ReadChannel
* Fixes for BlobWriteSession
* Fixes for BlobReadSession 
  * `s$com.google.cloud.storage.Storage/readAs$com.google.cloud.storage.Storage/blobReadSession/readAs$g`
* Fixes for BlobAppendableUpload

Prior to these fixes the `grpc/call/interceptCall` could end up orphaned from the span that the RPC was actually performed for.

Now it looks like the following (The interceptCall is properly parented by the createFrom span):
image

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/OtelStorageDecorator.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/OtelStorageDecorator.java
@@ -64,10 +64,12 @@ import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 @SuppressWarnings("DuplicatedCode")
-final class OtelStorageDecorator implements Storage {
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/OtelStorageDecorator.java
```diff
@@ -1650,21 +1655,38 @@ public boolean isOpen() {
...
+        scope = null;
+      }
+    }
+
+    public void setScope() {
+      if (scope != null) {
+        clearScope();
       }
+      scope = span.makeCurrent();
     }
```
cojenco, line 1678: This looks great! TIL to have explicit scope handling added to properly activate spans for lazy RPCs

cojenco (approved):

PR #3261, opened 2025-08-22
feat: add AppendableUploadWriteableByteChannel#flush()

BenWhitehead:
Allows blocking the invoking thread until the number of bytes acknowledged by GCS matches the number of written bytes prior to calling flush().

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
@@ -254,6 +254,10 @@ abstract class BidiUploadState {
     unimplemented();
   }
 
+  public void awaitAck(long writeOffset) throws InterruptedException {
+    unimplemented();
+  }
+
   enum State {
     INITIALIZING,
     TAKEOVER,
@@ -286,6 +290,7 @@ abstract class BidiUploadState {
     protected final Supplier<GrpcCallContext> baseCallContext;
     protected final ReentrantLock lock;
     protected final Condition stateUpdated;
+    protected final Condition confirmedBytesUpdated;
 
     /** The maximum number of bytes allowed to be enqueued in {@link #queue} across all messages. */
     protected final long maxBytes;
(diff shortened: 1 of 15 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
```diff
@@ -883,6 +894,21 @@ public void awaitTakeoverStateReconciliation(Runnable restart) {
         throw StorageException.coalesce(e);
       }
     }
+
+    @Override
+    public void awaitAck(long writeOffset) throws InterruptedException {
+      lock.lock();
```
bajajneha27, line 900: question: why are we using lock here?

bajajneha27 (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
BenWhitehead: The Condition `confirmedBytesUpdated` is tied to the lock instance.

> Conditions (also known as condition queues or condition variables) provide a means for one thread to suspend execution (to "wait") until notified by another thread that some state condition may now be true. Because access to this shared state information occurs in different threads, it must be protected, so a lock of some form is associated with the condition. The key property that waiting for a condition provides is that it atomically releases the associated lock and suspends the current thread, just like Object.wait. [1]

When we await the condition on line 903, the lock will automatically be released while the invoking thread waits for the call of `await` to return.


* [1] [`Condition`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Condition.html#await-long-java.util.concurrent.TimeUnit-:~:text=Conditions%20(also%20known%20as%20condition%20queues%20or%20condition%20variables)%20provide%20a%20means%20for%20one%20thread%20to%20suspend%20execution%20(to%20%22wait%22)%20until%20notified%20by%20another%20thread%20that%20some%20state%20condition%20may%20now%20be%20true.) javadocs

PR #3273, opened 2025-09-02
chore: update BlobInfo.contexts diff handling to be deep rather than shallow

BenWhitehead:
Add JsonUtils class to provide some helpers for performing arbitrarily deep field selection. Our existing implementation is flat for everything exception metadata, with the addition of object contexts we now have an N depth diff (metadata is always depth 2, but a context path is `contexts.custom.<key>.value`) and the value of the key is an object instead of a string. We accomplish this arbitrary diffing by first flattening the src object to a map of paths to leaves and their corresponding string values. Then we diff the keys to produce a new map, and then treeify that new map back to a json structure.

This is quite robust, but isn't terribly efficient so we only use it for contexts and metadata fields.

Update BlobInfo.BuilderImpl#setContexts to deeply resolve the diff against the provided value.

Update BlobInfo object contexts maps to use unmodifiable hashmaps instead of ImmutableMap. ImmutableMap doesn't allow null values, but we need to accept null values to allow customers to remove individual values.

Add object contexts to the existing ITNestedUpdateMaskTest.

gRPC didn't require any special handling as it's `update_mask` already takes care of things appropriately after the deep diffing is added.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonUtils.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonUtils.java
@@ -1,0 +1,268 @@
+/*
+ * Copyright 2025 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.storage;
+
+import com.google.api.client.json.GenericJson;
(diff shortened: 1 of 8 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/JsonUtils.java
```diff
@@ -0,0 +1,268 @@
...
+
+  /**
+   * Given the provided {@code inputJson} flatten it to a Map&lt;String, String> where keys are the
+   * field path, and values are the string representation of the value. Then, create a
+   * Map&lt;String, String> by defining an entry for each value from {@code fieldsInOutput} with a
+   * null value. Then, diff the two maps retaining those entries that present in both, and adding
+   * entries that only exist in the right. Then, turn that diffed map back into a tree.
+   */
+  @VisibleForTesting
+  static @NonNull JsonObject getOutputJson(JsonObject inputJson, Set<String> fieldsInOutput) {
```
nidhiii-27, line 104: nit: should we rename getOutputJson to something more specific that describes its function, (maybe getOutputJsonWithSelectedFields/ extractFields)?

BenWhitehead: Done

ShreyasSinha (approved):

PR #3275, opened 2025-09-04
samples: add samples for ranged reads and appendable uploads

nidhiii-27:
Add samples:
storage_read_appendable_object_single_range
storage_read_appendable_object_multiple_ranges
storage_read_appendable_object_full
storage_start_appendable_object_upload
storage_resume_appendable_object_upload
storage_finalize_appendable_object_upload

Changed the zone in Zone.java (used for testing) from us-east1-c to us-central1-a because zonal buckets cannot be created in us-east1 region.


google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/Zone.java
```diff
@@ -89,7 +89,7 @@ public Zone get() {
     @Override
     public void start() {
       try {
-        zone = MetadataService.zone().orElseGet(() -> parse("us-east1-c"));
+        zone = MetadataService.zone().orElseGet(() -> parse("us-central1-a"));
```
BenWhitehead: The `MetadataService.zone()` actually resolves the zone from the vm it runs on. The `us-east1-c` is a correctly formatted zone string for a zone that doesn't exist today, so we have a valid format but a zone that can't actually work.

Please revert this change.

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectMultipleRangedRead.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+      try (ScatteringByteChannel channel2 = blobReadSession.readAs(readAsChannelConfig2)) {
+        int currentRead = 0;
+        while (buf2.hasRemaining() && currentRead != -1) {
+          currentRead = channel2.read(buf2);
+          if (currentRead > 0) {
+            bytesRead2 += currentRead;
+          }
+        }
```
BenWhitehead: Channels are actually blocking, and won't allow multiple ranges to be read at the same time unless you put them into separate threads.

A better projection here is probably `ReadAsAsyncBytes` which will get both ranges as the same time, resolving the futures when all the bytes have been accumulated.

You can check out this integration test for an example of what that usage would look like: https://github.com/googleapis/java-storage/blob/f63c8c770ba16444f3cd81ab220c5f355a22171f/google-cloud-storage/src/test/java/com/google/cloud/storage/ITObjectReadSessionTest.java#L130

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectMultipleRangedRead.java
```diff
@@ -0,0 +1,86 @@
...
+import java.nio.ByteBuffer;
+import java.nio.channels.ScatteringByteChannel;
+import java.util.Locale;
+import java.util.concurrent.TimeUnit;
+
+public class AppendableObjectMultipleRangedRead {
+  public static void appendableObjectMultipleRangedRead(
+      String bucketName, String objectName, long offset1, int length1, long offset2, int length2)
+      throws Exception {
+    Storage storage = StorageOptions.grpc().build().getService();
```
BenWhitehead: This instance of `Storage` needs to be in a try-with-resource so that it's cleaned up when we're done with it.

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectReadFullObject.java
```diff
@@ -0,0 +1,68 @@
...
+import java.nio.ByteBuffer;
+import java.nio.channels.ScatteringByteChannel;
+import java.util.Locale;
+import java.util.concurrent.TimeUnit;
+
+public class AppendableObjectReadFullObject {
+  public static void appendableObjectReadFullObject(String bucketName, String objectName)
+      throws Exception {
+
+    Storage storage = StorageOptions.grpc().build().getService();
```
BenWhitehead: This instance of `Storage` needs to be in a try-with-resource so that it's cleaned up when we're done with it.

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectSingleRangedRead.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.cloud.storage.StorageOptions;
+import java.nio.ByteBuffer;
+import java.nio.channels.ScatteringByteChannel;
+import java.util.Locale;
+import java.util.concurrent.TimeUnit;
+
+public class AppendableObjectSingleRangedRead {
+  public static void appendableObjectSingleRangedRead(
+      String bucketName, String objectName, long offset, int length) throws Exception {
+    Storage storage = StorageOptions.grpc().build().getService();
```
BenWhitehead: This instance of `Storage` needs to be in a try-with-resource so that it's cleaned up when we're done with it.

samples/snippets/src/main/java/com/example/storage/object/FinalizeAppendableObjectUpload.java
```diff
@@ -0,0 +1,58 @@
...
+public class FinalizeAppendableObjectUpload {
+  public static void finalizeAppendableObjectUpload(String bucketName, String objectName)
+      throws Exception {
+    // The ID of your GCS bucket
+    // String bucketName = "your-unique-bucket-name";
+
+    // The ID of your GCS GCS unfinalized appendable object
+    // String objectName = "your-object-name";
+
+    Storage storage = StorageOptions.grpc().build().getService();
```
BenWhitehead: This instance of `Storage` needs to be in a try-with-resource so that it's cleaned up when we're done with it.

samples/snippets/src/main/java/com/example/storage/object/ResumeAppendableObjectUpload.java
```diff
@@ -0,0 +1,90 @@
...
+    // The ID of your GCS bucket
+    // String bucketName = "your-unique-bucket-name";
+
+    // The ID of your GCS unfinalized appendable object
+    // String objectName = "your-object-name";
+
+    // The path to the file to upload
+    // String filePath = "path/to/your/file";
+
+    Storage storage = StorageOptions.grpc().build().getService();
```
BenWhitehead: This instance of `Storage` needs to be in a try-with-resource so that it's cleaned up when we're done with it.

samples/snippets/src/main/java/com/example/storage/object/StartAppendableObjectUpload.java
```diff
@@ -0,0 +1,69 @@
...
+    // The ID of your GCS bucket
+    // String bucketName = "your-unique-bucket-name";
+
+    // The ID of your GCS object
+    // String objectName = "your-object-name";
+
+    // The path to the file to upload
+    // String filePath = "path/to/your/file";
+
+    Storage storage = StorageOptions.grpc().build().getService();
```
BenWhitehead: This instance of `Storage` needs to be in a try-with-resource so that it's cleaned up when we're done with it.

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectMultipleRangedRead.java
```diff
@@ -0,0 +1,86 @@
...
+    BlobId blobId = BlobId.of(bucketName, objectName);
+
+    ApiFuture<BlobReadSession> futureBlobReadSession = storage.blobReadSession(blobId);
+
+    try (BlobReadSession blobReadSession = futureBlobReadSession.get(10, TimeUnit.SECONDS)) {
+      RangeSpec rangeSpec1 = RangeSpec.of(offset1, offset1 + length1);
```
BenWhitehead: Second argument should just be the length.
```suggestion
      RangeSpec rangeSpec1 = RangeSpec.of(offset1, length1);
```

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectMultipleRangedRead.java
```diff
@@ -0,0 +1,86 @@
...
+      ReadAsChannel readAsChannelConfig1 =
+          ReadProjectionConfigs.asChannel().withRangeSpec(rangeSpec1);
+
+      RangeSpec rangeSpec2 = RangeSpec.of(offset2, offset2 + length2);
```
BenWhitehead: Second argument should just be the length.
```suggestion
      RangeSpec rangeSpec2 = RangeSpec.of(offset2, length2);
```

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectSingleRangedRead.java
```diff
@@ -0,0 +1,66 @@
...
+    BlobId blobId = BlobId.of(bucketName, objectName);
+
+    ApiFuture<BlobReadSession> futureBlobReadSession = storage.blobReadSession(blobId);
+
+    try (BlobReadSession blobReadSession = futureBlobReadSession.get(10, TimeUnit.SECONDS)) {
+      ByteBuffer buf = ByteBuffer.allocate(length);
+      RangeSpec rangeSpec = RangeSpec.of(offset, offset + length);
```
BenWhitehead: Second argument should just be the length.
```suggestion
      RangeSpec rangeSpec = RangeSpec.of(offset, length);
```

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectReadFullObject.java
```diff
@@ -0,0 +1,68 @@
...
+        while ((bytesRead = channel.read(buffer)) != -1) {
+          if (bytesRead > 0) {
+            buffer.flip();
+            while (buffer.hasRemaining()) {
+              if (buffer.get() == '\n') {
+                newlineCount++;
+              }
+            }
+            buffer.clear();
+          }
```
BenWhitehead: There's no guarantee the object being read has any new line characters. Do you know why this is the scenario that was selected for this sample?

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectReadFullObject.java
```diff
@@ -0,0 +1,68 @@
...
+        }
```
BenWhitehead: This loop can be simplified a bit by using our utility method:
```suggestion
        while (StorageChannelUtils.blockingFillFrom(buffer, channel) != -1) {
          buffer.flip();
          while (buffer.hasRemaining()) {
            if (buffer.get() == '\n') {
              newlineCount++;
            }
          }
          buffer.clear();
        }

```

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectReadFullObject.java
```diff
@@ -0,0 +1,68 @@
...
+
+      long newlineCount = 0;
+      try (ScatteringByteChannel channel = blobReadSession.readAs(readAsChannelConfig)) {
+        ByteBuffer buffer = ByteBuffer.allocate(8 * 1024);
+        int bytesRead = 0;
+        while ((bytesRead = channel.read(buffer)) != -1) {
+          if (bytesRead > 0) {
+            buffer.flip();
+            while (buffer.hasRemaining()) {
+              if (buffer.get() == '\n') {
```
BenWhitehead: This comparison isn't fully type safe. `buffer.get()`[1] returns a `byte` whereas the `'\n'` is a `char`.

If we want to do a single byte comparison, we should change the right hand side to be the byte value of `\n`, `0x0a`.

Alternatively, if we're wanting to the the character comparison, we should change the left hand size to be `buffer.getChar()`[2] (getChar will consume type bytes per call though).

* [1] https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#get--
* [2] https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#getChar--

samples/snippets/src/main/java/com/example/storage/object/AppendableObjectSingleRangedRead.java
```diff
@@ -0,0 +1,66 @@
...
+
+      ReadAsChannel readAsChannelConfig =
+          ReadProjectionConfigs.asChannel().withRangeSpec(rangeSpec);
+
+      try (ScatteringByteChannel channel = blobReadSession.readAs(readAsChannelConfig)) {
+        int bytesRead = 0;
+        while (buf.hasRemaining() && bytesRead != -1) {
+          bytesRead = channel.read(buf);
+        }
+      }
```
BenWhitehead: For a single range demo it might make more sense to use `ReadAsFutureBytes`

samples/snippets/src/main/java/com/example/storage/object/FinalizeAppendableObjectUpload.java
```diff
@@ -0,0 +1,58 @@
...
+    BlobId blobId = BlobId.of(bucketName, objectName);
+    BlobInfo blobInfo = BlobInfo.newBuilder(blobId).build();
+
+    BlobAppendableUploadConfig config =
+        BlobAppendableUploadConfig.of().withCloseAction(CloseAction.FINALIZE_WHEN_CLOSING);
+    BlobAppendableUpload finalizingSession = storage.blobAppendableUpload(blobInfo, config);
+
+    finalizingSession.open().close();
```
BenWhitehead: Just to confirm, we want to create a 0 byte object here? Or, is this supposed to be a takeover to finalize an upload?

If the latter, you need a generation in order to do a takeover. See our integration tests for this scenario for details: https://github.com/googleapis/java-storage/blob/f63c8c770ba16444f3cd81ab220c5f355a22171f/google-cloud-storage/src/test/java/com/google/cloud/storage/ITAppendableUploadTest.java#L237

samples/snippets/src/main/java/com/example/storage/object/ResumeAppendableObjectUpload.java
```diff
@@ -0,0 +1,90 @@
...
+    BlobId blobId = BlobId.of(bucketName, objectName);
+    Blob existingBlob = storage.get(blobId);
+    BlobInfo blobInfoForTakeover = BlobInfo.newBuilder(existingBlob.getBlobId()).build();
+    FileChannel fileChannel = FileChannel.open(Paths.get(filePath));
```
BenWhitehead: This should be moved down into the try-with so the file descriptor is closed.
```suggestion
```

samples/snippets/src/main/java/com/example/storage/object/ResumeAppendableObjectUpload.java
```diff
@@ -0,0 +1,90 @@
...
+        Locale.US,
+        "Resuming upload for %s. Currently uploaded size: %d bytes\n",
+        blobId.toGsUtilUri(),
+        currentObjectSize);
+
+    BlobAppendableUploadConfig config =
+        BlobAppendableUploadConfig.of().withCloseAction(CloseAction.CLOSE_WITHOUT_FINALIZING);
+    BlobAppendableUpload resumeUploadSession =
+        storage.blobAppendableUpload(blobInfoForTakeover, config);
+    try (AppendableUploadWriteableByteChannel channel = resumeUploadSession.open()) {
```
BenWhitehead:
```suggestion
    try (AppendableUploadWriteableByteChannel channel = resumeUploadSession.open();
        FileChannel fileChannel = FileChannel.open(Paths.get(filePath));
    ) {
```

samples/snippets/src/test/java/com/example/storage/ITAppendableObjectSnippets.java
```diff
@@ -0,0 +1,284 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+
+@RunWith(StorageITRunner.class)
+@CrossRun(
+    transports = {Transport.GRPC},
+    backends = {Backend.PROD})
+public class ITAppendableObjectSnippets {
```
BenWhitehead: Can we move these new tests to a followup PR? That will side step the failing build problem for now while we get the samples project access to rapid quota in the zone(s) it runs tests from.

BenWhitehead (changes requested):

samples/snippets/src/test/java/com/example/storage/ITAppendableObjectSnippets.java
nidhiii-27: Done

PR #3279, opened 2025-09-04
chore: update ChunkSegmenter to optionally allow a limit on the number of bytes it should consume

BenWhitehead:
"Child" change after #3278

Update BidiAppendableUnbufferedWritableByteChannel to only attempt to consume as many bytes as are available according to the stream -- this prevents over packing of segments we will for sure never be able to use.

With both this PR and the contents of #3278 our latency now looks like the following, where the `v2.56.0` line represents the workload running using version 2.56.0 and `v2.56.1+bw.2025_09_04_1715` represents the same workload bu using a version build from this PR.
image

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
@@ -18,6 +18,7 @@ package com.google.cloud.storage;
 
 import com.google.cloud.storage.Crc32cValue.Crc32cLengthKnown;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.math.IntMath;
 import com.google.common.primitives.Ints;
@@ -164,7 +165,7 @@ final class ChunkSegmenter {
         }
 
         int numBytesConsumable;
-        if (remaining >= blockSize) {
+        if (remaining >= blockSize && currentBlockPending == blockSize) {
           int blockCount = IntMath.divide(remaining, blockSize, RoundingMode.DOWN);
           numBytesConsumable = blockCount * blockSize;
         } else {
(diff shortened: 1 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
```diff
@@ -97,66 +98,95 @@ ChunkSegment[] segmentBuffers(
...
 
+    outerloop:
     for (int i = offset; i < length; i++) {
       ByteBuffer buffer = bbs[i];
       int remaining;
       while ((remaining = buffer.remaining()) > 0) {
         long overallRemaining = totalRemaining - consumedSoFar;
         if (overallRemaining < blockSize && currentBlockPending == blockSize) {
-          break;
+          break outerloop;
```
BrandonY, line 164: ooo, label break, those're rare!

google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
```diff
@@ -97,66 +98,95 @@ ChunkSegment[] segmentBuffers(
...
-          numBytesConsumable = currentBlockPending;
-          currentBlockPending = blockSize;
         } else {
-          numBytesConsumable = remaining;
-          currentBlockPending = currentBlockPending - remaining;
+          numBytesConsumable = Math.min(remaining, currentBlockPending);
         }
         if (numBytesConsumable <= 0) {
-          continue;
+          break outerloop;
```
BrandonY, line 175: K...now I'm starting to worry a bit about cyclomatic complexity

BrandonY (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
BenWhitehead: Prior to the appendable upload stuff we generally had a pretty firm bulkhead on the number of buffers that would be passed into these methods outside of tests. With the appendable addition we place less emphasis on early buffering in favor of passing things through wherever possible, so if multiple buffers are passed in here, logically the conditions would prevent consuming any bytes once a break from the while takes place, but by breaking the for as well we avoid the cycles performing work that isn't productive.

And, refactoring everything to nested method calls to allow early returns instead of break to label didn't seem worth it to me.

ddelgrosso1 (approved):

google-cloud-storage/src/main/java/com/google/cloud/storage/ChunkSegmenter.java
danielduhh: What's the follow up here?

BenWhitehead: Nothing from my perspective. There isn't a functional difference here between the label break and a separate method with early returns.

PR #3282, opened 2025-09-05
samples(fix): update all samples to apply try-with-resource to clients

BenWhitehead:
All Storage and TransferManager instances now have try-with-resource around them.

* chore: run mvn com.spotify.fmt:fmt-maven-plugin:format

The vast majority of the lines changed in here are whitespace due to additional indentation after using the try-with-resources. I did not change any logic of any of the tests or samples themselves.

### DIFF
--- a/samples/snippets/src/main/java/com/example/storage/ConfigureRetries.java
+++ b/samples/snippets/src/main/java/com/example/storage/ConfigureRetries.java
@@ -60,8 +60,8 @@ public final class ConfigureRetries {
       boolean success = storage.delete(blobId);
 
       System.out.printf(
-          "Deletion of Blob %s completed %s.%n", blobId,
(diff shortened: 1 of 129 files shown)


blakeli0:
I'm just passing by. One caveat of this is that some customers may copy paste the samples without thinking, which would end up creating one client per request and cause other performance issues. We've seen it in the past and actually plan to add update our samples to include both ways. Because in a typical application, these client are probably not created manually with try-with-resource, but created by frameworks like Spring as a long-lived bean.

BenWhitehead:
Understood, however I think the thing that is missing from most stuff today is any hint that the client(s) should be closed when you're done with them.

I'd hope that framework maintainers would be able to recognize a try-with-resource and determine how to fit that into the lifecycle of the abstraction they are implementing.

PR #3287, opened 2025-09-10
fix: Correct diff logic for nested context updates

nidhiii-27:


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITNestedUpdateMaskTest.java
```diff
@@ -109,9 +105,9 @@ public ImmutableList<Param> parameters() {
           new Param("2 keys, modify 1 value (fine)", k1a_k2b, k1z, k1z_k2b),
           new Param("2 keys, modify 1 null (full)", k1a_k2b, k1a_k2null, k1a),
           new Param("2 keys, modify 1 null (fine)", k1a_k2b, k2null, k1a),
-          new Param("1 key, set empty", k1a, empty, null),
+          new Param("1 key, set null", k1a, k1null, null),
```
BenWhitehead: This change results in the same permutation as the following line. I think we should keep it so that setting to an empty value is able to clear the value.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITNestedUpdateMaskTest.java
```diff
@@ -109,9 +105,9 @@ public ImmutableList<Param> parameters() {
...
           new Param("1 key, null key", k1a, k1null, null),
-          new Param("2 keys, set null", k1a_k2b, null, null));
+          new Param("2 keys, set null", k1a_k2b, k1null_k2null, null));
```
BenWhitehead, line 115: For this new scenario, can you add it to the readable table in the comment to this method?

BenWhitehead (changes requested):

PR #3295, opened 2025-09-16
fix: fix appendable upload finalization race condition

BenWhitehead:
When finalizing an appendable upload depending on how quickly gcs is acking bytes, we could run into a case where finish_write: true was sent before all bytes had been enqueued.

Regression introduced in 2.57.0

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiAppendableUnbufferedWritableByteChannel.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/BidiAppendableUnbufferedWritableByteChannel.java
@@ -145,6 +145,7 @@ final class BidiAppendableUnbufferedWritableByteChannel implements UnbufferedWri
     // those bytes to implicitly flag as dirty.
     rewindableContent.flagDirty();
 
+    long remainingAfterPacking = Buffers.totalRemaining(srcs, srcsOffset, srcsLength);
     long bytesConsumed = 0;
     for (int i = 0, len = data.length, lastIdx = len - 1; i < len; i++) {
       ChunkSegment datum = data[i];
@@ -153,7 +154,7 @@ final class BidiAppendableUnbufferedWritableByteChannel implements UnbufferedWri
       boolean appended;
       if (i < lastIdx && !shouldFlush) {
         appended = stream.append(datum);
-      } else if (i == lastIdx && nextWriteShouldFinalize) {
+      } else if (i == lastIdx && remainingAfterPacking == 0 && nextWriteShouldFinalize) {
         appended = stream.appendAndFinalize(datum);
       } else {
         appended = stream.appendAndFlush(datum);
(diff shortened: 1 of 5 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/BidiAppendableUnbufferedWritableByteChannel.java
```diff
@@ -153,7 +154,7 @@ private long internalWrite(ByteBuffer[] srcs, int srcsOffset, int srcsLength) th
       boolean appended;
       if (i < lastIdx && !shouldFlush) {
         appended = stream.append(datum);
-      } else if (i == lastIdx && nextWriteShouldFinalize) {
+      } else if (i == lastIdx && remainingAfterPacking == 0 && nextWriteShouldFinalize) {
```
BenWhitehead, line 157: These two lines are the actual fix

google-cloud-storage/src/main/java/com/google/cloud/storage/BidiUploadState.java
```diff
@@ -827,10 +827,16 @@ protected final void validateCurrentStateIsOneOf(State... allowed) {
...
       checkState(
           finishWriteOffset == -1,
-          "Attempting to append bytes even though finalization has previously been signaled.");
+          "Attempting to append bytes even though finalization has previously been signaled."
+              + " (finishWriteOffset: %s, totalSentBytes: %s, confirmedBytes: %s, size: %s)",
+          finishWriteOffset,
+          totalSentBytes,
+          confirmedBytes,
+          size
+        );
```
BenWhitehead: Adding more info to the diagnostic to make it more useful in the future.

google-cloud-storage/src/test/java/com/google/cloud/storage/BidiAppendableUnbufferedWritableByteChannelTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.rules.TestName;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class BidiAppendableUnbufferedWritableByteChannelTest {
+  private static final Logger LOGGER = LoggerFactory.getLogger(BidiAppendableUnbufferedWritableByteChannelTest.class);
+  @Rule public final TestName testName = new TestName();
+
+  @Test
+  public void appendAndFinalizeOnlyPerformedIfAllBytesConsumed() throws IOException {
```
BenWhitehead, line 42: The new test to verify the fix

google-cloud-storage/src/test/java/com/google/cloud/storage/BidiUploadTest.java
```diff
@@ -1841,58 +1838,6 @@ public <T extends Throwable> void recordError(
       }
     }
 
-    private static BidiStreamingCallable<BidiWriteObjectRequest, BidiWriteObjectResponse>
```
BenWhitehead, line 1844: Moved into BidiUploadTestUtils.java for sharing

danielduhh:
Was this caught somewhere? Is there a bug with context?

BenWhitehead:
> Was this caught somewhere? Is there a bug with context?

b/445496061

PR #3329, opened 2025-10-06
samples: add samples for object contexts

nidhiii-27:
Added samples :
storage_get_object_contexts
storage_list_object_contexts
storage_set_object_contexts

### DIFF
--- a/samples/snippets/src/main/java/com/example/storage/object/GetObjectContexts.java
+++ b/samples/snippets/src/main/java/com/example/storage/object/GetObjectContexts.java
@@ -1,0 +1,36 @@
+package com.example.storage.object;
+
+// [START storage_get_object_contexts]
+
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.BlobInfo.ObjectCustomContextPayload;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import java.util.Map;
+
+public class GetObjectContexts {
+  public static void getObjectMetadata(String projectId, String bucketName, String blobName)
+      throws Exception {
+    // The ID of your GCP project
+    // String projectId = "your-project-id";
+
+    // The ID of your GCS bucket
+    // String bucketName = "your-unique-bucket-name";
+
+    // The ID of your GCS object
+    // String objectName = "your-object-name";
+
+    try (Storage storage =
+        StorageOptions.newBuilder().setProjectId(projectId).build().getService()) {
+
+      Blob blob = storage.get(bucketName, blobName);
+      Map<String, ObjectCustomContextPayload> customContexts = blob.getContexts().getCustom();
+      // Print blob's object contexts
+      System.out.println("\nCustom Contexts:");
+      for (Map.Entry<String, ObjectCustomContextPayload> custom : customContexts.entrySet()) {
+        System.out.println(custom.getKey() + "=" + custom.getValue());
+      }
+    }
+  }
+}
+// [END storage_get_object_contexts]
--- a/samples/snippets/src/main/java/com/example/storage/object/ListObjectContexts.java
+++ b/samples/snippets/src/main/java/com/example/storage/object/ListObjectContexts.java
@@ -1,0 +1,44 @@
+package com.example.storage.object;
+
+// [START storage_list_object_contexts]
+
+import com.google.api.gax.paging.Page;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+
+public class ListObjectContexts {
+  public static void listObjectContexts(String projectId, String bucketName) throws Exception {
+    // The ID of your GCP project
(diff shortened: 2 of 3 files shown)


samples/snippets/src/main/java/com/example/storage/object/GetObjectContexts.java
```diff
@@ -0,0 +1,52 @@
...
+    // String bucketName = "your-unique-bucket-name";
+
+    // The ID of your GCS object
+    // String objectName = "your-object-name";
+
+    try (Storage storage =
+        StorageOptions.newBuilder().setProjectId(projectId).build().getService()) {
+
+      Blob blob = storage.get(bucketName, blobName);
+      Map<String, ObjectCustomContextPayload> customContexts = blob.getContexts().getCustom();
```
BenWhitehead: This isn't null safe. we need to make sure both the returned blob and the `blob.getContexts()` is not null.

samples/snippets/src/main/java/com/example/storage/object/GetObjectContexts.java
```diff
@@ -0,0 +1,52 @@
...
+// [START storage_get_object_contexts]
+
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.BlobInfo.ObjectCustomContextPayload;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import java.util.Map;
+
+public class GetObjectContexts {
+  public static void getObjectMetadata(String projectId, String bucketName, String blobName)
```
BenWhitehead: nit: method is named `getObjectMetadata` but the contents show how to get contexts.

samples/snippets/src/main/java/com/example/storage/object/ListObjectContexts.java
```diff
@@ -0,0 +1,60 @@
...
+       * List any object that that does not have a context with the specified key attached
+       * String filter = "NOT contexts.\"KEY\":*";
+       *
+       * List any object that has a context with the specified key and value attached
+       * String filter = "contexts.\"KEY\"=\"VALUE\"";
+       *
+       * List any object that does not have a context with the specified key and value attached
+       * String filter = "NOT contexts.\"KEY\"=\"VALUE\"";
+       */
+      String key = "your-context-key";
```
BenWhitehead: nit: making this a method parameter will probably make test authoring easier

samples/snippets/src/main/java/com/example/storage/object/ListObjectContexts.java
```diff
@@ -0,0 +1,60 @@
...
+      String filter = "contexts.\"" + key + "\":*";
+
+      System.out.println("Listing objects for bucket: " + bucketName + "with context key: " + key);
+      Page<Blob> blobs = storage.list(bucketName, Storage.BlobListOption.filter(filter));
+      for (Blob blob : blobs.iterateAll()) {
+        System.out.println(blob.getName());
```
BenWhitehead: nit: prefix the printing of the bucket name with some descriptor to put the value into context. Or change to printing `blob.getBlobId().toGsUtilUri()` so it is more obvious an object uri is the value.

samples/snippets/src/main/java/com/example/storage/object/SetObjectContexts.java
```diff
@@ -0,0 +1,81 @@
...
+    try (Storage storage =
+        StorageOptions.newBuilder().setProjectId(projectId).build().getService()) {
+      BlobId blobId = BlobId.of(bucketName, objectName);
+      Blob blob = storage.get(blobId);
+      if (blob == null) {
+        System.out.println("The object " + objectName + " was not found in " + bucketName);
+        return;
+      }
+
+      // Optional: Set a generation-match precondition to avoid potential race
```
BenWhitehead: `s/Optional/Recommended/g`

I think we want to be more firm that we recommend people do this.

samples/snippets/src/main/java/com/example/storage/object/SetObjectContexts.java
```diff
@@ -0,0 +1,81 @@
...
+      // conditions and data corruptions. The request to update returns a 412 error if
+      // the object's generation number does not match your precondition.
+      Storage.BlobTargetOption precondition = Storage.BlobTargetOption.generationMatch();
+
+      String key = "your-context-key";
+      String value = "your-context-value";
```
BenWhitehead: nit: providing these as method parameters will probably make test authoring easier

nidhiii-27: Thank you for the suggestion!

samples/snippets/src/main/java/com/example/storage/object/GetObjectContexts.java
BenWhitehead: We still need to check `blob.getContexts() != null` before calling `getCustom()`

Issue #3348, opened 2025-10-13
Support XML MPU in GCS Java SDK


ShreyasSinha:
Support XML MPU in GCS Java SDK https://cloud.google.com/storage/docs/multipart-uploads

PR #3350, opened 2025-10-14
chore: Refactoring retrier creation from HttpStorageOptions to StorageOptions

ShreyasSinha:

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
@@ -32,6 +32,7 @@ import com.google.cloud.TransportOptions;
 import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.spi.ServiceRpcFactory;
 import com.google.cloud.storage.BlobWriteSessionConfig.WriterFactory;
+import com.google.cloud.storage.Retrying.HttpRetrier;
 import com.google.cloud.storage.Retrying.RetryingDependencies;
 import com.google.cloud.storage.Storage.BlobWriteOption;
 import com.google.cloud.storage.TransportCompatibility.Transport;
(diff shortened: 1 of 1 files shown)


samples/snippets/src/test/java/com/example/storage/ITBucketSnippets.java
```diff
@@ -425,6 +426,7 @@ public void testAddListRemoveBucketIamMembers() throws Throwable {
             .build());
   }
 
+  @Ignore("This test cannot run in an environment with Public Access Prevention enforced.")
```
BenWhitehead: Can we move these `@Ignore`s to a separate PR that is more easily revertable once we figure out the project settings that need to change?

google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
```diff
@@ -68,6 +72,13 @@ public abstract class StorageOptions extends ServiceOptions<Storage, StorageOpti
     VERSION = tmp;
   }
 
+  Retrier createRetrier() {
+    return new HttpRetrier(
+        new DefaultRetrier(
+            OtelStorageDecorator.retryContextDecorator(getOpenTelemetry()),
+            RetryingDependencies.simple(getClock(), getRetrySettings())));
+  }
+
```
BenWhitehead, line 79: I think I initially recommended this, but I think this method is sufficient if it only exists in HttpStorageOptions.

If we did want to keep it here, we should remove the `HttpRetrier` decoration, as that can only be useful for HttpStorageOptions based instances.

ShreyasSinha: Agreed on HttpRetrier decoration. Have a slight preference for keeping it because of readability so keeping it as is.

samples/snippets/src/test/java/com/example/storage/ITBucketSnippets.java
ShreyasSinha: Done.

google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
```diff
@@ -408,14 +406,7 @@ public Storage create(StorageOptions options) {
...
           StorageImpl storage =
-              new StorageImpl(
-                  httpStorageOptions,
-                  factory,
-                  new HttpRetrier(
-                      new DefaultRetrier(
-                          OtelStorageDecorator.retryContextDecorator(otel),
-                          RetryingDependencies.simple(
-                              options.getClock(), options.getRetrySettings()))));
+              new StorageImpl(httpStorageOptions, factory, options.createRetrier());
```
BenWhitehead: We need to keep the HttpRetrier decorator for the HttpStorageOptions usage.
```suggestion
              new StorageImpl(httpStorageOptions, factory, new HttpRetrier(options.createRetrier()));
```

BenWhitehead (changes requested):

BenWhitehead:

PR #3356, opened 2025-10-21
feat: add preview MultipartUploadClient#createMultipartUpload

ShreyasSinha:
1) Adding model classes for CreateMultipartUpload request and response.
2) Setting Client for Multipart Upload.
3) Introducing data class for ObjectLockMode https://cloud.google.com/storage/docs/object-lock
4) Implemented Object parser for XML MPU.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/HttpStorageOptions.java
@@ -32,6 +32,7 @@ import com.google.cloud.TransportOptions;
 import com.google.cloud.http.HttpTransportOptions;
 import com.google.cloud.spi.ServiceRpcFactory;
 import com.google.cloud.storage.BlobWriteSessionConfig.WriterFactory;
+import com.google.cloud.storage.Retrying.HttpRetrier;
 import com.google.cloud.storage.Retrying.RetryingDependencies;
 import com.google.cloud.storage.Storage.BlobWriteOption;
 import com.google.cloud.storage.TransportCompatibility.Transport;
@@ -406,7 +407,8 @@ public class HttpStorageOptions extends StorageOptions {
           }
           WriterFactory factory = blobWriteSessionConfig.createFactory(clock);
           StorageImpl storage =
-              new StorageImpl(httpStorageOptions, factory, options.createRetrier());
+              new StorageImpl(
+                  httpStorageOptions, factory, new HttpRetrier(options.createRetrier()));
           return OtelStorageDecorator.decorate(storage, otel, Transport.HTTP);
         } catch (IOException e) {
           throw new IllegalStateException(
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageOptions.java
@@ -27,7 +27,6 @@ import com.google.cloud.storage.HttpStorageOptions.HttpStorageDefaults;
 import com.google.cloud.storage.HttpStorageOptions.HttpStorageFactory;
 import com.google.cloud.storage.HttpStorageOptions.HttpStorageRpcFactory;
 import com.google.cloud.storage.Retrying.DefaultRetrier;
-import com.google.cloud.storage.Retrying.HttpRetrier;
 import com.google.cloud.storage.Retrying.Retrier;
 import com.google.cloud.storage.Retrying.RetryingDependencies;
 import com.google.cloud.storage.Storage.BlobWriteOption;
@@ -73,10 +72,9 @@ public abstract class StorageOptions extends ServiceOptions<Storage, StorageOpti
   }
 
   Retrier createRetrier() {
-    return new HttpRetrier(
-        new DefaultRetrier(
-            OtelStorageDecorator.retryContextDecorator(getOpenTelemetry()),
-            RetryingDependencies.simple(getClock(), getRetrySettings())));
+    return new DefaultRetrier(
+        OtelStorageDecorator.retryContextDecorator(getOpenTelemetry()),
+        RetryingDependencies.simple(getClock(), getRetrySettings()));
   }
 
   /**


google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
```diff
@@ -0,0 +1,284 @@
...
+import com.google.common.base.MoreObjects;
+import java.util.Date;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a request to initiate a multipart upload. This class holds all the necessary
+ * information to create a new multipart upload session.
+ */
+public final class CreateMultipartUploadRequest {
```
BenWhitehead, line 32: You'll want to annotate all of these new classes and public methods with `@BetaApi` to signal that they are pre-GA and therefore could break.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
```diff
@@ -0,0 +1,284 @@
...
+import com.google.cloud.storage.StorageClass;
+import com.google.common.base.MoreObjects;
+import java.util.Date;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a request to initiate a multipart upload. This class holds all the necessary
+ * information to create a new multipart upload session.
+ */
```
BenWhitehead, line 30: Possibly something to do in a follow-up PR, but it's a good idea to add an `@since 2.60.0` with a comment to the java docs, so that it's obvious what version it was added in, and if there is any thing to be aware of with the addition.

See the BlobReadSession javadocs for an example of this: https://github.com/googleapis/java-storage/blob/e56e45a658fcd55b10635df4ebc776ed6689de8b/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobReadSession.java#L31

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
```diff
@@ -0,0 +1,284 @@
...
+    this.kmsKeyName = builder.kmsKeyName;
+    this.objectLockMode = builder.objectLockMode;
+    this.objectLockRetainUntilDate = builder.objectLockRetainUntilDate;
+  }
+
+  public String bucket() {
+    return bucket;
+  }
+
+  public String key() {
```
BenWhitehead, line 72: `key` isn't really a GCS~ism. We'll at a minimum want to document that this is the name of the object, and probably link to the docs outlining what are valid names.

https://cloud.google.com/storage/docs/objects#naming

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
```diff
@@ -0,0 +1,284 @@
...
+ * information to create a new multipart upload session.
+ */
+public final class CreateMultipartUploadRequest {
+  private final String bucket;
+  private final String key;
+  private final PredefinedAcl cannedAcl;
+  private final String contentType;
+  private final Map<String, String> metadata;
+  private final StorageClass storageClass;
+  private final Date customTime;
```
BenWhitehead: `java.util.Date` is pretty terrible in most respects.

Can we change this to be a `java.time.OffsetDateTime`? Check out BlobInfo for some examples of how we use OffsetDateTime currently.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/ObjectLockMode.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.api.core.ApiFunction;
+import com.google.cloud.StringEnumType;
+import com.google.cloud.StringEnumValue;
+
+/**
+ * Represents the object lock mode. See <a
+ * href="https://cloud.google.com/storage/docs/object-lock">https://cloud.google.com/storage/docs/object-lock</a>
+ * for details.
+ */
+public final class ObjectLockMode extends StringEnumValue {
```
BenWhitehead, line 29: This should probably move into the the `com.google.cloud.storage.multipartupload.model` package.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
```diff
@@ -0,0 +1,284 @@
...
+        .add("metadata", metadata)
+        .add("storageClass", storageClass)
+        .add("customTime", customTime)
+        .add("kmsKeyName", kmsKeyName)
+        .add("objectLockMode", objectLockMode)
+        .add("objectLockRetainUntilDate", objectLockRetainUntilDate)
+        .toString();
+  }
+
+  public static Builder builder() {
```
BenWhitehead, line 205: nit: style consistency


```suggestion
  public static Builder newBuilder() {
```

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
```diff
@@ -0,0 +1,284 @@
...
+    return new Builder();
+  }
+
+  public static class Builder {
```
BenWhitehead: nit
```suggestion
  public static final class Builder {
```

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TimeZone;
+
+/**
+ * This class is an implementation of {@link MultipartUploadClient} that uses the Google Cloud
+ * Storage XML API to perform multipart uploads.
+ */
+public final class MultipartUploadClientImpl extends MultipartUploadClient {
```
BenWhitehead: class should be package private
```suggestion
final class MultipartUploadClientImpl extends MultipartUploadClient {
```

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+
+  private final MultipartUploadHttpRequestManager httpRequestManager;
+  private final XmlMapper xmlMapper;
+  private final HttpStorageOptions options;
+  private final Retrier retrier;
+  private final URI uri;
+
+  public MultipartUploadClientImpl(
```
BenWhitehead: constructor should be package private
```suggestion
  MultipartUploadClientImpl(
```

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
```diff
@@ -0,0 +1,51 @@
...
+      Map<String, String> extensionHeaders)
+      throws IOException {
+    HttpRequest httpRequest =
+        requestFactory.buildPostRequest(
+            new GenericUrl(uri), new ByteArrayContent(contentType, new byte[0]));
+    httpRequest.getHeaders().setContentType(contentType);
+    for (Map.Entry<String, String> entry : extensionHeaders.entrySet()) {
+      httpRequest.getHeaders().set(entry.getKey(), entry.getValue());
+    }
+    httpRequest.setThrowExceptionOnExecuteError(false);
```
BenWhitehead: For the operations implemented for MPU, I think we want to leave this as it's default `true`. The reason being, when a request is not succssful the library will take care of unpacking any failure response and turning it into a `BaseHttpServiceException` with the failure reason, status code and etc.

Additionally, our retrier knows how to take a BaseHttpServiceException automatically to evaluate for retryability.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+
+    HttpResponse response =
+        httpRequestManager.sendCreateMultipartUploadRequest(
+            createUri, contentType, request, getExtensionHeadersForCreateMultipartUpload(request));
+
+    if (!response.isSuccessStatusCode()) {
+      String error = response.parseAsString();
+      throw new RuntimeException(
+          "Failed to initiate upload: " + response.getStatusCode() + " " + error);
+    }
```
BenWhitehead: We should remove this after update the manager to leave `setThrowExceptionOnExecuteError = true`

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+      extensionHeaders.put("x-goog-encryption-kms-key-name", request.getKmsKeyName());
+    }
+    // x-goog-object-lock-mode and x-goog-object-lock-retain-until-date should be specified together
+    // Refer: https://cloud.google.com/storage/docs/xml-api/post-object-multipart#request_headers
+    if (request.getObjectLockMode() != null && request.getObjectLockRetainUntilDate() != null) {
+      extensionHeaders.put("x-goog-object-lock-mode", request.getObjectLockMode().toString());
+      extensionHeaders.put(
+          "x-goog-object-lock-retain-until-date",
+          toRfc3339String(request.getObjectLockRetainUntilDate()));
+    }
```
BenWhitehead: Is it an error if the customer doesn't provide both headers? Does the server validate they are both set?

If so, I think it's probably better to plumb them individually and leave the validation to the server so we don't silently not send a header a customer provides.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+  private Map<String, String> getExtensionHeadersForCreateMultipartUpload(
+      CreateMultipartUploadRequest request) {
+    Map<String, String> extensionHeaders = getGenericExtensionHeader();
+    if (request.getCannedAcl() != null) {
+      extensionHeaders.put("x-goog-acl", request.getCannedAcl().toString());
+    }
+    if (request.getMetadata() != null) {
+      for (Map.Entry<String, String> entry : request.getMetadata().entrySet()) {
+        if (entry.getKey() != null || entry.getValue() != null) {
+          extensionHeaders.put("x-goog-meta-" + entry.getKey(), entry.getValue());
```
BenWhitehead: http header values should be URL encoded to ensure any reserved characters are not interpreted/impacting to request parsing.

Since the customer is able to provide arbitrary values here we need to encode them to make them http safe.

We should check with someone on GCS server side to see if they can handle encoded header keys as well.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+
+  private Map<String, String> getGenericExtensionHeader() {
+    Map<String, String> extensionHeaders = new HashMap<>();
+    if (options.getClientLibToken() != null) {
+      extensionHeaders.put("x-goog-api-client", options.getClientLibToken());
+    }
+    if (options.getProjectId() != null) {
+      extensionHeaders.put("x-goog-user-project", options.getProjectId());
+    }
+    extensionHeaders.put("Date", getRfc1123Date());
```
BenWhitehead: The json api doesn't require this to be passed, does the XML api? If not, we can probably remove this and the corresponding logic.

BenWhitehead: If we do want to keep it, we should refactor things to use the ApiClock from the HttpStorageOptions to get the current time, rather than using `ZonedDateTime.now()` directly.

In general, any time you're programming against time you should have a thing that gives you time so you aren't tied to the system clock.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+  public CreateMultipartUploadResponse createMultipartUpload(CreateMultipartUploadRequest request)
+      throws IOException {
+    String encodedBucket = encode(request.bucket());
+    String encodedKey = encode(request.key());
+    String resourcePath = "/" + encodedBucket + "/" + encodedKey;
+    String createUri = uri.toString() + resourcePath + "?uploads";
+
+    String contentType;
+    if (request.getContentType() == null) {
+      contentType = "application/x-www-form-urlencoded";
```
BenWhitehead: Why do we need to default to this value?

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+    } else {
+      try {
+        contentType = MediaType.parse(request.getContentType()).toString();
+      } catch (IllegalArgumentException e) {
+        throw new IllegalArgumentException(
+            "Invalid Content-Type header provided: " + request.getContentType(), e);
+      }
```
BenWhitehead: I prefer we not perform the validation here, instead
1) send the value to GCS as is if non-null and let GCS validate since it's the server and is the ultimate authority
2) instead validate when the request.setContentType(String) method is called.

There have been issues in the past due to client side validation not working as expected, and we generally don't validate anything that the server would validate itself.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,153 @@
...
+    extensionHeaders.put("Date", getRfc1123Date());
+    return extensionHeaders;
+  }
+
+  public CreateMultipartUploadResponse createMultipartUpload(CreateMultipartUploadRequest request)
+      throws IOException {
+    String encodedBucket = encode(request.bucket());
+    String encodedKey = encode(request.key());
+    String resourcePath = "/" + encodedBucket + "/" + encodedKey;
+    String createUri = uri.toString() + resourcePath + "?uploads";
```
BenWhitehead: I think we've got a bit of split brain happening between this class and MultipartUploadHttpRequestManager. This class is doing url encoding and construction, header processing and response parsing. But, Manager is doing some http request initialization, request execution and response returning.

I'd expect the logic of building the actual http request to happen over in manager, while this class would focus on any retry wrapping and additional info that would need to be provided in order for the manager to actually process an individual request. (for example, manager doesn't do anything with the provided CreateMultipartUploadRequest provided to it).

What is the intended delineation of responsibility between these two classes?

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
```diff
@@ -0,0 +1,51 @@
...
+    return httpRequest.execute();
```
BenWhitehead: We can plumb the xml deserializatin into the actual http response handling so that the higher level need to care about it outside of making sure an xml parsing failure is a retryable error.

prototype code to do this:
```suggestion
    httpRequest.setParser(objectParser);
    HttpResponse response = httpRequest.execute();
    return response.parseAs(CreateMultipartUploadResponse.class);
  }

  @VisibleForTesting
  static final class XmlObjectParser implements ObjectParser {
    private final XmlMapper xmlMapper;

    private XmlObjectParser(XmlMapper xmlMapper) {
      this.xmlMapper = xmlMapper;
    }

    @Override
    public <T> T parseAndClose(InputStream in, Charset charset, Class<T> dataClass)
        throws IOException {
      return parseAndClose(new InputStreamReader(in, charset), dataClass);
    }

    @Override
    public Object parseAndClose(InputStream in, Charset charset, Type dataType) throws IOException {
      throw new UnsupportedOperationException(
          "XmlObjectParse#" + CrossTransportUtils.fmtMethodName("parseAndClose", InputStream.class,
              Charset.class, Type.class));
    }

    @Override
    public <T> T parseAndClose(Reader reader, Class<T> dataClass) throws IOException {
      try (Reader r = reader) {
        return xmlMapper.readValue(r, dataClass);
      }
    }

    @Override
    public Object parseAndClose(Reader reader, Type dataType) throws IOException {
      throw new UnsupportedOperationException(
          "XmlObjectParse#" + CrossTransportUtils.fmtMethodName("parseAndClose", Reader.class,
              Type.class));
    }
  }

```

google-cloud-storage/src/test/java/com/google/cloud/storage/MultipartUploadClientImplTest.java
```diff
@@ -0,0 +1,192 @@
...
+    field.set(multipartUploadClient, httpRequestManager);
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    mocks.close();
+  }
+
+  @Test
+  public void testCreateMultipartUpload() throws IOException {
```
BenWhitehead: Because there isn't a clean line between the client and the manager, your unit tests here need to worry about mocking http specifics.

If HTTP is isolated to the manager, none of your client unit tests will ever need to worry about actual http requests/responses.

google-cloud-storage/src/test/java/com/google/cloud/storage/MultipartUploadClientImplTest.java
```diff
@@ -0,0 +1,192 @@
...
+        .thenReturn(httpResponse);
+
+    RuntimeException exception =
+        assertThrows(
+            RuntimeException.class, () -> multipartUploadClient.createMultipartUpload(request));
+    assertThat(exception.getMessage())
+        .isEqualTo("Failed to initiate upload: 500 Internal Server Error");
+  }
+
+  private HttpResponse createHttpResponse(int statusCode, String content) throws IOException {
```
BenWhitehead: We have a fixture to spin up an actual full http server for doing tests where we want to validate request/response handling. Some tests that use it can be seen here: https://github.com/googleapis/java-storage/blob/e56e45a658fcd55b10635df4ebc776ed6689de8b/google-cloud-storage/src/test/java/com/google/cloud/storage/ITJsonResumableSessionPutTaskTest.java#L206-L244

In general, these fake server tests end up being much easier to maintain and more comprehensive compared to doing mocks at the transport. Forcing the full http stack means that the full stack is engaged and things are hidden.

The fakes are also, IMO, easier to author an reason about because they are actual request handlers instead of mocks with mock lifecycles.

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
ShreyasSinha: Yes, its in my radar.

I'll add it in a followup pr closer to release.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/ObjectLockMode.java
ShreyasSinha: ObjectLockMode is not specific to MPU but a general object property. Refer: https://cloud.google.com/storage/docs/object-lock.

IMO it should stay out of MPU specific model.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
ShreyasSinha: I had just added it for public APIs. Added it now for classes as well.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
ShreyasSinha: Prefer to keep it as builder() to keep it consistent with aws sdk https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/CreateMultipartUploadRequest.html#builder() similar reasoning for all other fields as well.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
ShreyasSinha: Done

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: Done

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: Done

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: It was a requirement when I was directly using HTTPUrlConnection, looks like its getting automatically added with HttpRequestFactory (similar to auth stuff). Removing for now.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: The idea was for MultipartUploadClientImpl to be responsible for extracting all arguments from CreateMultipartUpload request and passing it to manager to perform actual httpRequest stuff.

MultipartUploadHttpRequestManager should be responsible for doing the actual requestFactory and creating httpRequest stuff.

I am fine with moving extraction part to manager as well, as we are anyways sending the request for some of the direct headers.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: Handled by server side as well we can remove.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: Verified that its been handled at server as well, removing from here.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: Verified server side validation is present removing the client side validation.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
ShreyasSinha: Done

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/CreateMultipartUploadRequest.java
ShreyasSinha: I had added javadocs for Builder (setters) earlier adding it for getters as well.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: Done

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: Adding a todo for now, will check with GCS and add encoding.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
ShreyasSinha: Nice refactor. Thanks for this suggestion. Done

google-cloud-storage/src/test/java/com/google/cloud/storage/MultipartUploadClientImplTest.java
ShreyasSinha: Comment is obsolete now, refactored test suite.

google-cloud-storage/src/test/java/com/google/cloud/storage/MultipartUploadClientImplTest.java
ShreyasSinha: I prefer fake over mocks. Thanks for bringing this fake to my notice.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/ObjectLockMode.java
BenWhitehead: We already have BlobInfo.Retention.Mode to represent the values for JSON.

The `GOVERNANCE` and `COMPLIANCE` are specific to XML, and as such should be with the xml model classes.

google-cloud-storage/src/main/java/com/google/cloud/storage/XmlObjectParser.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @VisibleForTesting
+  public XmlObjectParser(XmlMapper xmlMapper) {
+    this.xmlMapper = xmlMapper;
+  }
+
+  @Override
+  public <T> T parseAndClose(InputStream in, Charset charset, Class<T> dataClass)
+      throws IOException {
+    return xmlMapper.readValue(in, dataClass);
```
BenWhitehead: This is throwing away the charset argument. Since readValue doesn't allow passing the charse, these methods should construct a `Reader` with the provided `Charset` and delegate to the respective `parseAndClose(Reader, *)` method. See my comment where I did this: https://github.com/googleapis/java-storage/pull/3356#discussion_r2449930921

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
```diff
@@ -0,0 +1,375 @@
...
+  private HttpStorageOptions httpStorageOptions;
+
+  @Rule public final TemporaryFolder temp = new TemporaryFolder();
+
+  @Before
+  public void setUp() throws Exception {
+    multipartUploadHttpRequestManager =
+        new MultipartUploadHttpRequestManager(
+            transport.createRequestFactory(), new XmlObjectParser(new XmlMapper()));
+    httpStorageOptions = HttpStorageOptions.newBuilder().setProjectId("test-project").build();
```
BenWhitehead: probably also want to call `.setCredentials(NoCredentials.getInstance())` to avoid this trying to find credentials from the environment.

See how we do it for testbench for an example: https://github.com/googleapis/java-storage/blob/763be91bd8532b4fc8c3a3c2248509c0c5256086/google-cloud-storage/src/test/java/com/google/cloud/storage/it/runner/registry/BackendResources.java#L105

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
```diff
@@ -0,0 +1,375 @@
...
+              CreateMultipartUploadResponse.builder()
+                  .bucket("test-bucket")
+                  .key("test-key")
+                  .uploadId("test-upload-id")
+                  .build();
+          ByteBuf buf = Unpooled.wrappedBuffer(gson.toByteArray(response));
+
+          DefaultFullHttpResponse resp =
+              new DefaultFullHttpResponse(req.protocolVersion(), OK, buf);
+          resp.headers().set(CONTENT_TYPE, "application/xml; charset=utf-8");
```
BenWhitehead, line 83: `gson.toByteArray` will produce json not xml, yet we're setting a content-type of xml here. I think you want to use `new XmlMapper().writerFor(CreateMultipartUploadResponse.class).writeValueAsBytes(m)` to get actual xml bytes not json.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
```diff
@@ -0,0 +1,375 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+
+@RunWith(StorageITRunner.class)
+@SingleBackend(Backend.PROD)
+@ParallelFriendly
+public final class ITMultipartUploadHttpRequestManagerTest {
+  private static final GsonFactory gson = GsonFactory.getDefaultInstance();
```
BenWhitehead: use XmlMapper instead of gson throughout these tests.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
```diff
@@ -0,0 +1,375 @@
...
+              .build();
+
+      multipartUploadHttpRequestManager.sendCreateMultipartUploadRequest(
+          endpoint, request, httpStorageOptions);
+    }
+  }
+
+  @Test
+  public void sendCreateMultipartUploadRequest_withCustomTime() throws Exception {
+    HttpRequestHandler handler =
```
BenWhitehead, line 340: good job on these tests. In general I think they're much easier to reason about and read.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/ObjectLockMode.java
ShreyasSinha: In that case, sure I'll move it.

google-cloud-storage/src/main/java/com/google/cloud/storage/XmlObjectParser.java
ShreyasSinha: Done

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
ShreyasSinha: Done

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
ShreyasSinha: All the json stuff is ghost of something I was trying out earlier, I meant to clean it up. Done

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
ShreyasSinha: Refer previous comment. Done.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
```diff
@@ -0,0 +1,119 @@
...
+
+    String encodedBucket = encode(request.bucket());
+    String encodedKey = encode(request.key());
+    String resourcePath = "/" + encodedBucket + "/" + encodedKey;
+    String createUri = uri.toString() + resourcePath + "?uploads";
+
+    HttpRequest httpRequest =
+        requestFactory.buildPostRequest(
+            new GenericUrl(createUri), new ByteArrayContent(request.getContentType(), new byte[0]));
+    httpRequest.getHeaders().setContentType(request.getContentType());
```
BenWhitehead: nit: these two lines will both attempt to set the `content-type` header.

If you want to set the header manually, like line 57 does, you should pass `new EmptyContent()` to `buildPostRequest`. OR, keep line 56 as is and remove 57.

google-cloud-storage/pom.xml
```diff
@@ -19,6 +19,14 @@
...
   </properties>
   <dependencies>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-xml</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
```
BenWhitehead: The dependency build is failing because this is declared, but not used. Can we remove this?

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2023 Google LLC
```
BenWhitehead: Can you update all copyright years to be 2025? Not sure where this one got 2023, and I saw a 2024 in one of the other files too.

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
ShreyasSinha: Done

google-cloud-storage/pom.xml
ShreyasSinha: Done

BenWhitehead:
No more requested changes for this PR at this time, but we do have some followups later. Holding off merging for now.

BenWhitehead:

PR #3363, opened 2025-10-23
chore: Adding implmentation of RequestBody and RewindableContent

ShreyasSinha:
1. Adding RequestBody for MPU which internally uses RewindableContent.
2. Implementing checksum related functions in RewindableContent.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
@@ -24,6 +24,8 @@ import com.google.cloud.storage.multipartupload.model.CreateMultipartUploadReque
 import com.google.cloud.storage.multipartupload.model.CreateMultipartUploadResponse;
 import com.google.cloud.storage.multipartupload.model.ListPartsRequest;
 import com.google.cloud.storage.multipartupload.model.ListPartsResponse;
+import com.google.cloud.storage.multipartupload.model.UploadPartRequest;
+import com.google.cloud.storage.multipartupload.model.UploadPartResponse;
 import java.io.IOException;
 import java.net.URI;
 
@@ -73,6 +75,14 @@ final class MultipartUploadClientImpl extends MultipartUploadClient {
     return retrier.run(
         retryAlgorithmManager.idempotent(),
         () -> httpRequestManager.sendAbortMultipartUploadRequest(uri, request),
+        Decoder.identity());
+  }
+
+  @Override
+  public UploadPartResponse uploadPart(UploadPartRequest request, RequestBody requestBody) {
+    return retrier.run(
+        retryAlgorithmManager.idempotent(),
+        () -> httpRequestManager.sendUploadPartRequest(uri, request, requestBody),
         Decoder.identity());
   }
 }
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
@@ -34,6 +34,8 @@ import com.google.cloud.storage.multipartupload.model.CreateMultipartUploadReque
 import com.google.cloud.storage.multipartupload.model.CreateMultipartUploadResponse;
 import com.google.cloud.storage.multipartupload.model.ListPartsRequest;
 import com.google.cloud.storage.multipartupload.model.ListPartsResponse;
+import com.google.cloud.storage.multipartupload.model.UploadPartRequest;
+import com.google.cloud.storage.multipartupload.model.UploadPartResponse;
 import com.google.common.base.StandardSystemProperty;
 import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
@@ -112,6 +114,23 @@ final class MultipartUploadHttpRequestManager {
     return httpRequest.execute().parseAs(AbortMultipartUploadResponse.class);
   }
 
+  public UploadPartResponse sendUploadPartRequest(
+      URI uri, UploadPartRequest request, RequestBody requestBody) throws IOException {
+    String encodedBucket = urlEncode(request.bucket());
+    String encodedKey = urlEncode(request.key());
+    String resourcePath = "/" + encodedBucket + "/" + encodedKey;
+    String queryString =
+        "?partNumber=" + request.partNumber() + "&uploadId=" + urlEncode(request.uploadId());
+    String uploadUri = uri.toString() + resourcePath + queryString;
+    HttpRequest httpRequest =
+        requestFactory.buildPutRequest(new GenericUrl(uploadUri), requestBody.getContent());
+    httpRequest.getHeaders().putAll(headerProvider.getHeaders());
+    addHeadersForUploadPart(requestBody, httpRequest.getHeaders());
+    httpRequest.setParser(objectParser);
+    httpRequest.setThrowExceptionOnExecuteError(true);
+    return httpRequest.execute().parseAs(UploadPartResponse.class);
+  }
+
   static MultipartUploadHttpRequestManager createFrom(HttpStorageOptions options) {
     Storage storage = options.getStorageRpcV1().getStorage();
     ImmutableMap.Builder<String, String> stableHeaders =
@@ -131,6 +150,10 @@ final class MultipartUploadHttpRequestManager {
         storage.getRequestFactory(),
         new XmlObjectParser(new XmlMapper()),
         options.getMergedHeaderProvider(FixedHeaderProvider.create(stableHeaders.build())));
+  }
+
+  private void addHeadersForUploadPart(RequestBody requestBody, HttpHeaders headers) {
+    headers.put("x-goog-hash", "crc32c=" + requestBody.getContent().getCrc32c());
   }
 
   private void addHeadersForCreateMultipartUpload(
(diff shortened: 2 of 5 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
```diff
@@ -165,6 +183,43 @@ long writeTo(GatheringByteChannel gbc) throws IOException {
...
+        while (in.read(byteBuffer) != -1) {
+          byteBuffer.flip();
+          md.update(byteBuffer);
+          byteBuffer.clear();
+        }
+        return Base64.getEncoder().encodeToString(md.digest());
+      } catch (NoSuchAlgorithmException e) {
+        throw new IOException(e);
+      }
+    }
```
BenWhitehead: Do we have to send MD5 to gcs, or can we send just the crc32c?

There's not much use in computing both.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
```diff
@@ -260,5 +315,30 @@ void rewindTo(long offset) {
...
+    @Override
+    String getCrc32c() throws IOException {
+      com.google.common.hash.Hasher crc32cHasher = Hashing.crc32c().newHasher();
+      for (ByteBuffer buffer : buffers) {
+        crc32cHasher.putBytes(buffer.duplicate());
+      }
+      int crc32cInt = crc32cHasher.hash().asInt();
+      ByteBuffer bb = ByteBuffer.allocate(4);
+      bb.putInt(crc32cInt);
+      return Base64.getEncoder().encodeToString(bb.array());
```
BenWhitehead: This should use our existing crc32c infrastructure and options.
```suggestion
      GuavaHasher hasher;
      {
        Hasher defaultHasher = Hasher.defaultHasher();
        if (defaultHasher instanceof NoOpHasher) {
          return null;
        } else {
          hasher = Hasher.enabled();
        }
      }
      Crc32cLengthKnown cumulative = Crc32cValue.zero();
      for (ByteBuffer buffer : buffers) {
        cumulative = cumulative.concat(hasher.hash(buffer::duplicate));
      }
      return Utils.crc32cCodec.encode(cumulative.getValue());
```

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
```diff
@@ -111,6 +119,16 @@ protected void rewindTo(long offset) {}
...
+
+    @Override
+    String getMd5() throws IOException {
+      return "";
+    }
+
+    @Override
+    String getCrc32c() throws IOException {
+      return "";
+    }
```
BenWhitehead: These values aren't correct. Empty bytes do have md5 and crc32c values that aren't empty string.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
```diff
@@ -55,6 +59,10 @@ public final boolean retrySupported() {
     return false;
   }
 
+  abstract String getMd5() throws IOException;
+
+  abstract String getCrc32c() throws IOException;
```
BenWhitehead: I'm not sure this is the best place for this logic to live. Especially in the case of the File based one, the entire file would have to be read from disk before processing could proceed. gRPC has streaming checksum support, and JSON should be getting it soon as well.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
```diff
@@ -30,7 +31,10 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
+import java.util.Base64; // Changed from com.google.api.client.util.Base64
```
BenWhitehead: This line is new, not changed from a previous value.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
```diff
@@ -223,7 +278,7 @@ long writeTo(GatheringByteChannel gbc) throws IOException {
     void rewindTo(long offset) {
       Preconditions.checkArgument(
           offset <= totalLength,
-          "provided offset must be less than or equal to totalLength (%s <= %s)",
+          "provided offset must be less than orequal to totalLength (%s <= %s)",
```
BenWhitehead: revert this please

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
```diff
@@ -520,11 +520,11 @@ public void sendAbortMultipartUploadRequest_error() throws Exception {
...
-          assertThrows(
-              HttpResponseException.class,
-              () ->
-                  multipartUploadHttpRequestManager.sendAbortMultipartUploadRequest(
-                      endpoint, request, httpStorageOptions));
+      assertThrows(
+          HttpResponseException.class,
+          () ->
+              multipartUploadHttpRequestManager.sendAbortMultipartUploadRequest(
+                  endpoint, request, httpStorageOptions));
```
BenWhitehead: Not related to this change, please revert here.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadClientImpl.java
```diff
@@ -75,4 +77,12 @@ public AbortMultipartUploadResponse abortMultipartUpload(AbortMultipartUploadReq
         () -> httpRequestManager.sendAbortMultipartUploadRequest(uri, request),
         Decoder.identity());
   }
+
+  @Override
+  public UploadPartResponse uploadPart(UploadPartRequest request, RequestBody requestBody) {
+    return retrier.run(
+        retryAlgorithmManager.idempotent(),
+        () -> httpRequestManager.sendUploadPartRequest(uri, request, requestBody),
```
BenWhitehead, line 85: This needs to do dirty state tracking, and to actually rewind the content before a followup invocation. See https://github.com/googleapis/java-storage/blob/763be91bd8532b4fc8c3a3c2248509c0c5256086/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableSession.java#L62-L76 for how this is done for json resumable uploads.

Additionally, this should pass the RewindableContent to `sendUploadPartRequest` so that this method manages the lifecycle of it and integrates correctly with the retries.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
```diff
@@ -55,6 +58,8 @@ public final boolean retrySupported() {
     return false;
   }
 
+  abstract String getCrc32c();
```
BenWhitehead, line 61: nit: return Crc32cLengthKnown and let the consumer of this value own the logic of encoding the value for the respected use.

For example, in grpc, crc32c values are `int`s not base64 encoded 4-byte big-endian representation of an unsigned 32-bit integer.

google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
```diff
@@ -133,6 +152,10 @@ static MultipartUploadHttpRequestManager createFrom(HttpStorageOptions options)
         options.getMergedHeaderProvider(FixedHeaderProvider.create(stableHeaders.build())));
   }
 
+  private void addHeadersForUploadPart(RequestBody requestBody, HttpHeaders headers) {
+    headers.put("x-goog-hash", "crc32c=" + requestBody.getContent().getCrc32c());
+  }
```
BenWhitehead, line 157: nit: one-line logic-less helper methods don't help

Is there a plan to add more in here?

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
ShreyasSinha: We can do without MD5 for now.
Will add it later if we get a strong pushback from Salesforce.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
ShreyasSinha: WIP changes. It was a draft PR.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
ShreyasSinha: Let me raise this PR along with UploadParts changes only, it would make it easierin terms of review.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
ShreyasSinha: WIP changes. It was a draft PR.

google-cloud-storage/src/main/java/com/google/cloud/storage/RewindableContent.java
ShreyasSinha: WIP changes. It was a draft PR.

google-cloud-storage/src/test/java/com/google/cloud/storage/ITMultipartUploadHttpRequestManagerTest.java
ShreyasSinha: WIP changes. It was a draft PR.

PR #3365, opened 2025-10-27
fix: add new system property (com.google.cloud.storage.grpc.bound_token) to allow disabling bound token use with grpc

rockspore:
Inspired by https://github.com/googleapis/java-bigtable/pull/2695

The token issuing side is still behind so hopefully it's not too late to add this knob out of extra caution. Users will be able to use it to turn off this flow if anything goes wrong.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
@@ -136,7 +136,8 @@ public final class GrpcStorageOptions extends StorageOptions
   private static final String DEFAULT_HOST = "https://storage.googleapis.com";
   // If true, disable the bound-token-by-default feature for DirectPath.
   private static final boolean DIRECT_PATH_BOUND_TOKEN_DISABLED =
-Boolean.parseBoolean(System.getProperty("com.google.cloud.storage.grpc.bound_token", "false"));
+      Boolean.parseBoolean(
+          System.getProperty("com.google.cloud.storage.grpc.bound_token", "false"));
 
   private final GrpcRetryAlgorithmManager retryAlgorithmManager;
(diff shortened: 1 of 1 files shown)


rockspore:
@BenWhitehead Sorry for not doing this from the beginning. Would you mind taking a look?

rockspore:
@BenWhitehead I changed it to using system property. PTAL. Thanks!

google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcStorageOptions.java
```diff
@@ -134,6 +134,9 @@ public final class GrpcStorageOptions extends StorageOptions
   private static final String GCS_SCOPE = "https://www.googleapis.com/auth/devstorage.full_control";
   private static final Set<String> SCOPES = ImmutableSet.of(GCS_SCOPE);
   private static final String DEFAULT_HOST = "https://storage.googleapis.com";
+  // If true, disable the bound-token-by-default feature for DirectPath.
+  private static final boolean DIRECT_PATH_BOUND_TOKEN_DISABLED =
+      Boolean.parseBoolean(System.getProperty("STORAGE_DISABLE_DIRECTPATH_BOUND_TOKEN", "false"));
```
BenWhitehead: What do you think of the following name instead?
```suggestion
Boolean.parseBoolean(System.getProperty("com.google.cloud.storage.grpc.bound_token", "false"));
```

It's namespaced to storage specifically, and has grpc in the name making it more obvious it only applies to grpc. I omitted directpath, because IIRC this setting won't have any impact on CFE traffic even if it is used.

rockspore: Yeah this is great. Sorry I wasn't familiar with the best practice here. Thanks for the tip!

BenWhitehead (approved):

PR #3377, opened 2025-10-30
chore: fix xml parsing of StringEnumValue's so that the enum contract is not broken

BenWhitehead:


google-cloud-storage/src/main/java/com/google/cloud/storage/XmlObjectParser.java
```diff
@@ -33,6 +48,31 @@ final class XmlObjectParser implements ObjectParser {
   public XmlObjectParser(XmlMapper xmlMapper) {
     this.xmlMapper = xmlMapper;
     this.xmlMapper.registerModule(new JavaTimeModule());
+    this.xmlMapper.registerModule(
+        new Module() {
```
ShreyasSinha, line 52: nit: Can we move it to a different module,to improve readability.

BenWhitehead:

PR #3382, opened 2025-11-04
chore: add ObjectLockMode [de]serializers to XmlObjectParser for sake of completeness

BenWhitehead:
The value provided in CreateMultipartUploadRequest is set to the `x-goog-object-lock-mode` header, this change simply ensures that if the request is serialized/deserialized using our mapper it will work as expected.

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/XmlObjectParser.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/XmlObjectParser.java
@@ -31,6 +31,7 @@ import com.fasterxml.jackson.dataformat.xml.XmlMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.google.api.client.util.ObjectParser;
 import com.google.cloud.StringEnumValue;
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/XmlObjectParser.java
```diff
@@ -67,13 +68,18 @@ public Version version() {
           public void setupModule(SetupContext context) {
             context.addSerializers(
                 new SimpleSerializers(
-                    ImmutableList.of(new StringEnumValueSerializer<>(StorageClass.class))));
+                    ImmutableList.of(
+                        new StringEnumValueSerializer<>(StorageClass.class),
+                        new StringEnumValueSerializer<>(ObjectLockMode.class))));
```
ShreyasSinha, line 73: Is something similar required for PredefinedAcl?

BenWhitehead: PredefinedAcl is a java enum, not a StringEnumValue, for that we would need to add both a new serializer and deserializer, which would be more appropriate for a separate PR.

PR #3404, opened 2025-11-17
feat: add support for partial success in ListBuckets for json

nidhiii-27:
This PR adds support for the returnPartialSuccess parameter in ListBuckets calls for JSON.
A new Storage.BucketListOption.returnPartialSuccess(true) is introduced. When set, the list operation returns a set containing both reachable buckets and unreachable buckets. Unreachable buckets are distinguished by having the isUnreachable=true flag set and are named using their full resource form (e.g., "projects/_/buckets/bucket1").

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonConversions.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonConversions.java
@@ -18,6 +18,7 @@ package com.google.cloud.storage;
 
 import static com.google.cloud.storage.Storage.BucketField.IP_FILTER;
 import static com.google.cloud.storage.Storage.BucketField.SOFT_DELETE_POLICY;
+import static com.google.cloud.storage.Utils.bucketNameCodec;
 import static com.google.cloud.storage.Utils.dateTimeCodec;
 import static com.google.cloud.storage.Utils.durationSecondsCodec;
 import static com.google.cloud.storage.Utils.ifNonNull;
@@ -609,7 +610,7 @@ final class JsonConversions {
 
   @SuppressWarnings("deprecation")
   private BucketInfo bucketInfoDecode(com.google.api.services.storage.model.Bucket from) {
-    BucketInfo.Builder to = new BucketInfo.BuilderImpl(from.getName());
+    BucketInfo.Builder to = new BucketInfo.BuilderImpl(bucketNameCodec.decode(from.getName()));
     ifNonNull(from.getProjectNumber(), to::setProject);
     ifNonNull(from.getAcl(), toListOf(bucketAcl()::decode), to::setAcl);
     ifNonNull(from.getCors(), toListOf(cors()::decode), to::setCors);
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
@@ -17,22 +17,23 @@
 package com.google.cloud.storage.it;
 
 import static com.google.common.truth.Truth.assertThat;
-import static org.junit.Assert.assertTrue;
 
 import com.google.api.gax.paging.Page;
 import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.Storage.BucketListOption;
 import com.google.cloud.storage.TransportCompatibility.Transport;
 import com.google.cloud.storage.it.runner.StorageITRunner;
 import com.google.cloud.storage.it.runner.annotations.Backend;
+import com.google.cloud.storage.it.runner.annotations.BucketFixture;
+import com.google.cloud.storage.it.runner.annotations.BucketType;
 import com.google.cloud.storage.it.runner.annotations.CrossRun;
 import com.google.cloud.storage.it.runner.annotations.Inject;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
-import java.util.stream.StreamSupport;
-import org.junit.After;
-import org.junit.Before;
+import com.google.cloud.storage.it.runner.registry.Generator;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+import java.util.stream.Collectors;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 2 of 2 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/spi/v1/HttpStorageRpc.java
```diff
@@ -530,6 +537,14 @@ private static String detectContentType(StorageObject object, Map<Option, ?> opt
     return firstNonNull(contentType, "application/octet-stream");
   }
 
+  private static Function<String, Bucket> createUnreachableBucket() {
+    return new Function<String, Bucket>() {
+      public Bucket apply(String bucketName) {
+        return new Bucket().setName(bucketName).set("isUnreachable", "true");
+      }
+    };
```
BenWhitehead: nit: this can be much more terse
```suggestion
    return bucketName -> new Bucket().setName(bucketName).set("isUnreachable", "true");
```

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -2638,6 +2639,8 @@ public Builder setRetentionPeriodDuration(Duration retentionPeriod) {
      */
     public abstract Builder setIpFilter(IpFilter ipFilter);
 
+    abstract Builder setIsUnreachable(Boolean isUnreachable);
```
BenWhitehead: this should be public, so customers can set it in any tests they might want to write.

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
```diff
@@ -3886,6 +3907,11 @@ public HierarchicalNamespace getHierarchicalNamespace() {
     return customerSuppliedEncryptionEnforcementConfig;
   }
 
+  /** */
```
BenWhitehead: Empty comment.

Please fill this out, and link to any GCS documentation that explains what unreachable buckets are.

google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java
```diff
@@ -2476,6 +2476,11 @@ public static BucketListOption pageToken(@NonNull String pageToken) {
       return new BucketListOption(UnifiedOpts.pageToken(pageToken));
     }
 
+    @TransportCompatibility({Transport.HTTP, Transport.GRPC})
```
BenWhitehead: This PR only updates handling for HTTP, this annotation value should not include GRPC until GRPC is updated.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+  private static final String NORMAL_BUCKET_NAME = "normal_bucket";
+  // For testing purposes, the TESTBENCH considers a bucket to be unreachable if the bucket name
+  // contains "unreachable"
+  private static final String UNREACHABLE_BUCKET_NAME = "unreachable_bucket";
+
+  // The unreachable buckets are returned as a list of bucket resource names in string form. (e.g.
+  // "projects/_/buckets/bucket1")
+  private static final String EXPECTED_UNREACHABLE_BUCKET_NAME =
+      "projects/_/buckets/" + UNREACHABLE_BUCKET_NAME;
```
BenWhitehead: testbench is returning `projects/_/buckets/*` from the JSON endpoint? That's unexpected to me. Nothing else in the json api today uses that format for bucket names.

The comment from the [discovery doc](https://storage.googleapis.com/$discovery/rest?version=v1) doesn't make any mention of this either
> The list of bucket resource names that could not be reached during the listing operation

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplMockitoTest.java
```diff
@@ -785,6 +787,47 @@ public void testListBuckets() {
     assertArrayEquals(bucketList.toArray(), Iterables.toArray(page.getValues(), Bucket.class));
   }
 
+  @Test
+  public void testListBucketWithPartialSuccess() {
```
BenWhitehead: this test isn't terribly interesting, because StorageImpl doesn't contain any logic for the change.

This should probably be tested at the HttpStorageRpc layer to verify it is unioning things correctly.

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
nidhiii-27: done

google-cloud-storage/src/main/java/com/google/cloud/storage/BucketInfo.java
nidhiii-27: no specific documentation, added explanation though

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
nidhiii-27: resolved offline

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @Test
+  public void testListBucketWithoutPartialSuccess() {
+    Page<Bucket> page = storage.list();
+    ImmutableList<String> bucketNames =
+        StreamSupport.stream(page.iterateAll().spliterator(), false)
+            .map(Bucket::getName)
+            .collect(ImmutableList.toImmutableList());
+    assertThat(bucketNames).contains(NORMAL_BUCKET_NAME);
```
Dhriti07: Nit: Refactor test cases into Arrange, Act Asset

nidhiii-27: done

google-cloud-storage/src/test/java/com/google/cloud/storage/StorageImplMockitoTest.java
nidhiii-27: I'll add this test in a follow-up PR.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(StorageITRunner.class)
+@CrossRun(
+    backends = {Backend.TEST_BENCH},
+    transports = {Transport.HTTP})
+public class ITListBucketTest {
```
ShreyasSinha, line 44: Can you add a testcase with multiple unreachable buckets

google-cloud-storage/src/main/java/com/google/cloud/storage/JsonConversions.java
```diff
@@ -674,6 +674,9 @@ private BucketInfo bucketInfoDecode(com.google.api.services.storage.model.Bucket
     ifNonNull(from.getObjectRetention(), this::objectRetentionDecode, to::setObjectRetention);
     ifNonNull(from.getSoftDeletePolicy(), this::softDeletePolicyDecode, to::setSoftDeletePolicy);
     ifNonNull(from.getIpFilter(), ipFilterCodec::decode, to::setIpFilter);
+    if (from.containsKey("isUnreachable")) {
+      to.setIsUnreachable(Boolean.TRUE);
+    }
```
BenWhitehead, line 680: Spoke about this offline as well, but we need to update the logic here to handle the bucket resource names the same way grpc does.

On line https://github.com/googleapis/java-storage/pull/3404/files#diff-02854c6353038f671563a52b22658f9dcb49e66d8eb45a822439e90aa60b11f6R612 we need to add change the first line to be `BucketInfo.Builder to = new BucketInfo.BuilderImpl(bucketNameCodec.decode(from.getName()));` like in https://github.com/googleapis/java-storage/blob/3a95990b632261e1b1f68b3be01d60abc32a6cf5/google-cloud-storage/src/main/java/com/google/cloud/storage/GrpcConversions.java#L322


Even in grpc we don't expose `projects/_/buckets/` everything is normalized to json style bucket names.

BenWhitehead (changes requested):

google-cloud-storage/src/main/java/com/google/cloud/storage/JsonConversions.java
BenWhitehead: While this does deviate from how the API returns the value, and possibly how other languages have exposed things. This change does mean that the Storage sdk here is internally consistent, and customers can reliably use bucket names regardless of unavailability.

nidhiii-27: Done in PR #3415

PR #3413, opened 2025-12-05
feat(mpu): breaking change modifying getters and setters in MPU to be more inline with s3 interface.

ShreyasSinha:
This PR renames the getters and setters to be more inline with s3 interface.

BREAKING CHANGE: modifies the Beta APIs introduced in version 2.60


BenWhitehead:
if this is merged with `feat!` in the subject it will increment the major version of the whole repo. Instead I've changed it to scope it to mpu with breaking in the description.

PR #3415, opened 2025-12-05
feat: add support for partial success in ListBuckets for json

nidhiii-27:
This PR adds support for the returnPartialSuccess parameter in ListBuckets calls for JSON.
A new Storage.BucketListOption.returnPartialSuccess(true) is introduced. When set, the list operation returns a set containing both reachable buckets and unreachable buckets. Unreachable buckets are distinguished by having the isUnreachable=true flag set.

### DIFF
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
@@ -52,6 +52,8 @@ public class ITListBucketTest {
 
   @Inject public Generator generator;
 
+  private static final String UNREACHABLE_BUCKET_SUFFIX = ".unreachable";
+
   @Test
   public void testListBucketWithPartialSuccess() throws Exception {
     doTest(Reachability.Unreachable, BucketListOption.returnPartialSuccess(true));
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
```diff
@@ -0,0 +1,101 @@
...
+  @Test
+  public void testListBucketWithoutPartialSuccess() throws Exception {
+    doTest(Reachability.Reachable);
+  }
+
+  private void doTest(
+      Reachability expectedReachabilityOfUnreachableBucket, BucketListOption... bucketListOption)
+      throws Exception {
+    // TESTBENCH considers a bucket to be unreachable if the bucket name contains "unreachable"
+    String name = generator.randomBucketName() + ".unreachable";
```
Dhriti07: Nit: Refactor ".unreachable" into a const like UNREACHABLE_BUCKET_SUFFIX

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
```diff
@@ -0,0 +1,101 @@
...
+    BucketInfo info = BucketInfo.of(name);
+    try (TemporaryBucket tmpBucket =
+        TemporaryBucket.newBuilder().setBucketInfo(info).setStorage(storage).build()) {
+      // bucket name to unreachable status
```
Dhriti07: Nit: "Mapping bucket name to unreachable status" or we can remove this comment since the code is clear.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
nidhiii-27: done

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITListBucketTest.java
nidhiii-27: done

PR #3434, opened 2025-12-15
deps: update sdk-platform-java-config to 3.55.0-rc1

diegomarquezp:


.pom.xml.swp
blakeli0, line 1: Are these .swp files accidental changes?

diegomarquezp: Yes. I just removed them.

PR #3445, opened 2025-12-19
feat: add end-to-end checksumming for json resumable uploads

BenWhitehead:
If a crc32cMatch or md5Match are not specified when creating the resumable upload, a running checksum will be computed and included in the finalization call.

* Add plumbing logic for capture/restore logic of BlobWriteChannelV2
* Add JsonResumableWrite hasher and cumulativeCrc32c to track the hasher and running checksum when necessary
* Update JsonResumableSession to send cumulativeCrc32c when finalizing, and to update the running crc32c when a successful result is received
* Add integration tests to ITJsonResumableSessionTest to verify checksum is sent and correctly computed, even when retried or partial consumed
* Update Hasher.nullSafeConcat to be type dependent upon the left hand side provided type
* Include `X-Goog-Hash` if present when an upload request fails


ShreyasSinha:
Merged as part of Nidhi's PR. Closing this one.

PR #3452, opened 2026-01-04
chore: normalize host URL in MultipartUploadHttpRequestManager

### LINKED ISSUE #3451: MultipartUploadClient fails with custom host (missing path separator)
## Context

I ran into this while working on adding native GCS multipart upload support to [s3proxy](https://github.com/gaul/s3proxy) (see [gaul/s3proxy#76](https://github.com/gaul/s3proxy/issues/76)). We use `setHost()` to configure the GCS endpoint, and some of the `MultipartUploadClient` operations started failing with weird DNS errors.

After some digging, I found that three of the six methods in `MultipartUploadClient` have a URL construction bug - they concatenate the bucket name directly to the host without a `/` separator.

## What happens

With `setHost("https://storage.googleapis.com")` (no trailing slash):

```
Expected: https://storage.googleapis.com/my-bucket?uploads
Actual:   https://storage.googleapis.commy-bucket?uploads
                                       ^^^ missing slash!
```

This causes a DNS lookup failure since `storage.googleapis.commy-bucket` isn't a valid hostname.

## Which methods are affected

| Method | Works? |
|--------|--------|
| `createMultipartUpload()` | ✅ |
| `listParts()` | ✅ |
| `abortMultipartUpload()` | ✅ |
| `uploadPart()` | ❌ |
| `completeMultipartUpload()` | ❌ |
| `listMultipartUploads()` | ❌ |

The frustrating part is that `createMultipartUpload()` works fine, so you don't realize anything is wrong until you try to actually upload a part or complete the upload.

## Reproduction

```java
HttpStorageOptions options = StorageOptions.http()
    .setHost("https://storage.googleapis.com")  // no trailing slash
    .build();

MultipartUploadClient client = MultipartUploadClient.create(
    MultipartUploadSettings.of(options));

// This works fine:
CreateMultipartUploadResponse create = client.createMultipartUpload(...);

// This fails with DNS error:
client.uploadPart(...);  // StorageException: storage.googleapis.commy-bucket
```

## Workaround

Add a trailing slash to your host:

```java
.setHost("https://storage.googleapis.com/")  // note the trailing slash
```

Or just don't call `setHost()` at all if you're using the default GCS endpoint - the SDK adds the trailing slash automatically in that case.

## Root cause

In `MultipartUploadHttpRequestManager.java`, the working methods use:
```java
UriTemplate.expand(uri.toString(), "{bucket}/{key}?uploads", params, false);
```

But the broken methods do:
```java
UriTemplate.expand(uri.toString() + "{bucket}?uploads...", params, false);
```

The difference is subtle but the second form concatenates directly without ensuring there's a `/` between the host and the bucket.

## Why tests didn't catch this

The test harness (`FakeHttpServer`) always creates endpoints with a trailing slash:
```java
URI endpoint = URI.create("http://localhost:" + port + "/");
```

So the bug is masked in tests. Real-world usage with explicit `setHost()` exposes it.

louisnow:
Fixes #3451

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/MultipartUploadHttpRequestManager.java
@@ -253,7 +253,11 @@ final class MultipartUploadHttpRequestManager {
         storage.getRequestFactory(),
         new XmlObjectParser(new XmlMapper()),
         options.getMergedHeaderProvider(FixedHeaderProvider.create(stableHeaders.build())),
-        URI.create(options.getHost()));
+        URI.create(ensureTrailingSlash(options.getHost())));
+  }
+
+  private static String ensureTrailingSlash(String host) {
+    return host.endsWith("/") ? host : host + "/";
   }
 
   private void addChecksumHeader(@Nullable Crc32cLengthKnown crc32c, HttpHeaders headers) {
--- a/google-cloud-storage/src/test/java/com/google/cloud/storage/FakeHttpServer.java
+++ b/google-cloud-storage/src/test/java/com/google/cloud/storage/FakeHttpServer.java
@@ -93,6 +93,10 @@ final class FakeHttpServer implements AutoCloseable {
   }
 
   static FakeHttpServer of(HttpRequestHandler server) {
+    return of(server, true);
+  }
+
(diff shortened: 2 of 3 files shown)


Dhriti07 (changes requested):
Thanks for raising the pr.
Can you please change the title from fix to chore and update the pr description.

louisnow:
@Dhriti07 I've updated the PR title and description and changed the commit to also say chore:

PR #3455, opened 2026-01-05
fix: validate blob paths to prevent directory traversal in TransferManager downloads

nidhiii-27:
This change ensures that blobs containing traversal sequences (e.g., "../")  are skipped during directory downloads if they resolve outside the intended target directory. A warning in the form of an exception is attached to the DownloadResults for each skipped blob.
Fixes: b/449616429


google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/DirectDownloadCallable.java
```diff
@@ -89,7 +92,7 @@ public DownloadResult call() {
     }
     DownloadResult result =
         DownloadResult.newBuilder(originalBlob, TransferStatus.SUCCESS)
-            .setOutputDestination(path.toAbsolutePath())
+            .setOutputDestination(destPath.toAbsolutePath())
```
Dhriti07: From the flow of this PR it looks like destPath being passed is already converted to absolutePath in createAndValidateDestPath function. We can skip this call unless there is another path that would require this to be called.

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITTransferManagerTest.java
```diff
@@ -635,6 +636,87 @@ public void bucketNameFromUploadBlobInfoFactoryMustMatchConfig() throws Exceptio
...
+            results.stream().filter(res -> res.getStatus() == TransferStatus.SUCCESS).count();
+        assertThat(successCount).isEqualTo(blobs.size());
+
+        // Verify that the malicious blob was blocked/skipped
+        Optional<DownloadResult> blockedResult =
+            results.stream()
+                .filter(res -> res.getInput().getName().equals(maliciousName))
+                .findFirst();
+
+        assertThat(blockedResult.isPresent()).isTrue();
```
Dhriti07: Nit: Can we use assertThat(blockedResult).isPresent();

google-cloud-storage/src/test/java/com/google/cloud/storage/it/ITTransferManagerTest.java
```diff
@@ -635,6 +636,87 @@ public void bucketNameFromUploadBlobInfoFactoryMustMatchConfig() throws Exceptio
...
+        assertThat(blockedResult.get().getStatus()).isEqualTo(TransferStatus.FAILED_TO_START);
+        assertThat(blockedResult.get().getException())
+            .isInstanceOf(PathTraversalBlockedException.class);
+        assertThat(blockedResult.get().getException().getMessage()).contains("blocked");
+      } finally {
+        storage.delete(maliciousBlob.getBlobId());
```
Dhriti07, line 680: Consider adding cleanUpFiles call here for the case where the test fails (due to any future changes) and files are created.

google-cloud-storage/src/main/java/com/google/cloud/storage/transfermanager/TransferManagerImpl.java
```diff
@@ -145,15 +145,20 @@ public void close() throws Exception {
     Storage.BlobSourceOption[] opts =
         config.getOptionsPerRequest().toArray(new Storage.BlobSourceOption[0]);
     List<ApiFuture<DownloadResult>> downloadTasks = new ArrayList<>();
-    if (!transferManagerConfig.isAllowDivideAndConquerDownload()) {
-      for (BlobInfo blob : blobs) {
-        DirectDownloadCallable callable = new DirectDownloadCallable(storage, blob, config, opts);
-        downloadTasks.add(convert(executor.submit(callable)));
+    for (BlobInfo blob : blobs) {
+      Path destPath = TransferManagerUtils.createAndValidateDestPath(config, blob);
```
Dhriti07, line 149: In createAndValidateDestPath, the targetDirectory is calculated for every blob in the loop by calling .toAbsolutePath().normalize(). Since the download directory is dependent on config and not blob, we can pre-calculate this path once and pass it as an argument to avoid redundant operations.

PR #3465, opened 2026-01-21
chore: remove @BetaApi tag from MPU

nidhiii-27:
Removed the @BetaApi tag from all the MPU APIs

### DIFF
--- a/google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/AbortMultipartUploadResponse.java
+++ b/google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/AbortMultipartUploadResponse.java
@@ -24,5 +24,4 @@ import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
  * @since 2.60.0
  */
 @JacksonXmlRootElement(localName = "AbortMultipartUploadResponse")
-public final class AbortMultipartUploadResponse {
-}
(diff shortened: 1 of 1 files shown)


google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/AbortMultipartUploadRequest.java
```diff
@@ -15,15 +15,12 @@
  */
```
ShreyasSinha, line 15: Can you also remove beta annotation from APIs as well

ShreyasSinha: Taken care in a different PR.

google-cloud-storage/src/main/java/com/google/cloud/storage/multipartupload/model/UploadPartRequest.java
```diff
@@ -108,9 +101,8 @@ public String crc32c() {
    * @return the user-project.
    * @see <a
    *     href="https://docs.cloud.google.com/storage/docs/xml-api/reference-headers#xgooguserproject">x-goog-user-project</a>
-   * @since 2.61 This new api is in preview and is subject to breaking changes.
+   * @since 2.61
```
Dhriti07: Nit: Since 2.61.0

PR #3474, opened 2026-01-27
chore: exclude opentelemetry in grpc

JoeWang1127:
Prepare for shared config 3.56.0 upgrade. We need to exclude opentelemetry in grpc because the version is downgraded in shared config.


JoeWang1127:
Superseded by https://github.com/googleapis/java-storage/pull/3468

